PR #3, opened 2016-09-16
Generator: remove genxml & minixsv in favor of standart lxml

khancyr:
The output is the same but it generate file faster and is compatible with python 3


peterbarker:
This would be wonderful if we can get it in with no loss of functionality.  Fast and lots of code removed!

However, if I introduce elements into the schema which should cause it to fail to validate against the schema, the old code fails but the new code ignores the badness:

```
diff --git a/message_definitions/v1.0/matrixpilot.xml b/message_definitions/v1.0/matrixpilot.xml
index 29d368d..f859b03 100644
--- a/message_definitions/v1.0/matrixpilot.xml
+++ b/message_definitions/v1.0/matrixpilot.xml
@@ -9,7 +9,7 @@
       is general enough to move into common.xml later
     -->
 
-
+ <invalidx>xyzzy</invalidx>
 
 
      <enums>
```

khancyr:
Sorry I didn't check xml validation. I looking at it. But it is really actual xsd schema isn't valid... but is validate by current parser... so many regex T.T

khancyr:
@peterbarker 
Work in progress. I correct xml validation but there are problems.
I need to  modify xsd schema since it wasn't compilant to xsd1.0 norm ... 
I need to remove check on name as the regex isn't working on xsd (can't use ?! ) and I don't find alternative.
Now it catch an error on ardupilotmega.xml see https://github.com/ArduPilot/mavlink/pull/17

khancyr:
but wait that not an error in xml... in in xsd schema again. the sequence is wrong...

peterbarker:
@khancyr This looks like it's going to require some staring at.  Let me know when you're happy with it, please?

khancyr:
@peterbarker good to go ! 
As xsd regex doesn't permit negation or word check, I put it in mavgen_validate.
I also make build failed with red comments in case of error (I don't know how to raise custom exception for regex check...)
Finally, I add future as import for print ( I have added latter for python3 compatibility)

I assume I forget nothing.

generator/mavgen.py
```diff
@@ -168,11 +193,14 @@ def mavgen_python_dialect(dialect, wire_protocol):
     sys.stdout = io.StringIO()
     try:
         xml = os.path.relpath(xml)
-        mavgen( opts, [xml] )
+        if not mavgen( opts, [xml] ):
```
peterbarker: May as well remove those extra spaces so when the style police come along its one less thing to fix :-)

setup.py
```diff
@@ -53,16 +53,18 @@
...
     for xml in v20_dialects:
         dialect = os.path.basename(xml)[:-4]
         wildcard = os.getenv("MAVLINK_DIALECT",'*')
         if not fnmatch.fnmatch(dialect, wildcard):
             continue
         print("Building %s for protocol 2.0" % xml)
-        mavgen.mavgen_python_dialect(dialect, mavparse.PROTOCOL_2_0)
-
+        if not mavgen.mavgen_python_dialect(dialect, mavparse.PROTOCOL_2_0):
+            print("Building %s for protocol 2.0" % xml)
```
peterbarker: This message is inconsistent with the one for 1.0 - needs the word "failed"

generator/mavgen.py
```diff
@@ -34,23 +35,45 @@ def mavgen(opts, args) :
...
         try:
-            from lib.genxmlif import GenXmlIfError
-            from lib.minixsv import pyxsval
+            from lxml import etree
+            with open(schemaFile, 'r') as f:
+                xmlschema_root = etree.parse(f)
+                xmlschema = etree.XMLSchema(xmlschema_root)
         except:
-            print("WARNING: Unable to load XML validator libraries. XML validation will not be performed")
+            print("\033[1;31;40mWARNING: Unable to load XML validator libraries. XML validation will not be performed\033[0m", file=sys.stderr)
```
khancyr: Ok I will remove color. It doesn't work on windows BTW...

generator/mavgen.py
```diff
@@ -34,23 +35,45 @@ def mavgen(opts, args) :
...
             opts.validate = False
 
-    def mavgen_validate(fname, schema, errorLimitNumber) :
-        """Uses minixsv to validate an XML file with a given XSD schema file. We define mavgen_validate
+    def mavgen_validate(fname):
+        """Uses lxml to validate an XML file with a given XSD schema file. We define mavgen_validate
```
khancyr: Yep should update that. It use the defaul xsdschema defined at start

generator/mavgen.py
khancyr: yep : did you see http://discuss.ardupilot.org/t/python-code-standard/9128/6 ?

setup.py
khancyr: oups ctrl-v error !

khancyr:
@peterbarker I will correct that later today.
I test on windows and it works ( but not color 😢 )  lxml in from pip doesn't work but they provide an installer.
I am installing cygwin to test but it should be good. I don't have OSX device to test....

khancyr:
@peterbarker @tridge 
Ok I correct my errors ! I test on Windows. It works natively with python (Need to install lmxl with msi...pip install seems broken). And it also works in cygwin (32 or 64bits)
When you are good with the change, I will force push a PR rebased and cleaner !

generator/mavgen.py
```diff
@@ -34,23 +37,45 @@ def mavgen(opts, args) :
...
+                xmldocument = etree.parse(f)
+                xmlschema.assertValid(xmldocument)
+                regex = re.compile("^(break$|case$|class$|catch$|const$|continue$|debugger$|default$|delete$|do$|else$|\
+                                    export$|extends$|finally$|for$|function$|if$|import$|in$|instanceof$|let$|new$|\
+                                    return$|super$|switch$|this$|throw$|try$|typeof$|var$|void$|while$|with$|yield$|\
+                                    enum$|await$|implements$|package$|protected$|static$|interface$|private$|public$|\
+                                    abstract$|boolean$|byte$|char$|double$|final$|float$|goto$|int$|long$|native$|\
+                                    short$|synchronized$|transient$|volatile$).*", re.IGNORECASE)
+                for element in xmldocument.iter("*"):
+                    if element.tag in ["enum", "entry", "message", "field"]:
```
vooon: Maybe:
```python
for element in filter(lambda e: e.tag in ['enum', ...], xmldocument.iter('*')):
```

generator/mavgen.py
```diff
@@ -34,23 +37,45 @@ def mavgen(opts, args) :
...
+                        match = regex.search(element.get('name'))
+                        if match is not None:
```
vooon:
```python
if regex.search(...):
```

generator/mavgen.py
```diff
@@ -34,23 +37,45 @@ def mavgen(opts, args) :
...
+        """Uses lxml to validate an XML file. We define mavgen_validate
            here because it relies on the XML libs that were loaded in mavgen(), so it can't be called standalone"""
-        # use default values of minixsv, location of the schema file must be specified in the XML file
-        domTreeWrapper = pyxsval.parseAndValidate(fname, xsdFile=schema, errorLimit=errorLimitNumber)
+        xmlvalid = True
+        try:
+            with open(xmlfile, 'r') as f:
+                xmldocument = etree.parse(f)
+                xmlschema.assertValid(xmldocument)
+                regex = re.compile("^(break$|case$|class$|catch$|const$|continue$|debugger$|default$|delete$|do$|else$|\
```
vooon: s/regex/forbidden_names_re/ ?

generator/mavgen.py
vooon: Or combination:
```python
lambda e: e.tag in [...] and r.search(e.get('name')
```

generator/mavgen.py
vooon: Hmm, instead of filter better to use generator expression `(e for e in xml.iter('*') if ...)`

generator/mavgen.py
khancyr: yes, it's better!

generator/mavgen.py
khancyr: will look at that! I am not expert on subject

khancyr:
@vooon Did see that there was another parser in mavparse.py. I will try to switch it to lxml too.

khancyr:
@peterbarker Ok, I squash everything, one last review and we can get in.
@vooon I not expert on advance function, if you want to make a PR to improve the parsing with regex, you are welcome ! 
As for mavparse.py, I try to remove expat parser for lxml, but it is quite simplier to use in that case or it will need heavy modification. 
My first motivation on this PR was to get ride of old genxml & minixsv that are not python3 compatible

vooon:
@khancyr ok. Just using only lxml may give some speed-up (no double parsing).

Filter may be:

```python
for element in (e for e in xmldocument.iter('*') \
    if e.tag in ('enum', 'entry', 'message', 'field') and r.search(e.get('name')) ):
    pass
```

Hmm, looks to complex. Reduce all to:

```python
for element in xmldocument.iter('*'):
    if element.tag in ('enum', 'entry', 'message', 'field') and forbidden_names_re.search(element.get('name')):
        pass
```

vooon:
Function documentation (Py 3.5):
```
In [14]: root.iter?
Type:        builtin_function_or_method
String form: <built-in method iter of lxml.etree._Element object at 0x7f4afb21c548>
Docstring:
iter(self, tag=None, *tags)

Iterate over all elements in the subtree in document order (depth
first pre-order), starting with this element.

Can be restricted to find only elements with a specific tag:
pass ``"{ns}localname"`` as tag. Either or both of ``ns`` and
``localname`` can be ``*`` for a wildcard; ``ns`` can be empty
for no namespace. ``"localname"`` is equivalent to ``"{}localname"``
(i.e. no namespace) but ``"*"`` is ``"{*}*"`` (any or no namespace),
not ``"{}*"``.

You can also pass the Element, Comment, ProcessingInstruction and
Entity factory functions to look only for the specific element type.

Passing more than one tag will let the iterator return all elements
matching any of these tags, in document order.
```

So right condition:
```python
for element in xmldocument.iter('enum', 'entry', 'message', 'field'):
    if forbidden_names_re.search(element.get('name')):
        pass
```

khancyr:
@vooon done !
BTW, why use pass ?

vooon:
Pass is just a place-holder.

khancyr:
ok! Thanks for help ! Did you agree with other changes? 
Normaly it is working both python2 and 3 for code generation and parsing. I test with ardupilot autotest and all vehicules are working in python 3 !

vooon:
Yes, code looks good to me. I think maintainer maybe ok too.

khancyr:
@peterbaker,  @tridge lets merge!

tridge:
looks good to me! applied, and thank you very much for the effort on this!

PR #10, opened 2016-09-29
All: initial python3 support

khancyr:
this PR gives pymavlink python3 support 
Generator is working fine (no difference with python2) 
I don't test all examples (some aren't working) and all cases but it should be working

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,8 +2,8 @@ language: python
 python:
   - "2.6"
   - "2.7"
-  - "3.2"
   - "3.3"
+  - "3.5"
 
 script:
   # NOTE: we must do all testing on the installed python package, not 
--- a/examples/apmsetrate.py
+++ b/examples/apmsetrate.py
@@ -3,6 +3,8 @@
 '''
 set stream rate on an APM
 '''
+from __future__ import print_function
+from builtins import range
 
 import sys, struct, time, os
 
@@ -47,7 +49,7 @@ master = mavutil.mavlink_connection(args.device, baud=args.baudrate)
 wait_heartbeat(master)
 
 print("Sending all stream request for rate %u" % args.rate)
-for i in range(0, 3):
+for i in list(range(0, 3)):
     master.mav.request_data_stream_send(master.target_system, master.target_component,
                                         mavutil.mavlink.MAV_DATA_STREAM_ALL, args.rate, 1)
 if args.showmessages:
--- a/examples/apmsetrate.py
+++ b/examples/apmsetrate.py
@@ -49,7 +49,7 @@ master = mavutil.mavlink_connection(args.device, baud=args.baudrate)
 wait_heartbeat(master)
 
 print("Sending all stream request for rate %u" % args.rate)
-for i in list(range(0, 3)):
+for i in range(0, 3):
     master.mav.request_data_stream_send(master.target_system, master.target_component,
                                         mavutil.mavlink.MAV_DATA_STREAM_ALL, args.rate, 1)
 if args.showmessages:
--- a/examples/bwtest.py
+++ b/examples/bwtest.py
@@ -3,6 +3,7 @@
 '''
 check bandwidth of link
 '''
+from __future__ import print_function
 
 import sys, struct, time, os
 
--- a/examples/magtest.py
+++ b/examples/magtest.py
@@ -4,6 +4,7 @@
 rotate APMs on bench to test magnetometers
 
 '''
+from __future__ import print_function
 
 import sys, os, time
 from math import radians
--- a/examples/mav_accel.py
+++ b/examples/mav_accel.py
@@ -3,6 +3,7 @@
 '''
 show accel calibration for a set of logs
 '''
+from __future__ import print_function
 
 import sys, time, os
 
--- a/examples/mavtcpsniff.py
+++ b/examples/mavtcpsniff.py
@@ -13,6 +13,7 @@ hint:
 Copyright Sept 2012 David "Buzz" Bussenschutt
 Released under GNU GPL version 3 or later
 '''
+from __future__ import print_function
 
 import sys, time, os, struct
 
--- a/examples/mavtest.py
+++ b/examples/mavtest.py
@@ -1,5 +1,8 @@
 #!/usr/bin/env python
 
+from __future__ import print_function
+from builtins import chr
+from builtins import object
 import sys, os
 
 from pymavlink.dialects.v10 import ardupilotmega as mavlink1
--- a/examples/mavtest.py
+++ b/examples/mavtest.py
@@ -1,7 +1,6 @@
 #!/usr/bin/env python
 
 from __future__ import print_function
-from builtins import chr
 from builtins import object
 import sys, os
 
--- a/generator/mavgen.py
+++ b/generator/mavgen.py
@@ -9,6 +9,9 @@ Released under GNU GPL version 3 or later
 '''
 
 from __future__ import print_function
+from future import standard_library
+standard_library.install_aliases()
+from builtins import object
 import sys, textwrap, os, copy
 import re
 from . import mavparse
@@ -154,7 +157,7 @@ def mavgen(opts, args):
     return True
 
 # build all the dialects in the dialects subpackage
-class Opts:
+class Opts(object):
     def __init__(self, output, wire_protocol=DEFAULT_WIRE_PROTOCOL, language=DEFAULT_LANGUAGE, validate=DEFAULT_VALIDATE, error_limit=DEFAULT_ERROR_LIMIT):
         self.wire_protocol = wire_protocol
         self.error_limit = error_limit
--- a/generator/mavgen_c.py
+++ b/generator/mavgen_c.py
@@ -5,6 +5,10 @@ parse a MAVLink protocol XML file and generate a C implementation
 Copyright Andrew Tridgell 2011
 Released under GNU GPL version 3 or later
 '''
+from __future__ import print_function
+from builtins import str
+from builtins import range
+from builtins import object
 
 import sys, textwrap, os, time
 from . import mavparse, mavtemplate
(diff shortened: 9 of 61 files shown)


generator/mavparse.py
```diff
@@ -100,7 +104,7 @@ def set_test_value(self):
         '''set a testsuite value for a MAVField'''
         if self.array_length:
             self.test_value = []
-            for i in range(self.array_length):
+            for i in list(range(self.array_length)):
```
vooon: Why?

mavutil.py
```diff
@@ -1589,7 +1593,7 @@ def mode_mapping_byname(mav_type):
         map = mode_mapping_tracker
     if map is None:
         return None
-    inv_map = dict((a, b) for (b, a) in map.items())
+    inv_map = dict((a, b) for (b, a) in list(map.items()))
```
vooon: Why list()?

mavwp.py
```diff
@@ -79,7 +82,7 @@ def insert(self, idx, w, comment=''):
 
     def reindex(self):
         '''reindex waypoints'''
-        for i in range(self.count()):
+        for i in list(range(self.count())):
```
vooon: Why???

rotmat.py
```diff
@@ -362,9 +365,9 @@ def test_euler():
     '''check that from_euler() and to_euler() are consistent'''
     m = Matrix3()
     from math import radians, degrees
-    for r in range(-179, 179, 3):
-        for p in range(-89, 89, 3):
-            for y in range(-179, 179, 3):
+    for r in list(range(-179, 179, 3)):
+        for p in list(range(-89, 89, 3)):
+            for y in list(range(-179, 179, 3)):
```
vooon: And again...

Py2 range: `list(xrange())`, In Py 3 range() works like xrange().
I do not see any reason to make list from range object.

setup.py
```diff
@@ -53,16 +55,18 @@
...
         dialect = os.path.basename(xml)[:-4]
         wildcard = os.getenv("MAVLINK_DIALECT",'*')
         if not fnmatch.fnmatch(dialect, wildcard):
             continue
         print("Building %s for protocol 2.0" % xml)
-        mavgen.mavgen_python_dialect(dialect, mavparse.PROTOCOL_2_0)
-
+        if not mavgen.mavgen_python_dialect(dialect, mavparse.PROTOCOL_2_0):
+            print("Building failed %s for protocol 2.0" % xml)
+            sys.exit(1)
```
vooon: While line after block end.

generator/mavgen.py
```diff
@@ -34,23 +40,45 @@ def mavgen(opts, args) :
...
+                                    enum$|await$|implements$|package$|protected$|static$|interface$|private$|public$|\
+                                    abstract$|boolean$|byte$|char$|double$|final$|float$|goto$|int$|long$|native$|\
+                                    short$|synchronized$|transient$|volatile$).*", re.IGNORECASE)
+                for element in xmldocument.iter("*"):
+                    if element.tag in ["enum", "entry", "message", "field"]:
+                        match = forbidden_names_re.search(element.get('name'))
+                        if match is not None:
+                            print("Validation error:", file=sys.stderr)
+                            print("Element : %s at line : %s contains forbidden word" % (element.tag, element.sourceline), file=sys.stderr)
+                            xmlvalid = False
```
vooon: Check this comment: https://github.com/ArduPilot/pymavlink/pull/3#issuecomment-250541991

khancyr:
@vooon thanks for help ! 
About the adding of list. I wasn't sure. I follow http://python-future.org/compatible_idioms.html#range
Atfer much reading, in for loop the behaviour is the same in pt2 and pt3... so I will revert that! 
map is not map()... mistake...

.travis.yml
```diff
@@ -2,8 +2,8 @@ language: python
 python:
   - "2.6"
```
cclauss: Are Python 2.6 and Python 3.2 support really needed?
Should we add "Nightly" to get the Python 3.6 builds now that it is in beta?

examples/apmsetrate.py
```diff
@@ -3,6 +3,8 @@
 '''
 set stream rate on an APM
 '''
+from __future__ import print_function
+from builtins import range
 
 import sys, struct, time, os
```
cclauss: struct, time, os are imported but not used.

Running flake8 or autopep8 on these files might help find other issues.

examples/bwtest.py
```diff
@@ -3,6 +3,7 @@
 '''
 check bandwidth of link
 '''
+from __future__ import print_function
 
 import sys, struct, time, os
```
cclauss: sys, struct, and os are imported but not used.

examples/magtest.py
```diff
@@ -4,6 +4,7 @@
 rotate APMs on bench to test magnetometers
 
 '''
+from __future__ import print_function
 
 import sys, os, time
```
cclauss: import sys, os, time are all unused.

examples/mav_accel.py
```diff
@@ -3,6 +3,7 @@
 '''
 show accel calibration for a set of logs
 '''
+from __future__ import print_function
 
 import sys, time, os
```
cclauss: import sys, time, os are all unused.

examples/mavtcpsniff.py
```diff
@@ -13,6 +13,7 @@
 Copyright Sept 2012 David "Buzz" Bussenschutt
 Released under GNU GPL version 3 or later
 '''
+from __future__ import print_function
 
 import sys, time, os, struct
```
cclauss: sys, os, and struct are imported but unused.  It would be useful to run flake8 or autopep8 on this file because the spacing was a bit messed up.

examples/mavtest.py
```diff
@@ -1,5 +1,8 @@
 #!/usr/bin/env python
 
+from __future__ import print_function
+from builtins import chr
+from builtins import object
 import sys, os
```
cclauss: sys and os are imported but not used.

.travis.yml
cclauss: Perhaps it would be good for hygiene to add:

```
    - pip install flake8
    # flake8 provides code improvement suggestions 
    # the --exit-zero flag prevents flake8 from halting the build
    flake8 . --count --exit-zero --max-line-length=127 --statistics
```

khancyr:
@cclauss thank for interest ! 
About : list(range it should have been remove by last commit.
About travis : I don't know which python version is currently supported (I assume only 2.7). So I didn't modify all. I let repo maintenair decide on this point ! Idealy we just need 2.7 and 3.5 support. 

About style ! Big question ! There is currently no guideline. I try to propose pep8 but it need to be adapted a little (get ride of lenght limit for example) and not mess with current comments, and last point, it will break all other works/pr that for non functional change. I agree that must be done and using flake8 and autopep in futur could improve redability, here was my proposal http://discuss.ardupilot.org/t/python-code-standard/9128/6
So style change should come in another PR !

cclauss:
@khancyr Good stuff.  Agreed on all points.

[setup.py says Python 2.7](https://github.com/dronekit/dronekit-python/blob/master/setup.py#L25) and not 2.6 and [the module's PyPI record](https://pypi.python.org/pypi/pymavlink/) matches that so dropping 2.6 seems safe.  I like the approach of 2.7 and 3.5 only.

I did ready your post on style and saw there was not yet broad buy-in so I put the --exit-zero option on the flake8 line.  That makes all of the flake8 output mere warnings/suggestions and not showstoppers.  I also set --max-line-length=120 because the GitHub builtin editor is 127 chars wide and the PEP8 limit of 79 chars is limiting on older codebases.  Again, this is only a waning, not a build stopper.

khancyr:
@cclauss --exit-zero option in interesting. Will try to speak about it to maintainers !

generator/mavgen_c.py
```diff
@@ -16,8 +20,8 @@ def generate_version_h(directory, xml):
     f = open(os.path.join(directory, "version.h"), mode='w')
     t.write(f,'''
 /** @file
- *	@brief MAVLink comm protocol built from ${basename}.xml
- *	@see http://mavlink.org
+ *  @brief MAVLink comm protocol built from ${basename}.xml
+ *  @see http://mavlink.org
  */
```
tridge: nearly all the changes in mavgen_c.py aren't in the python code but instead are in the formatting of the generated C code. I'd prefer to leave the formatting of the C alone unless there is some good reason to change it (I tried to make the generated C code as readable as I could)

tridge:
i've got this PR installed on my laptop and I'll see if I run across any issues using it in the next day or two

cclauss:
@tridge `future` is [already in requirements.txt](https://github.com/ArduPilot/pymavlink/blob/master/requirements.txt) but I tend to agree with your approach of avoiding using it.  Once you have a lot of code depending on it, removing it will become difficult.  We should try to make the code work in both P2 and P3 without future, builtins, etc. and then see where its use is absolutely necessary.

generator/mavgen_c.py
khancyr: Those are tabs, and tabs are evil ! hum by relooking at the file now, it seems that I make a mess of the indentation. Will rework that. Tabs are evil because they mix indentation ( I try to do the same on mavproxy, and break everything since tab and space are mixed ...so replace tabs by space break the syntaxe, it a nice time bomb)

khancyr:
@tridge I doubt we can avoid using future and it is pretty and simplier that try to catch python 2 to 3 or 3 to 2 compatibility errors . We could try to reduce its use to print only .

cclauss:
The module tab nanny is built into the standard library.  It does a good job of tab to space conversion.  That one file is a difficult one however.  Perhaps it would be safer to take that file out of this mega pull request and put in its own or to focus the code review on those changes. Whitespace matters.

tridge:
I've pushed this PR in as I really didn't want it to bitrot. I did notice an issue with python3 in mavlogdump.py though. The DFReader.py module which is used when dumping DF logs is broken under python3. Any chance you could look at that?

PR #14, opened 2016-10-07
Simplify with max()

cclauss:
Also see the comment on line 180.  If v1.0 is correct then we need a comment explaining why.

### DIFF
--- a/generator/mavgen.py
+++ b/generator/mavgen.py
@@ -109,7 +109,7 @@ def mavgen(opts, args):
             x.largest_payload = max(x.largest_payload, xml[-1].largest_payload)
 
     # work out max payload size across all includes
-    largest_payload = max(x.largest_payload for x in xml)
+    largest_payload = max(x.largest_payload for x in xml) if xml else 0
     for x in xml:
         x.largest_payload = largest_payload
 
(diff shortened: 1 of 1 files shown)


generator/mavgen.py
```diff
@@ -109,10 +109,7 @@ def mavgen_validate(xmlfile):
             x.largest_payload = max(x.largest_payload, xml[-1].largest_payload)
 
     # work out max payload size across all includes
-    largest_payload = 0
-    for x in xml:
-        if x.largest_payload > largest_payload:
-            largest_payload = x.largest_payload
+    largest_payload = max(x.largest_payload for x in xml)
```
peterbarker: Nice change!  However, it seems the xml list might be empty, and that isn't going to end well.  We need some protection in for this case.

generator/mavgen.py
```diff
@@ -180,7 +177,7 @@ def mavgen_python_dialect(dialect, wire_protocol):
         py = os.path.join(dialects, 'v20', dialect + '.py')
         xml = os.path.join(dialects, 'v20', dialect + '.xml')
         if not os.path.exists(xml):
-            xml = os.path.join(mdef, 'v1.0', dialect + '.xml')
+            xml = os.path.join(mdef, 'v1.0', dialect + '.xml')  # is v1.0 correct here??
```
peterbarker: not-so-nice a change :-)

I'm guessing this might have snuck in accidentally.  I'm not 100% sure, but I think v1.0 here is just an artifact of us using v1.0 message definitions (https://github.com/ArduPilot/pymavlink/blob/master/setup.py#L40).

This needs either a comment along the same lines as that one I just referenced, or correction.  Whichever it is, it doesn't belong in this PR :-)  Could you split this patch out, please?

peterbarker:
merged, thanks!

I squashed the two commits together as I pushed it in.

PR #15, opened 2016-10-08
Add flake8 --exit-zero to the Travis CI process

cclauss:
This provides maintainers with warnings about questionable code usage.

These are only warnings because of the --exit-zero option.


squilter:
Running this locally just gave me 2328 messages. I don't think this is particularly useful to anybody if it will give more that 2000 messages each time. Is there any setting to ignore warnings and just report errors?

cclauss:
Great minds think alike!  I am traveling now, but in a couple of hours I will update.  I will supress the whitespace messages that we can get autopep8 to do later.

squilter:
There's still more than 200. Most of them are valid style issues, but I think it's still too much to be useful. We could either suppress more, or we could try fixing the issues. I think that ideally, the test should always return 0 issues on master branch, so that it is then very obvious if a PR is introducing something bad.

cclauss:
~150 of the reported issues are imports of modules that do not get used.  This slows execution and uses ram so they should be fixed.  Experience shows that we should be able to get to zero reported quite quickly now that a light is shined on these issues.   @khancyr has a ton of theses issues fixed in his open PRs.  Several of them can be handled through the judicious use of autopep8.

cclauss:
The following is injudicious use of autopep8 and I do NOT recommed it but it is interesting to see how much automatic formatting can do:

- flake8 . --count --max-line-length=127 --statistics  # on Python3 generates 2266 issues
- autopep8 -i -r .
- flake8 . --count --max-line-length=127 --statistics  # on Python3 generates  657 issues

.travis.yml
```diff
@@ -1,9 +1,33 @@
...
-  - "3.3"
+  - "2.7.12"
+  - "3.5"
+
+install:
+  - pip install -r requirements.txt
+  - pip install flake8
+
+before_script:
+  # --exit-zero means that flake8 will only warn and not stop the build
```
peterbarker: Could we drop most of this into a config file for flake8?  Possibly in the "files" subdirectory?

We could then suggest git commit hooks to ensure we don't get any worse according to our chosen standards...

.travis.yml
```diff
@@ -1,9 +1,33 @@
 language: python
 python:
-  - "2.6"
```
peterbarker: Why are we removing these older version lines here?

cclauss: This matches setup.py as discussed https://github.com/ArduPilot/pymavlink/pull/10#issuecomment-251911149

.travis.yml
cclauss: Nice idea.  I will create a `.flake8` file at root.  I do not know how to make "git commit hooks" but would be fascinated to learn.

khancyr:
@cclauss or we can use http://pre-commit.com/ or https://codeclimate.com/ . I am looking into codeclimate to do the same for c++ on apm side

cclauss:
In https://github.com/ArduPilot/pymavlink/pull/20#issuecomment-253397194
@peterbarker wrote> _Some ("F401 'os' imported but unused") can often be fixed without stuffing this history up._

I am unclear how you want this done.  Perhaps you could demonstrate with one of the files.

peterbarker:
@cclauss I've squash your commit together, moved the flake8 configuration file back to .flake8 (sorry for stuff you around on that one; .flake8 looks to be the best idea as it is search-for by the flake8 tool natively, so you can just "flake8" in the pymavlink root).

I've also pushed up a patch which acts on a lot of the F401 (unused import) warnings.

However, it really does highlight how a tool could have things go very badly.
```
./build/lib.linux-x86_64-2.7/pymavlink/mavexpression.py:12:1: F401 'math.*' imported but unused
./build/lib.linux-x86_64-2.7/pymavlink/mavexpression.py:13:1: F401 'mavextra.*' imported but unused
./build/lib.linux-x86_64-2.7/pymavlink/mavexpression.py:24:9: F401 'pymavlink.mavuser.*' imported but unused
```

Those are critical imports, as we eval() user-provided python expressions and pollute its namespace with commonly-used functions.  Fixing those warnings would mean finding a better way to do that pollution.

Could you have a look at my branch, and if you're happy with it I'll push it in.

https://github.com/ArduPilot/pymavlink/compare/master...peterbarker:patch-2?expand=1

Thanks!

peterbarker:
@cclauss I should also mention I picked F401 because fixing those does have a tangible advantage - reducing startup time.  Fixing those also often gives me the excuse to fix the multiple-imports-on-one-line warning.

Fixing the 'use "not in"' warning doesn't give any tangible advantage - we shouldn't increase the warning count for that one, but going through and fixing them all makes things like bisection more difficult.  If you have to change the relevant line for some other reason, you'd fix the warning.

peterbarker:
Merged in as part of the branch I referenced earlier.

Thanks!

PR #16, opened 2016-10-10
setup.py: add dependencies

khancyr:


khancyr:
This add the correct dependency for pip install ! I am not pip expert, but it should works . I also add python 3 as supported so pip install with python3 should work

tridge:
this fails for me with python2, as I have 3.4.2 of lxml installed and installing 3.6.4 has further requirements which aren't automatically satisfied.
I've added it without the version numbers for now, plus I've added it to the travis setup for the ardupilot build

PR #19, opened 2016-10-11
Python 3: long = int.  Plus flake8 recommended changes

cclauss:


DFReader.py
```diff
@@ -175,10 +188,8 @@ def find_time_base(self, gps, first_us_stamp):
 
     def type_has_good_TimeMS(self, type):
         '''The TimeMS in some messages is not from *our* clock!'''
-        if type.startswith('ACC'):
```
peterbarker: Not sure I agree with this one; it really is a switch-style construct.

If we did go down this path you could just return the comparison rather than have the "if"..

DFReader.py
```diff
@@ -11,9 +11,14 @@
 from builtins import range
 from builtins import object
 
-import struct, time, os
+import struct
 from . import mavutil
 
+try:
```
peterbarker: Are we going to have to carry this forever-more?

There's not future thing we can use to fix this instead?

DFReader.py
```diff
@@ -59,7 +65,8 @@ def __init__(self, type, name, flen, format, columns):
                 msg_mults.append(mul)
                 msg_types.append(type)
             except KeyError as e:
-                raise Exception("Unsupported format char: '%s' in message %s" % (c, name))
```
peterbarker: Stylistically I'm not in favour.  I'd rather have the longer line than create a variable, regardless of what pep says :-)

I'd also like to see the Exception narrowed to something else - ValueError if nothing else...

DFReader.py
```diff
@@ -497,24 +514,23 @@ def check_condition(self, condition):
...
 
+
 class DFReader_binary(DFReader):
     '''parse a binary dataflash file'''
     def __init__(self, filename, zero_time_base=False):
         DFReader.__init__(self)
         # read the whole file into memory for simplicity
-        f = open(filename, mode='rb')
-        self.data = f.read()
+        with open(filename, mode='rb') as f:
```
peterbarker: Why not `self.data = open(filename, mode='rb').read()`?

DFReader.py
```diff
@@ -547,13 +565,14 @@ def _parse_next(self):
         if skip_bytes != 0:
             if self.remaining < 528:
                 return None
-            print("Skipped %u bad bytes in log at offset %u, type=%s" % (skip_bytes, skip_start, skip_type))
```
peterbarker: Again, not in favour of polluting namespaces just to keep PEP happy....

DFReader.py
```diff
@@ -574,44 +593,45 @@ def _parse_next(self):
...
         return m
 
+
 def DFReader_is_text_log(filename):
     '''return True if a file appears to be a valid text log'''
-    f = open(filename)
-    ret = (f.read(8000).find('FMT, ') != -1)
-    f.close()
-    return ret
+    with open(filename) as f:
```
peterbarker: `return 'FMT, ' in open(filename).read(8000)`?

DFReader.py
```diff
@@ -574,44 +593,45 @@ def _parse_next(self):
...
     def __init__(self, filename, zero_time_base=False):
         DFReader.__init__(self)
         # read the whole file into memory for simplicity
-        f = open(filename, mode='r')
-        self.lines = f.readlines()
-        f.close()
-        self.formats = {
-            'FMT' : DFFormat(0x80, 'FMT', 89, 'BBnNZ', "Type,Length,Name,Format,Columns")
-        }
+        with open(filename, mode='r') as f:
```
peterbarker: Again :-)

DFReader.py
```diff
@@ -256,16 +279,10 @@ def rewind_event(self):
 
     def message_arrived(self, m):
         type = m.get_type()
-        if not type in self.counts:
-            self.counts[type] = 1
-        else:
-            self.counts[type] += 1
+        self.counts[type] = self.counts(type, 0) + 1
```
peterbarker: This doesn't actually work :-)

Were you thinking of using an accessor into the dict here?

DFReader.py
```diff
@@ -256,16 +279,10 @@ def rewind_event(self):
...
         # this preserves existing behaviour - but should we be doing this
         # if type == 'GPS'?
-        if not type in self.counts_since_gps:
-            self.counts_since_gps[type] = 1
-        else:
-            self.counts_since_gps[type] += 1
+        self.counts_since_gps[type] += self.counts_since_gps(type, 0) + 1
```
peterbarker: This also doesn't work :-)

DFReader.py
cclauss: The Python specification does not guaranteed your file will be closed()...  Dangling resource plus chance that file writes are not flushed.  The results could be different between CPython, Pypy, Jython, etc.  `with` does the right thing on all implementations.

DFReader.py
cclauss: The Python specification does not guaranteed your file will be closed()...  Dangling resource plus chance that file writes are not flushed.  The results could be different between CPython, Pypy, Jython, etc.  `with` does the right thing on all implementations.

DFReader.py
cclauss: The Python specification does not guaranteed your file will be closed()...  Dangling resource plus chance that file writes are not flushed.  The results could be different between CPython, Pypy, Jython, etc.  `with` does the right thing on all implementations.

DFReader.py
cclauss: I was trying to avoid `future`wherever possible.

DFReader.py
cclauss: I am OK with long lines it that is the way to go.

I too have great worries about unqualified exceptions... https://realpython.com/blog/python/the-most-diabolical-python-antipattern/

DFReader.py
cclauss: I picked the latter but perhaps you prefer the former?

DFReader.py
cclauss: Good catch.  Modified.

DFReader.py
cclauss: Good catch.  Modified.

DFReader.py
cclauss: http://stackoverflow.com/questions/7395542/is-explicitly-closing-files-important

peterbarker:
On Tue, 11 Oct 2016, cclauss wrote:

> The Python specification does not guaranteed your file will be closed()...
> Dangling resource plus chance that file writes are not flushed. The results
> could be different between CPython, Pypy, Jython, etc. with does the right
> thing on all implementations.

We're reading each of these files, and I'd be willing to burn a filehandle 
or two until the GC got around to fixing things :-)

I expect best practise would be to close explicitly, as you've done.  If 
that's what you'd like, that's fine.

DFReader.py
cclauss: Reraising a KeyError

DFReader.py
cclauss: Undone.

PR #20, opened 2016-10-11
Move flake8 options into a .flake8 file

cclauss:


peterbarker:
Why hide this in a dotfile?  We'd like people to be able to discover it easily :-)

Could just as easily go into the "files" subdirectory.

cclauss:
@peterbarker Please read through http://flake8.readthedocs.io/en/latest/user/configuration.html#configuration-locations .  We can have the flake8 configuration in `./setup.cfg`, `./tox.ini`, or `./.flake8`.  I am 100% flexible to track your suggestions but...  I dislike setup.cfg because it feels too much like setup.py and yet it is completely different.  I dislike tox.ini because we are not running tox.

Also, please help me to understand how flake8 can be modified to look in ./files to find its configuration data.

peterbarker:
The flake8 command-line tool takes a "--config" option to specify a config file to use.
flake8 --config files/bob DFReader.py

I don't suggest "bob" as a config file name...

Note that the flake8 config file this PR seeks to create doesn't work for me; the "select" items must be separated by commas AFAICS.

peterbarker:
OK, so flake8 will actually walk up directories until it finds a config file - that's unexpected!

I think in that case .flake8 is fine as a filename.  Still need to fix the format in this PR, 'though...

peterbarker:
Also, a note on actually fixing the problems.

I'd like to take a "no worse" approach to using flake8-raised issues, rather than wholesale-fixing problems.  Any new commits to pymavlink should make the flake8 output "no worse" than it was - and hopefully improve it!

Wholesale fixing problems (like "E713 test for membership should be 'not in'") means code-churn, and obscures the true history of a file.

Some ("F401 'os' imported but unused") can often be fixed without stuffing this history up.

cclauss:
I moved these changes into PR https://github.com/ArduPilot/pymavlink/pull/15

PR #22, opened 2016-10-11
Add #include <stdbool.h>

CarloWood:
This header is used by C99 source files, which do not
necessarily include &lt;stdbool.h&gt; before including this
header. stdbool.h is needed to define the type 'bool'
that is used in this header. See
https://en.wikibooks.org/wiki/C_Programming/C_Reference/stdbool.h

I need this to be added for the PX4/Firmware project which
doesn't compile anymore after reordering headers (aka, not
having a #include &lt;stdbool.h&gt; anymore before including
this header).

### DIFF
--- a/generator/C/include_v2.0/mavlink_types.h
+++ b/generator/C/include_v2.0/mavlink_types.h
@@ -5,6 +5,7 @@
 #error "The C-MAVLink implementation requires Visual Studio 2010 or greater"
 #endif
 
+#include <stdbool.h>
 #include <stdint.h>
 
 // Macro to define packed structures


CarloWood:
@tridge Someone said you're the guy I need to apply this ;). Please let me know if there's a problem?

khancyr:
it isn't need for mavlink v1.0 ?

magicrub:
>I need this to be added for the PX4/Firmware project 

Then add it to that project instead of this one.

CarloWood:
@khancyr No, the usage of 'bool' in this header appeared only in version 2.

PR #27, opened 2016-10-21
MAVLink2 C++11 library generator

vooon:
This PR is rebased and moved https://github.com/mavlink/mavlink/pull/568 .

What is missed - Travis tests.

Ping @tridge


generator/CPP11/include_v2.0/msgmap.hpp
```diff
@@ -0,0 +1,236 @@
...
+
+	//std::cout << "M< s: " << sizeof(_T) << " p: " << pos << " d: " << data << std::endl;
+
+	pos += sizeof(_T);
+}
+
+template<class _T, size_t _Size>
+void mavlink::MsgMap::operator<< (const std::array<_T, _Size> &data)
+{
+	for (auto &v : data)
```
khancyr: I don't think there is a fixed code style, but I far prefer to always have parenthesis even in single line statement. I prevent many error now and in the future !

generator/CPP11/include_v2.0/msgmap.hpp
```diff
@@ -0,0 +1,236 @@
...
+
+	case 8:
+		buf = le64toh(buf);
+		break;
+
+	default:
+		assert(false);
+	}
+
+	if (std::is_floating_point<_Tout>::value)
```
khancyr: parenthesis

generator/CPP11/include_v2.0/msgmap.hpp
```diff
@@ -0,0 +1,236 @@
...
+
+	//std::cout << "M> s: " << sizeof(_T) << " p: " << pos << " d: " << data << std::endl;
+
+	pos += sizeof(_T);
+}
+
+template<class _T, size_t _Size>
+void mavlink::MsgMap::operator>> (std::array<_T, _Size> &data)
+{
+	for (auto &v : data)
```
khancyr: parenthesis

khancyr:
If it is the same version used in ROS package for some times already, I think it is good to go !

vooon:
Yes, this is same code, just rebased and moved to this repo.

vooon:
Ping @tridge, @peterbarker. Is there a problem that not allow merge?

peterbarker:
On Fri, 18 Nov 2016, Vladimir Ermakov wrote:

> Ping @tridge, @peterbarker. Is there a problem that not allow merge?

Sorry, just haven't made time to look at it.  Thanks, we will look!

Peter

tridge:
@vooon sorry for the slow response on this one.
My main concern with this PR is that it has a potential impact on the current C generator, especially when used in C++ programs.
What I'd like to see is that the C++ interface have its own headers to wrap the C ones. Maybe make them .hpp files, and add all the needed namespace guards there. The hpp file could #include the existing C headers.
That will make me much more confident that we won't be breaking any of the existing users. Does that make sense?

tridge:
actually, I've changed my mind. The header wrapping is actually fine, I was just being silly.
I've pushed this to master, with changes to the test suite so that ./test_generator.sh tests the C++ as well as the C interface. That lowers the chance I will break this in the future.
thanks very much for your work on this!

tridge:
I also changed it to auto-run the C generator for the C headers, so the user doesn't have to run that separately

vooon:
@tridge thank you!

Generation of C headers handled by build system, that's why i don't generate C lib at C++ gen.
But i think it is ok, and helps for others.

I don't remember does run_tests fail if test fails.

PR #31, opened 2016-11-02
Payload field of struct is now correctly labeled within the Wireshark filter string

courtarro:
Before, both the CRC and Payload fields were ".crc" in the Wireshark filter string definition. Payload is now correctly named ".payload".


peterbarker:
Thanks @courtarro, merged.

PR #32, opened 2016-11-04
mavutil.py: OFFBOARD mode interpretation fix

bkakilli:
OFFBOARD mode is interpreted as UNKNOWN.
Changing custom_sub_mode to custom_main_mode in line 1597 solves this issue.

### DIFF
--- a/mavutil.py
+++ b/mavutil.py
@@ -1594,7 +1594,7 @@ def interpret_px4_mode(base_mode, custom_mode):
                 return "LAND"
             elif custom_sub_mode == PX4_CUSTOM_SUB_MODE_AUTO_RTGS:
                 return "RTGS"
-            elif custom_sub_mode == PX4_CUSTOM_MAIN_MODE_OFFBOARD:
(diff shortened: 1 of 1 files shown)


peterbarker:
I'm not sure that the check for this mode shouldn't go at the top of the `
 elif (base_mode & auto_mode_flags) == auto_mode_flags: #auto modes` block.

It seems odd that you'd have this comparison against the main mode where every other elif is checkout against teh submode.  I've only the loosest of understandings of how PX4 modes work, mind you....

@squilter Does this look right to you?  If not "right", will it do?

bkakilli:
I checked the binary sequence of _base_mode_ and _control_mode of OFFBOARD mode_. auto_mode_flags = 00011100 and the _base_mode_ for OFFBOARD is 10011101. Therefore it passes that check you mentioned. And yeah, the comparison seems not consistent.

squilter:
Yes, this looks correct.

peterbarker:
Merged, thanks @bkakilli and @squilter

PR #33, opened 2016-11-07
setup.py: generate file during install instead of before

khancyr:


peterbarker:
@khancyr Could you fold that correction back into where it is supposed to go, please?

peterbarker:
Sadly,  doesn't fix the "future" issue for me:

```
vagrant@vagrant-ubuntu-trusty-64:~/pymavlink$ git checkout khancry/setup
Note: checking out 'khancry/setup'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by performing another checkout.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -b with the checkout command again. Example:

  git checkout -b new_branch_name

HEAD is now at e03abd1... setup.py: use generator in setup function
vagrant@vagrant-ubuntu-trusty-64:~/pymavlink$ python setup.py build install --user --force
running build
running build_py
  File "setup.py", line 154, in <module>
    ext_modules = extensions
  File "/usr/lib/python2.7/distutils/core.py", line 151, in setup
    dist.run_commands()
  File "/usr/lib/python2.7/distutils/dist.py", line 953, in run_commands
    self.run_command(cmd)
  File "/usr/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "/usr/lib/python2.7/distutils/command/build.py", line 128, in run
    self.run_command(cmd_name)
  File "/usr/lib/python2.7/distutils/cmd.py", line 326, in run_command
    self.distribution.run_command(command)
  File "/usr/lib/python2.7/distutils/dist.py", line 972, in run_command
    cmd_obj.run()
  File "setup.py", line 94, in run
    generate_content()
  File "setup.py", line 22, in generate_content
    from generator import mavgen, mavparse
  File "/home/vagrant/pymavlink/generator/mavgen.py", line 12, in <module>
    from future import standard_library
ImportError: No module named future
```

Still a\ nice cleanup, 'though....

khancyr:
@peterbaker use pip install .  --user - v

OXINARF:
If the setup_requires is added it will work with @peterbarker's command. But it will still not install the dependencies for runtime, only for setup. And if lxml is added in the setup_requires you will need to install the packages needed for compilation.

PR #39, opened 2016-11-22
mavutil: add force-connect option

peterbarker:
This will allow sim_vehicle.py to regain gdb_stopped.

It also allows you to start mavproxy without a named serial device attached, and then connect that device.


khancyr:
@peterbarker Ok, I found one error, I try with a 3dr radio and now it don't crash if not plugged, but when I plug it does nothing. 
It appear that autoreconnect is not implemented

khancyr:
It appear that autoreconnect is try only once when the first write failed. mavproxy should keep trying to write until it got it working.

Otherwise, instead of using a fakeserial, we could have use https://pythonhosted.org/pyserial/shortintro.html#configuring-ports-later and try to open the port only if it appear is port list

amilcarlucas:
Needs a rebase

PR #42, opened 2016-12-27
Fix PX4 mode bug

sanderux:


sanderux:
@tridge could you have a look at this?

mavutil.py
```diff
@@ -1580,9 +1580,9 @@ def interpret_px4_mode(base_mode, custom_mode):
             return "POSCTL"
     elif (base_mode & auto_mode_flags) == auto_mode_flags: #auto modes
         if custom_main_mode & PX4_CUSTOM_MAIN_MODE_AUTO != 0:
-            if custom_sub_mode == PX4_CUSTOM_SUB_MODE_AUTO_MISSION:
+	    if custom_sub_mode == PX4_CUSTOM_SUB_MODE_AUTO_MISSION:
```
lucasdemarchi: your fix here is wrong... you are adding tabs instead of using only spaces

lucasdemarchi:
You are adding tabs on the first commit and "fixing" it on the second by adding more.  Python sources here use only spaces. I fixed that and applied the first commit.

sanderux:
@lucasdemarchi yeah i noticed QT was being a bit difficult replacing them at save. thanks!

sanderux:
@lucasdemarchi any plans for a version bump and a pypi update? we would like to get dronekit compatible  but this issue makes it a bit hard t work with.

PR #47, opened 2017-01-12
Add optional units and bitmask fields to message definitions

amilcarlucas:
This will simplify machine parsing and presentation of data

### DIFF
--- a/generator/mavparse.py
+++ b/generator/mavparse.py
@@ -33,12 +33,14 @@ class MAVParseError(Exception):
         return self.message
 
 class MAVField(object):
-    def __init__(self, name, type, print_format, xml, description='', enum=''):
+    def __init__(self, name, type, print_format, xml, description='', enum='', bitmask='', units=''):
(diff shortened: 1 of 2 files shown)


OXINARF:
This is a duplicate of #7

OXINARF:
Actually this puts it on fields, instead of just command params. But I think it is just a small extension of #7 and should be discussed there.

amilcarlucas:
No, this is not duplicate. This one is for all messages, not just for MAV_CMD. Please reopen.

magicrub:
This is blocking https://github.com/mavlink/mavlink/pull/663

OXINARF:
@amilcarlucas Like I said, this adds it to fields, #7 to command params. But the idea is the same and #7 has much more attributes. What would be the reason to keep two PRs open with the same idea?

tridge:
i will look at merging the two PRs

tridge:
merged, thanks!

PR #51, opened 2017-02-04
Python 2 has long but Python 3 does not

cclauss:
Also used https://docs.python.org/2/library/collections.html#collections.Counter

Flake8 suggestions.

### DIFF
--- a/DFReader.py
+++ b/DFReader.py
@@ -10,6 +10,7 @@ Partly based on SDLog2Parser by Anton Babushkin
 from __future__ import print_function
 from builtins import range
 from builtins import object
+from collections import Counter
 
 import struct
 from . import mavutil
@@ -64,7 +65,7 @@ class DFFormat(object):
                 msg_struct += s
                 msg_mults.append(mul)
                 msg_types.append(type)
-            except KeyError as e:
+            except KeyError:
                 raise Exception("Unsupported format char: '%s' in message %s" % (c, name))
 
         self.msg_struct = msg_struct
@@ -212,8 +213,8 @@ class DFReaderClock_usec(DFReaderClock):
 
 
 class DFReaderClock_msec(DFReaderClock):
-    '''DFReaderClock_msec - a format where many messages have TimeMS in their formats,
-       and GPS messages have a "T" field giving msecs '''
+    '''DFReaderClock_msec - a format where many messages have TimeMS in their
+       formats, and GPS messages have a "T" field giving msecs '''
     def find_time_base(self, gps, first_ms_stamp):
         '''work out time basis for the log - new style'''
         t = self._gpsTimeToTime(gps.Week, gps.TimeMS)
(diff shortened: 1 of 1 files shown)


DFReader.py
```diff
@@ -175,10 +186,8 @@ def find_time_base(self, gps, first_us_stamp):
 
     def type_has_good_TimeMS(self, type):
```
peterbarker: I prefer having multiple if statements here.  Once you understand the pattern you can glance down the list of values to see what cases are covered, which is nice.  Assuming this isn't a flake8 change I'd prefer not to have this patch,

DFReader.py
```diff
@@ -363,40 +372,38 @@ def init_clock(self):
...
-                break;
+                break
 
             type = m.get_type()
 
-            if first_us_stamp is None:
-                first_us_stamp = getattr(m, "TimeUS", None);
-
-            if first_ms_stamp is None and (type != 'GPS' and type != 'GPS2'):
+            first_us_stamp = first_us_stamp or getattr(m, "TimeUS", None)
```
peterbarker: This is a functional change.  The original check is for None, this is for truth.

Further, the separate "if" blocks are nicer in Python.  And this is coming from a Perl programmer who really likes `//=`

DFReader.py
```diff
@@ -552,17 +560,17 @@ def _parse_next(self):
...
         self.remaining -= 3
 
-        if not msg_type in self.formats:
+        if msg_type not in self.formats:
             if self.verbose:
                 print("unknown message type %02x" % msg_type)
             raise Exception("Unknown message type %02x" % msg_type)
         fmt = self.formats[msg_type]
-        if self.remaining < fmt.len-3:
+        if self.remaining < fmt.len - 3:
```
peterbarker: The added whitespace here makes things less clear, not more clear.

DFReader.py
```diff
@@ -552,17 +560,17 @@ def _parse_next(self):
...
             # out of data - can often happen half way through a message
             if self.verbose:
                 print("out of data")
             return None
-        body = self.data[self.offset:self.offset+(fmt.len-3)]
+        body = self.data[self.offset:self.offset + (fmt.len - 3)]
```
peterbarker: Again, added whitespace makes things less clear.

DFReader.py
```diff
@@ -584,21 +592,19 @@ def _parse_next(self):
                                                  null_term(elements[2]), elements[1],
                                                  null_term(elements[3]), null_term(elements[4]))
 
-        self.offset += fmt.len-3
-        self.remaining -= fmt.len-3
+        self.offset += fmt.len - 3
```
peterbarker: Also less clear.

DFReader.py
```diff
@@ -689,17 +692,13 @@ def _parse_next(self):
         profiler.add_function(DFReader_binary._add_msg)
         profiler.add_function(DFReader._set_time)
         profiler.enable_by_count()
-                    
+
     filename = sys.argv[1]
-    if filename.endswith('.log'):
```
peterbarker: This change makes things less clear.

Further, considering I'm vaguely working on a DFReader_CSV, this would need to be reverted later.

cclauss: How are things coming on this rewrite?  The [current version of DFReader.py fails](https://github.com/ArduPilot/pymavlink/pull/55/files#diff-354f30a63fb0907d4ad57269548329e3R50) on all versions of Python.

PR #58, opened 2017-03-23
WIP: add dropout check to fft plotter

kd0aij:
we're regularly dropping every 10th accel sample. Need to check whether this is happening at the sensor, or just in the log:
```
sample range:  4382 38370
N samples:  33988
ndeltas:  33987
duration: 37.785 seconds
average sample rate: 899 Hz
mean sample interval:  1.11 msec
std:  342 usec
```
dropping_samples

### DIFF
--- a/tools/mavfft_int.py
+++ b/tools/mavfft_int.py
@@ -25,10 +25,10 @@ def fft(logfile):
     print("Processing log %s" % filename)
     mlog = mavutil.mavlink_connection(filename)
 
-    data = {'ACC1.rate' : 900,
+    data = {'ACC1.rate' : 1000,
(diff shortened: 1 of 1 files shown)


kd0aij:
resampling using numpy.interp, appears to have added harmonic distortion 
resampled

kd0aij:
this version assumes the SampleUS fields in ACCn and GYRn log records are sequential sample counters representing the total number of samples received from the sensor in fast_sampling mode since the last fifo error or reset. It also computes the sample rate by comparing SampleUS to the timestamps (TimeUS)

kd0aij:
the hardwired sample rates don't look right for fmu-v4

PR #59, opened 2017-03-24
Add a file attribute to the XML schema

WickedShell:
A file attribute would allow easy tracking of the include hierachy when parsing
from any form that doesn't come with the correct file name.
(InputStreams for example, where you can't track what the original name was)

### DIFF
--- a/generator/mavschema.xsd
+++ b/generator/mavschema.xsd
@@ -146,6 +146,7 @@
             <xs:element ref="enums" minOccurs="0"/>
             <xs:element ref="messages"/>
         </xs:sequence>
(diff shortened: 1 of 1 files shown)


OXINARF:
@WickedShell I figured out last night that this change broke XML validation. It has a reference to _file_ but *file* isn't defined anywhere.

I can submit a PR to fix it, but I don't really understand what is its purpose nor I'm completely sure what type it is supposed to have. I think what you wanted here was `<xs:attribute name="file" type="xs:anyURI"/>` but I'd like you to confirm.

PR #64, opened 2017-03-29
generator/C: fix mavlink v2.0 parser bug

julianoes:
This fixes a parser bug in the mavlink v2.0 parser where we would get a
message flagged as valid when actually the CRC check failed.

The problem was that status->msg_received which was set to
MAVLINK_FRAMING_BAD_CRC got overwritten in the later signing steps.

Found with @bkueng. 

Edit:

Even though signing was not used, what happens is that `rxmsg->incompat_flags & MAVLINK_IFLAG_SIGNED` can be true because it is essentially garbage.

### DIFF
--- a/generator/C/include_v2.0/mavlink_helpers.h
+++ b/generator/C/include_v2.0/mavlink_helpers.h
@@ -735,18 +735,27 @@ MAVLINK_HELPER uint8_t mavlink_frame_char_buffer(mavlink_message_t* rxmsg,
 		} else {
 			// Successfully got message
 			status->msg_received = MAVLINK_FRAMING_OK;
-                }
-                rxmsg->ck[1] = c;
+		}
+        rxmsg->ck[1] = c;
+
 		if (rxmsg->incompat_flags & MAVLINK_IFLAG_SIGNED) {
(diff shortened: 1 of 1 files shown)


generator/C/include_v2.0/mavlink_helpers.h
```diff
@@ -735,18 +735,27 @@ MAVLINK_HELPER uint8_t mavlink_frame_char_buffer(mavlink_message_t* rxmsg,
 		} else {
 			// Successfully got message
 			status->msg_received = MAVLINK_FRAMING_OK;
-                }
-                rxmsg->ck[1] = c;
+		}
+        rxmsg->ck[1] = c;
```
khancyr: non important issue but the indentation isn't right

khancyr:
well ... I didn't see it was using tabs ... I will correct that in another PR !

julianoes:
@khancyr thanks. I tried to adapt to the majority of the rest of the file :smile:. I fixed the lines around my change but it makes sense to do another PR to fix all whitespace.

julianoes:
@tridge could you please review this?

tridge:
thanks @julianoes, very good find!
pushed to master

PR #65, opened 2017-03-29
Compilation bugfix for certain compilers: array initialization with n…

mazahner:
Compilation bugfix for certain compilers: array initialization with non-const variable

initializing an array with a non const variable is
1. dangerous
2. is rejected by some compilers


tridge:
looks good, thanks!

PR #72, opened 2017-04-23
Fix mavschema.xsd

OXINARF:
As I wrote in https://github.com/ArduPilot/pymavlink/pull/59#issuecomment-296444133, that change broke the XML validation. It added an attribute referencing a type named file, but none existed. This changes it to the intended line.


peterbarker:
Merged, thanks!

PR #73, opened 2017-05-03
A GUI tool that can be used to calculate telemetry link datarates

amilcarlucas:
The mavlink messages size is autogenerated, I can add the generation code later, if requested.
It works with both Python2 and 3


gmorph:
I've run it.  Interesting tool.  You can select which messages are sent in mavlink and thus calculate how many bits/sec are required.  For instance you can include the RANGEFINDER message if you have a rangefinder but if not you can leave it out.
The HEARTBEAT message didn't seem to show up in the GUI.  Ah I see how you gathered the messages.  So there are extra messages we need to add in.  It would be good when we populated a NONE field with a value if another NONE field appeared below.  But we can always change change the python code easily enough as well.

So, an interesting tool that I think we should merge in as its standalone.

I still think what would be interesting is a tool that analyses a tlog file and determines the data rate from that.  That would be a real world data value.

Thanks, Grant.

tridge:
looks worthwhile, thanks!
could we put them in the tools/ directory, instead of top level?

amilcarlucas:
@gmorph I had the same idea as you regarding populating the a new None combobox, but never got to implement it.
One other nice feature would be a combobox on top to select the vehicle :) 

@tridge I've bee thinking about this and maybee is better to put this one in the ardupilot/Tools directory.
It does have a lot of ardupilot specific code in there.

mavlink_messages_size.py
```diff
@@ -0,0 +1,134 @@
+from collections import OrderedDict
+
+mavlink_message_lengths_dict = OrderedDict([
```
peterbarker: Should be able to get these message lengths with a little bit of introspection on pymavlink?

telemetry_datarates.py
```diff
@@ -0,0 +1,193 @@
...
+from tkinter import Tk, Text, TOP, BOTH, X, Y, N, LEFT,RIGHT, CENTER, RIDGE, VERTICAL, END, IntVar, IntVar, Scrollbar
+from tkinter.ttk import Frame, Label, Entry, Combobox, Checkbutton
+
+# import MAVLink messages length dictionary
+# this file is generated by a script
+mav_message_size = 'mavlink_messages_size.py'
+exec(compile(source=open(mav_message_size).read(), filename=mav_message_size, mode='exec'))
+
+# Please adapt this list to the vehicle you are using
+# These were done by looking at the GCS_MAVLINK_Copter::data_stream_send() function
```
peterbarker: Obviously having the other vehicles would be awesome.

telemetry_datarates.py
```diff
@@ -0,0 +1,193 @@
+#!/usr/bin/env python
```
peterbarker: This file lacks the execute bit.

peterbarker: .... and should probably be in the tools directory...

It should also possibly be renamed to "pytelemetry_rates.py" or something so it can be installed :-)

amilcarlucas:
Should I merge this or are we waiting for more reviews ?
BWT this works with both Python2 and Python3

gmorph:
Merge away I say!

PR #81, opened 2017-05-29
enable use of multiple config files

matanhavi:
by using comma sepereted file list as argument


matanhavi:
By using <include> on one file the request above is solved. so there is no need for the proposed commit

PR #84, opened 2017-06-01
Fix the mavgen_c import in mavgen_cpp11.py

kartikmohta:


cclauss:
@kartikmohta Did you test this change under both Python 2 and Python 3?

kartikmohta:
No, I tested it only on Python 3 but this is how it's done in the mavgen.py script (https://github.com/ArduPilot/pymavlink/blob/master/generator/mavgen.py#L132) so should be good, right?

kartikmohta:
For reference, see http://python3porting.com/differences.html#imports

cclauss:
@kartikmohta Thanks.  That is a great explanation.

kartikmohta:
Any issues holding up merging this?

peterbarker:
Merged, thanks!

PR #85, opened 2017-06-02
Update mode mappings for ArduSub

jaxxzer:


peterbarker:
Removing these would stop us being able to easily test older versions of sub with this pymavlinlk.  Also, reusing mode numbers may lead to fun-and-games.

Copter's mode mapping still has modes in it that were removed, I guess for pretty much these reasons.

You're the boss when it comes to Sub stuff, 'though - so if you want this in as-is it goes in... but did you want to leave in the modes that were once-implemented but have now been removed?

jaxxzer:
@peterbarker Thank you for the valid considerations. These modes were never implemented (some were, but were just experimental) and should never have been used in the first place. It is my intention to make sure we don't reuse these mode numbers in the future.

jaxxzer:
It's good to go as-is from my end.

peterbarker:
Merged!

PR #86, opened 2017-06-19
fix macros for compiling in MDK5 with armcc

matthewzhenggong:
Inspired by https://github.com/oceanstack/MAVLink_MDK_STM32, adjust some macro defines for armcc compiler in Kail uVision5 embed-system develpment environment.

### DIFF
--- a/generator/C/include_v2.0/mavlink_types.h
+++ b/generator/C/include_v2.0/mavlink_types.h
@@ -15,6 +15,10 @@ namespace mavlink {
 // Macro to define packed structures
 #ifdef __GNUC__
   #define MAVPACKED( __Declaration__ ) __Declaration__ __attribute__((packed))
+#elif defined __arm__
+  #pragma anon_unions
+  #pragma pack(push, 1)
+  #define MAVPACKED( __Declaration__ ) __Declaration__	
 #else
   #define MAVPACKED( __Declaration__ ) __pragma( pack(push, 1) ) __Declaration__ __pragma( pack(pop) )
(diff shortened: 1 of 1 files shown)


generator/C/include_v2.0/mavlink_types.h
```diff
@@ -15,6 +15,10 @@ namespace mavlink {
 // Macro to define packed structures
 #ifdef __GNUC__
   #define MAVPACKED( __Declaration__ ) __Declaration__ __attribute__((packed))
+#elif defined __arm__
```
peterbarker: Could you find a tighter constraint on the check here?

Does `__UVISION_VERSION` work?

matthewzhenggong: default
Yes. I tried, and it works.

matthewzhenggong: Could we use the macro __ARMCC_VERSION instead, which is the predefined macro for ARM KEIL compiler, while __UVISION_VERSION is related with MDK IDE environment.

peterbarker:
On Mon, 19 Jun 2017, Zheng GONG wrote:

> Could we use the macro __ARMCC_VERSION instead, which is the predefined
> macro for ARM KEIL compiler, while __UVISION_VERSION is related with MDK IDE
> environment.

So long as nobody else produces a compiler called "armcc" we're probably 
good with that :-)

Linjieqiang:
Tested it just now. Works well.

PR #87, opened 2017-06-27
mavutil: handle ONBOARD_CONTROLLER heartbeat as GCS one's

khancyr:
solve issue with companion computer that are spam with : 
Mode(0x000000c0)> STABILIZE> Mode(0x000000c0)> STABILIZE> Mode(0x000000c0)

See : https://github.com/mavlink/mavros/issues/646 and https://gitter.im/ArduPilot/MAVProxy?at=594a92586462d8493c36fbb8

### DIFF
--- a/mavutil.py
+++ b/mavutil.py
@@ -229,7 +229,7 @@ class mavfile(object):
         msg._posted = True
         msg._timestamp = time.time()
         type = msg.get_type()
-        if type != 'HEARTBEAT' or (msg.type != mavlink.MAV_TYPE_GCS and msg.type != mavlink.MAV_TYPE_GIMBAL):
+        if type != 'HEARTBEAT' or msg.type not in (mavlink.MAV_TYPE_GCS, mavlink.MAV_TYPE_ONBOARD_CONTROLLER, mavlink.MAV_TYPE_GIMBAL):
             self.messages[type] = msg
 
         if 'usec' in msg.__dict__:
@@ -265,7 +265,7 @@ class mavfile(object):
         if type == 'HEARTBEAT' and msg.get_srcComponent() != mavlink.MAV_COMP_ID_GIMBAL:
             self.target_system = msg.get_srcSystem()
             self.target_component = msg.get_srcComponent()
-            if float(mavlink.WIRE_PROTOCOL_VERSION) >= 1 and msg.type != mavlink.MAV_TYPE_GCS:
+            if float(mavlink.WIRE_PROTOCOL_VERSION) >= 1 and msg.type not in (mavlink.MAV_TYPE_GCS, mavlink.MAV_TYPE_ONBOARD_CONTROLLER):
                 self.flightmode = mode_string_v10(msg)
                 self.mav_type = msg.type
                 self.base_mode = msg.base_mode


mavutil.py
```diff
@@ -228,7 +228,7 @@ def post_message(self, msg):
         msg._posted = True
         msg._timestamp = time.time()
         type = msg.get_type()
-        if type != 'HEARTBEAT' or (msg.type != mavlink.MAV_TYPE_GCS and msg.type != mavlink.MAV_TYPE_GIMBAL):
+        if type != 'HEARTBEAT' or (msg.type != mavlink.MAV_TYPE_GCS and msg.type != mavlink.MAV_TYPE_ONBOARD_CONTROLLER and msg.type != mavlink.MAV_TYPE_GIMBAL):
```
cclauss: if type != 'HEARTBEAT' or msg.type not in (mavlink.MAV_TYPE_GCS, mavlink.MAV_TYPE_ONBOARD_CONTROLLER, mavlink.MAV_TYPE_GIMBAL):

cclauss: also `type()` is a builtin function in Python so `msg_type` might be a better variable name

khancyr: yep better thanks

mavutil.py
```diff
@@ -264,7 +264,7 @@ def post_message(self, msg):
         if type == 'HEARTBEAT' and msg.get_srcComponent() != mavlink.MAV_COMP_ID_GIMBAL:
             self.target_system = msg.get_srcSystem()
             self.target_component = msg.get_srcComponent()
-            if float(mavlink.WIRE_PROTOCOL_VERSION) >= 1 and msg.type != mavlink.MAV_TYPE_GCS:
+            if float(mavlink.WIRE_PROTOCOL_VERSION) >= 1 and msg.type != mavlink.MAV_TYPE_GCS and msg.type != mavlink.MAV_TYPE_ONBOARD_CONTROLLER:
```
cclauss: if float(mavlink.WIRE_PROTOCOL_VERSION) >= 1 and msg.type not it (mavlink.MAV_TYPE_GCS, mavlink.MAV_TYPE_ONBOARD_CONTROLLER):

peterbarker:
I wonder why I don't get these more often when running companion computers here... what's the setup routing-wise in all of this?

We should contain this nastiness to a function - `is_our_vehicle_heartbeat(msg)`, perhaps?  The two cases appear to be essentially the same, leaving just a check for is-mavlink>=1.0 in the second case.  Something like this: https://github.com/ArduPilot/pymavlink/compare/master...peterbarker:heartbeat-changes?expand=1 Sadly, this is a logic change - comparison of component-id rather than type, and also when we set our target sysid/compid.

My concern with adding companion computer - we're playing whack-a-mole.  Perhaps we should have a whitelist of MAV_TYPEs and component IDs in that new function I created?

In any case, this is still bad.  We're special-casing heartbeats for special treatment, pretty much just to avoid the mode switching issues.  All sorts of whackiness will happen with other parameters - notably param_values.  Thus mavsource: https://github.com/peterbarker/pymavlink/tree/mavsource

khancyr:
I think the default behaviour should be like on ardupilot, we only use the packet from our target, other are only passed to mavlink routing !

khancyr:
@peterbarker , Could we push this as temp solution ?

khancyr:
closed by https://github.com/ArduPilot/pymavlink/pull/144

PR #88, opened 2017-07-05
Update telemetry streams for copter

amilcarlucas:
Adds a couple of new streams
Disables some streams that are rarely used


rmackay9:
I probably overstepped my bounds on this one 'cuz I know so little about pymavlink but.. I suspect it's ok because I hear from amilcarlucas that he's essentially the owner of this file.

PR #89, opened 2017-07-05
Thread save signature check

simonegu:
@tridge I'm trying to port MAVLink signing to PX4, from my analysis here: https://docs.google.com/drawings/d/1VVA5qY_JXMPEm4Af2SJMrxo4cz9acLq8EMkuB2xFVrw/edit?usp=sharing
It seems that a locking mechanism is needed because of possible concurrent access to the `signing_streams` struct.
I based my proposition (98be05e) on the previous work done for the UART locking here: https://github.com/ArduPilot/pymavlink/blob/master/generator/C/include_v2.0/mavlink_helpers.h#L348 in relation with this: https://github.com/PX4/Firmware/blob/master/src/modules/mavlink/mavlink_bridge_header.h#L50-L51


simonegu:
ported to https://github.com/mavlink/pymavlink/pull/1

PR #94, opened 2017-07-27
Fixed c# compilation status: detects compile failure & raises exception.

mday299:
This patch addresses an issue where when C# compiles fails the developer is unaware of the failure.

### DIFF
--- a/generator/mavgen_cs.py
+++ b/generator/mavgen_cs.py
@@ -341,7 +341,7 @@ using System.Reflection;
     #print("Cmd:" + compileCommand)
     res = os.system (compileCommand)
     
-    if res == '0':
-        print("Generated %s OK" % filename)
+    if res == 0:
+        print("Generated %s OK" % outputLibraryPath)
     else:
-        print("Error")
+        raise SystemError("Compilation failed. (" + str(res) + ")")


rickpresley:
We ran into an issue where our build process, which uses the CS mavlink generator python script, was failing because the Mavlink dll generated by the script was missing.  This issue had two parts:

- The CS code generated by the script was failing to compile due to a legitimate problem with our input XML.  However, the script was incorrectly checking the result of the compile command. os.system returns an integer, not a string.  Therefore, the script as it originally stood always reports an error, even when the compile was actually successful. The fix was to change _result =='0'_ to _result ==0_.
- When an actual error occurs (i.e when the compile fails) nothing happens other than the word "Error" being printed to screen.  Any build process relying on this script continues to run and there is very little to indicate why the build process eventually fails. The fix for this was to raise an exception indicating that compilation failed.

amilcarlucas:
Looks good to me, but I'm using the C stuff, not the cs stuff.

mday299:
@amilcarlucas -- do you know who we ask about getting this merged in?

amilcarlucas:
Just wait for it, someone will do it. Sometimes I wait for 6 months and I already commited more than 200 times :)

magicrub:
@peterbarker @rmackay9 @tridge  the change looks good to me.

generator/mavgen_cs.py
```diff
@@ -341,7 +341,9 @@ def generate(basename, xml):
     #print("Cmd:" + compileCommand)
     res = os.system (compileCommand)
     
-    if res == '0':
-        print("Generated %s OK" % filename)
+    if res == 0:
+        print("Generated %s OK" % outputLibraryPath)
     else:
         print("Error")
+        print("Error: Compilation failed. (" + str(res) + ")")
```
gmorph: No point in having the print("Error") above if we are then going to print a more comprehensive error message.

rickpresley: Weird... somehow that commit got mangled.  Originally the more detailed error message replaced the one above it.  Thanks for catching that!

peterbarker:
Any reason to both print and raise?

rickpresley:
I'm okay with removing the prints altogether and retaining only the exception.

mday299:
@peterbarker Removed spurious prints from the PR branch.

peterbarker:
Too slow!   Merged!

Thanks for the fix.  I'm still curious about both print and raise 'though.

Issue #97, opened 2017-07-28
fatal: memory overflow if length=0 with limited buffer size


az0uz:
In the C API v2.0, helper receive function `mavlink_frame_char_buffer`:
file: `generator/C/include_v2.0/mavlink_helpers.h`

In case of a glitchy packet with a `len` field value of `0` (or if you would have defined a message of length zero), and MAVLINK_MAX_PAYLOAD_LEN smaller than 249 (it is rounded to the next 8 bytes alignement, not sure if that's a good idea, so effectively 256 bytes):

The state machine doesn't handle correctly a null length:
In my case this happens because of a glitchy reception that changed the `len` field to zero (I don't have any zero-length messages) which happens every few minutes given my external conditions.

when it get to the state `MAVLINK_PARSE_STATE_GOT_MSGID3` (line 714)
The code read the first byte even if `len==0` and test `status->packet_idx == rxmsg->len` only after reading this first byte witch will be only true after reading 256 bytes instead of 0.

if you used `MAVLINK_MAX_PAYLOAD_LEN` to reduced the size of the buffer, their will be bytes overriding an unknown segment of memory witch leads to undefined behavior.

In my case using it on a atmega mcu it totally freezes the execution.

peterbarker:
On Fri, 28 Jul 2017, Antoine de Maleprade wrote:

> In the C API v2.0, helper receive function mavlink_frame_char_buffer:
> file: generator/C/include_v2.0/mavlink_helpers.h
> 
> In case of a glitchy packet with a len field value of 0 (or if you would
> have defined a message of length zero), and MAVLINK_MAX_PAYLOAD_LEN smaller
> than 249 (it is rounded to the next 8 bytes alignement, not sure if that's a
> good idea, so effectively 256 bytes):
> 
> The state machine doesn't handle correctly a null length:
> In my case this happens because of a glitchy reception that changed the len
> field to zero (I don't have any zero-length messages) which happens every
> few minutes given my external conditions.

Nice find.  Do you have a handy tlog which replicates?

az0uz:
No I haven't any, I am running the code on an atmega32u4 MCU without real debugging tools but I managed to track down what was wrong using Serial and I fixed the issue, You can review Merge Request #98.
To reproduce, set MAVLINK_MAX_PAYLOAD_LEN to, say, 64 bytes, and send any message but replace the length field by 0.

PR #98, opened 2017-07-28
C_API_v2.0: skip payload read when len is null ( fix: #97 )

az0uz:
fix memory overflow on ( len==0 && MAVLINK_MAX_PAYLOAD_LEN < 249 && not(MAVLINK_CHECK_MESSAGE_LENGTH) )

see issue #97

### DIFF
--- a/generator/C/include_v2.0/mavlink_helpers.h
+++ b/generator/C/include_v2.0/mavlink_helpers.h
@@ -677,7 +677,11 @@ MAVLINK_HELPER uint8_t mavlink_frame_char_buffer(mavlink_message_t* rxmsg,
 		rxmsg->msgid = c;
 		mavlink_update_checksum(rxmsg, c);
                 if (status->flags & MAVLINK_STATUS_FLAG_IN_MAVLINK1) {
-                    status->parse_state = MAVLINK_PARSE_STATE_GOT_MSGID3;
+                    if(rxmsg->len > 0){
+                        status->parse_state = MAVLINK_PARSE_STATE_GOT_MSGID3;
+                    } else {
+                        status->parse_state = MAVLINK_PARSE_STATE_GOT_PAYLOAD;
+                    }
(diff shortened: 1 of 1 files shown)


WickedShell:
I've been informed in the past that a 0 length payload is actually out of spec (and makes sense from the point of view that there isn't value to a message with no fields). In fact the strip 0 byte code is required to have a payload length of 1 even if it's a null byte. I'd propose it makes more sense to flag this as an invalid packet, drop it and restart the decoding sequence.

az0uz:
It makes more sense for me to keep it like this, and if you want to stop decoding on 0 length, just activate the MAVLINK_CHECK_MESSAGE_LENGTH. I don't see a good reason why 0 length message should be out of specs aside from not having any message that have no payload (then use the MAVLINK_CHECK_MESSAGE_LENGTH).

I could want to add a custom message like "CMD_HALT_ALL" having no payload, that shutdown everything with the lowest latency and lowest number of bytes to have more chance to get through with a very noisy data channel for example.
I doesn't make sense to exclude a specific value like this that could get someone not familiar with each line of the code get crazy over just implementing its custom message.

amilcarlucas:
Good point.

vooon:
Situation when len==0 is possible in Mavlink 2, if message only have one array filed and it is zeroed.

tridge:
good point on mavlink2 trimming, merged

peterbarker:
On Wed, 10 Jan 2018, Vladimir Ermakov wrote:

> Situation when len==0 is possible in Mavlink 2, if message only have one
> array filed and it is zeroed.

That's actually contrary to spec; you're not supposed to strip the last 
byte.... of course, we're really not sure where that was actually 
specified.

I guess we can say it's documented via code:

https://github.com/ArduPilot/pymavlink/blob/master/generator/mavgen_python.py#L207

Still, tridge has pulled the patch in as being at the very least benign, 
and possibly protecting against a bad read with non-spec client.

PR #105, opened 2017-08-24
Python 3.3 end of life

cclauss:
Python 3.3 goes [EOL](https://docs.python.org/devguide/index.html#branchstatus) next month so let's shift up one.


peterbarker:
Thanks!

PR #111, opened 2017-08-27
Option to filter messages according to target system

crashmatt:
Basic optional filter to ignore messages that are not from the connected/defined system

This is completely optional and defaults to not have any effect.

Note: The only intentional changes are in mavutil.py.


amilcarlucas:
Can you please rebase this?

amilcarlucas:
Now that multiple vehicles are supported, is this still needed ?
BWT You should remove all but the last commit from this PR.

peterbarker:
On Mon, 10 Dec 2018, Amilcar Lucas wrote:

> Now that multiple vehicles are supported, is this still needed ?
> BWT You should remove all but the last commit from this PR.

It is not, but I was planning on working with dronekit-python to ensure it 
works well with the new pymavlink stuff before removing this PR.  Don't 
want to break what meagre functionality we did have until the new stuff is 
shown to work in dronekit-python.

amilcarlucas:
Needs a rebase

PR #108, opened 2017-08-26
Flag syntax errors, undefined names, tab indentation

cclauss:
Ready for review but #109 (or similar) must be merged before this will pass on Python 3.

Use flake8 to prevent pull requests that contain Python syntax errors, undefined names, or tab indentation.

For the past 9 months, we have been running flake8 in `--exit-zero` mode which treats all flake8 errors as warnings.  This has helped us to find and fix many important errors.  Our issue counts are now reasonable: [Py2](https://travis-ci.org/ArduPilot/pymavlink/jobs/267437794#L531), [Py3](https://travis-ci.org/ArduPilot/pymavlink/jobs/267437796#L520).

So, the time has come to select a __handful of the the most serious errors__ and fail the Travis CI build if new PRs introduce those errors.  We do this by making two flake8 runs in our Travis process.  The first run can fail the build but only focuses on a few serious errors and the second runs in --exit-zero mode which will continue our current process of reporting but not failing the build on "style" violations.

PR #112, opened 2017-08-28
Flag Travis builds that contain syntax errors

cclauss:
Ready for review.  This is a "rebase" of #108

Use flake8 to prevent pull requests that contain Python syntax errors, undefined names, or tab indentation.

For the past 9 months, we have been running flake8 in --exit-zero mode which treats all flake8 errors as warnings. This has helped us to find and fix many important errors. Our issue counts are now reasonable: Py2, Py3.

So, the time has come to select a handful of the the most serious errors and fail the Travis CI build if new PRs introduce those errors. We do this by making two flake8 runs in our Travis process. The first run can fail the build but only focuses on a few serious errors and the second runs in --exit-zero mode which will continue our current process of reporting but not failing the build on "style" violations.


peterbarker:
On Mon, 28 Aug 2017, cclauss wrote:

> Ready for review. This is a "rebase" of #108

This changes behaviour for someone running flake8 locally, yes?

Not saying we can't do that, but we should do so knowingly...

The mavgen lines also seem to have been uglified; why the whitespace 
changes?

cclauss:
You are correct that removing `.flake8` would effect both local and Travis CI runs of flake8 tests.  As stated above, the "blanket coverage" of the current system limits flexibility but _an alternative approach is possible_ if you wish to keep the limitations in place on local flake8 runs.

We could `rm .flake8` in the .travis.yml before doing the flake8 runs to give ourselves maximum flexibility to target issues and fail the build on a key subset.  The file deletion change is not written back to the repo so .flake8 would remain in place for local runs.  My thought is that this is not a great idea because users would find it difficult to understand why Travis was generating different results than they are seeing locally.

peterbarker:
Merged, thanks!

PR #113, opened 2017-08-28
Simplify with dict.get() with default value

cclauss:


peterbarker:
Merged, thanks.  Could probably do with more specific commit messages (in this case specifying `mavparse.py`)

Issue #116, opened 2017-08-31
MAVError is an undefined name in generator/mavparse.py


cclauss:
`MAVError` is [defined in `generator/mavgen_python.py`](https://github.com/ArduPilot/pymavlink/blob/master/generator/mavgen_python.py#L377) and is [used in `generator/mavparse.py`](https://github.com/ArduPilot/pymavlink/blob/master/generator/mavparse.py#L107) but is never imported.

amilcarlucas:
No file changes here. why ?

cclauss:
@amilcarlucas My attempt to fix resulted in circular imports so I opened an issue in hopes that fresh eyes could find a solution.

PR #115, opened 2017-08-31
mavtelemetry_datarates.py: lengths_dict = None

cclauss:
Ensure proper alerting if the exec(compile(open())) operation fails.  `mavlink_message_lengths_dict` is used on lines 111, 147, and 149.

https://github.com/ArduPilot/pymavlink/search?q=mavlink_message_lengths_dict

PR #117, opened 2017-08-31
Add F821 `undefined name` detection to the testing

cclauss:
Undefined names can cause `NameError` exceptions to be raised at runtime.  Detecting them is also helpful for spotting Python 3 incompatibilities (raw_input(), unicode(), xrange(), etc.).  Adding this test to our critical error search will increase runtime stability and help us to achieve Python 3 compatibility.   ~#115 and~ #116 should be fixed __before__ this PR is merged.

While it does not block this PR, #114 will also be required for the Python 3 flake8 tests to pass .

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,8 +9,8 @@ install:
   - pip install -r requirements.txt
   - pip install flake8
 before_script:
-  # fail the build if there are Python syntax errors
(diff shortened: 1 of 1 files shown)


cclauss:
Bump.  Can we please get this one reviewed?

peterbarker:
Merged.

PR #122, opened 2017-10-26
Fixed an issue with the way that the Java Import statements were generated.

byron-conroy:
The "MAVLinkPacket" class in the Java implementation always imported the "common" messages and the name of the message definition XML file you generated from.

This was incorrect for two reasons:
- The "common" package may not exist if the message definition XML file does not include the "common.xml"
- If the message definition XML file includes additional message sets they will not be included properly.

This commit fixes both issues.


WickedShell:
Sorry for the delay on testing this. Accidentally spaced that it was pending. Only tested with the common XML included (as thats where my best test suite/GCS use case is present) but that works fine for me, and makes sense for why it would work on the other use case.

Issue #123, opened 2017-11-06
MAVLink v2.0 support for Java generated library


Hunter522:
The Java generated library has no support for MAVLink v2.0.

seeul8er:
MAVLink v2 would be really nice indeed :+1:

Hunter522:
I began working on this. Making progress. Will make a pull request hopefully in the next couple weeks.

WickedShell:
If you are looking at the Java parser at a fundamental level it might be worth keeping in mind that uint64_t is decoded incorrectly at the moment. Unfortunately the only solution I've found for this is to decode it into a BigInteger, which results in a lot of unintentional breakage/API changes for all current users which is why I didn't do this. I'm not sure if it's worth the effort of doing it now, or just continue to rely on all users who want to use a uint64_t needing to be aware of large values becoming negative. Just mentioning it as food for thought.

gmorph:
Thanks for working on this.  I too am keen to see this done as we can then incorporate it into Tower and have it support MAVLink 2.0 messages.  Thanks.

amilcarlucas:
I think this is done now and can be closed.

PR #127, opened 2017-12-04
MAVLink v2.0 support for Java generated library

Hunter522:
This pull request provides the necessary changes to implement MAVLink2 support for the Java autogenerated library. 

Tried my best to maintain current API and not make any large breaking changes.

Resolves issue #123

# Unsupported MAVLink2 features

These are some things I have some limited or no understanding on how they work. Maybe someone can help me out with these. Since they are not required this library still works for most cases. 

* packet signing
    * Not enough documentation on how this works. I've put some placeholders in the code right now for it.
* some of the optional MAVLink2 frame fields
    * `target_sysid` - ignored since they are used only for point-to-point setups which I do not know how to set up
    * `target_compid` - ignored since they are used only for point-to-point setups which I do not know how to set up

# Notable API changes

* `MAVLinkPacket` of course has the new MAVLink2 related fields, in adddition to `isMavlink2` boolean field to distinguish if the packet is MAVLink2 or not.
* Additional constructor `MAVLinkPacket(int payloadLength, boolean isMavlink2)` to construct a MAVLink2 packet. `MAVLinkPacket(int payloadLength)` constructor still exists and just delegates to `this(payloadLength, false)`
* MAVLink2 is **disabled by default** in construction of packets, messages, etc. 
* `Parser` can handle MAVLink1 and MAVLink2 packets making it backwards compatible. `isMavlink2` field in the `MAVLinkPacket` will be set accordingly. No need to configure the parser for this.
* Keep in mind that depending on if the `isMavlink2` field is set in `MAVLinkPacket` or `MAVLinkMessage`, the encoding/decoding methods will behave differently. For example, the MAVLink2-specific fields in `MAVLinkPacket` will be ignored when being encoded if `isMavlink2` is `false`. 
* `CRC::finish_checksum(int msgid)` will throw `IllegalArgumentException("invalid msgid")` if msgid is invalid since the CRC arrays are implemented as HashMaps now

# Migration notes:

For the most part this should not affect user's existing code. To integrate these changes, you mostly need to:

1. Use `new MAVLinkPacket(payloadLength, true)` when creating MAVLink2 packets
2. Check `mavlinkPacket.isMavlink2 == true` before using the MAVLink2-specific fields in the MAVLink packet
3. Check `mavlinkMessage.isMavlink2 == true` before using the extended fields of the MAVLink message

# Misc. notes:

* @WickedShell mentioned in #123 that `uint64_t` is not decoded correctly. I didn't touch this in this pull request since I think it would be better to fix it in a separate pull request.


julianoes:
@Hunter522 I just saw this and realized that there was some duplicate work over here: https://github.com/mavlink/pymavlink/pull/3/. It's a shame you had not seen it. We will have to figure out which parts to take from which forks. I don't have an opinion which changes should survive, I mostly hope we can get back in sync.

Hunter522:
@julianoes That's too bad :-( I did not notice there was a mavlink/pymavlink fork of this project. I too have no opinion of what's pulled from what. I briefly looked through the changes and there are a lot of similarities to what I did. When I find some time I will check it out.

Hunter522:
@WickedShell it's been a while ;-) I will take a look at this this weekend. For the most part I agree with your comments

rrr6399:
@Hunter522 Does this work yet? I'm going to try it. I'll push back any changes required that I find during testing.

WickedShell:
@rrr6399 You may also want to look at #290 which is an updated version of this PR. It's the one I'm probably looking at going forward, but haven't quite made the time to get to yet.

rrr6399:
> @rrr6399 You may also want to look at #290 which is an updated version of this PR. It's the one I'm probably looking at going forward, but haven't quite made the time to get to yet.

Thanks @WickedShell, I'll check it out - literally. :-)

magicrub:

PR #128, opened 2017-12-06
mavutil: Add set_mode_guided

patrickelectric:

### DIFF
--- a/mavutil.py
+++ b/mavutil.py
@@ -592,6 +592,16 @@ class mavfile(object):
             MAV_ACTION_LOITER = 27
             self.mav.action_send(self.target_system, self.target_component, MAV_ACTION_LOITER)
 
+    def set_mode_guided(self):
+        '''enter guided mode'''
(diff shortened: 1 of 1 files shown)


peterbarker:
This routine is not like the others around it.

There are two ways to set modes; the one you're using (the current one) sets a mavlink-custom-mode using set_mode.

The routines around this don't do that.  They're an abstraction layer to try to allow pymavlink to work across different autopilots.  This old mode-setting code is deprecated and only works on Plane.

Given that, this code is out-of-place.  It could also be reimplemented in terms of the `set_mode(...)` call in mavutil.py - but given how thin the wrapper is I don't think it's necessarily an improvement to `mavutil.py`.  Most users of pymavlink will probably want the "soft" option of being able to specify the new mode as a parameter.

Thoughts?

patrickelectric:
Ty @peterbarker, I have not seen the `set_mode`!

PR #129, opened 2017-12-06
Remove trailing spaces

patrickelectric:


peterbarker:
Thanks for the patch, Patrick.

However, this is a whitespace-fixing patch, and I'll explain why I won't merge it.

The problem with such a patch is that it kills the ability to quickly decide whether a given line has changed meaning recently.  It creates a point-in-time where it becomes harder to see the past.  I often use `git blame` to see why a problem might have been introduced - and the base form of the patch doesn't give details on the commit - and obviously only the most recent change (the whitespace) is given in that output.

The second reason is the merging issue.  If anyone has patches outstanding crossing these lines the users will get conflicts on those lines which they'll need to resolve.

I'm for gradual improvement for this stuff as much as possible.  A "no worse" policy on commits sort of thing.  Doesn't help that tridge tends to ignore trailing whitespace in his commits :-)

cclauss:
We can always configure flake8 to block any future PR that have tridge-isms. :-)
* E2 errors @ https://pycodestyle.readthedocs.io/en/latest/intro.html#error-codes

amilcarlucas:
I'm in favor of adding the flake8 tests for new code. But against this PR, just like @peterbarker

PR #131, opened 2017-12-07
mavgen_python: Check reset buffer logic even if a message is generated

### LINKED ISSUE #130: Memory leak in generated dialect code
Hi,
While working on the stand alone MAVCesium server @fnoop had identified that memory resources were continuing to be used. I eventually tracked down the cause to this code segment within pyMAVLink:
https://github.com/ArduPilot/pymavlink/blob/master/generator/mavgen_python.py#L526

33600756-1f414d24-d9a2-11e7-9004-f56532e35bef

Issue where the problem was identified: https://github.com/SamuelDudley/MAVCesium/issues/34

The behavior we observed was the result of a bytearray being extended over and over again, taking up more and more memory over time.

The issue is that it is quite possible to __always__ have a complete message waiting to be decoded when `parse_char()` is called. In this case the final `else` statement of  `parse_char()` never gets ran and the buffer continues to grow.

```
        def parse_char(self, c):
            '''input some data bytes, possibly returning a new message'''
            self.buf.extend(c)
            self.total_bytes_received += len(c)
            .....
            .....
            if m != None:               # <---- this always returns true with a new message.
                self.total_packets_received += 1
                self.__callbacks(m)
            else:                            # <---- so this code never gets a chance to run, causing a memory leak.
                if self.buf_len() == 0 and self.buf_index != 0:
                    self.buf = bytearray()
                    self.buf_index = 0
            return m
```
Although the logic in that statement `self.buf_len() == 0 and self.buf_index != 0` is true every time this method is called, `self.buf` and `self.buf_index` are never able to be reset... personally I think the else could be removed and we should be testing that condition every time that method is called. e.g.:
```
        def parse_char(self, c):
            '''input some data bytes, possibly returning a new message'''
            self.buf.extend(c)
            self.total_bytes_received += len(c)
            .....
            .....
            if m != None:
                self.total_packets_received += 1
                self.__callbacks(m)
            if self.buf_len() == 0 and self.buf_index != 0:
                self.buf = bytearray()
                self.buf_index = 0
            return m
```
Which equivalent to what is now in the MAVCesium server code: https://github.com/SamuelDudley/MAVCesium/blob/master/app/cesium_web_server.py#L215

SamuelDudley:
fixes #130

### DIFF
--- a/generator/mavgen_python.py
+++ b/generator/mavgen_python.py
@@ -523,12 +523,12 @@ class MAVLink(object):
             if m != None:
                 self.total_packets_received += 1
                 self.__callbacks(m)
-            else:
-                # XXX The idea here is if we've read something and there's nothing left in
-                # the buffer, reset it to 0 which frees the memory
-                if self.buf_len() == 0 and self.buf_index != 0:
-                    self.buf = bytearray()
-                    self.buf_index = 0
+            
+            # XXX The idea here is if we've read something and there's nothing left in
+            # the buffer, reset it to 0 which frees the memory
+            if self.buf_len() == 0 and self.buf_index != 0:
+                self.buf = bytearray()
+                self.buf_index = 0
 
             return m
 


SamuelDudley:
Any feedback on this fix would be greatly appreciated.
Thanks again!

amilcarlucas:
I have an application that runs 24/7 so I'll test this one.

SamuelDudley:
Thanks. It's very unlikely to be a problem unless you are receiving via a
very reliable connection e.g. TCP localhost.

amilcarlucas:
I do have a cable based TCP connection :(

peterbarker:
Merged it.  Thanks @SamuelDudley - it did take us a while!

SamuelDudley:
Thanks @peterbarker for carrying this over the line... I have a lot less hair than when I submitted this PR 🤣

PR #132, opened 2018-01-08
Sync with mavlink/pymavlink

julianoes:
I think it makes sense to bring ArduPilot/pymavlink and mavlink/pymavlink back in sync.
If we're not in sync we end up doing duplicate work as it unfortunately happened for Java mavlink 2.0 support (see https://github.com/ArduPilot/pymavlink/pull/127 vs. https://github.com/mavlink/pymavlink/pull/3.

Next to this PR, we need to do the other direction as well (I don't have permission though).


auturgy:
Not just pymavlink, but mavlink itself.
Probably worthy of a combined dev call to discuss, as we really need to put some guidance in place to get this fixed, and then keep it fixed.

Regards,

James

magicrub:
+1 for combined dev call

dagar:
I'd also be up for that.

generator/mavgen_java.py
```diff
@@ -93,9 +101,29 @@ def generate_CRC(directory, xml):
...
     * @param msgid
     *            The message id number
+    * @return crc_extra
+    *            The crc_extra value for the given message
     */
-    public void finish_checksum(int msgid) {
-        update_checksum(MAVLINK_MESSAGE_CRCS[msgid]);
+    public int finish_checksum(int msgid) {
+        int crc = 0;
+        if (msgid == 0) {
```
WickedShell: This should go away, heartbeat isn't actually a high frequency message on almost any deployed application, and is frankly not a parser concern.

generator/mavgen_java.py
```diff
@@ -93,9 +101,29 @@ def generate_CRC(directory, xml):
...
+            /* special handling to high frequency heartbeat message */
+            crc = MAVLINK_MESSAGE_CRCS[0];
+        } else {
+            int low = 0, high = MAVLINK_MESSAGE_IDS.length - 1;
```
WickedShell: Why is this rolling it's own binary search? You have a standard library which provides a fast and stable solution to this for you. https://docs.oracle.com/javase/7/docs/api/java/util/Arrays.html#binarySearch(int[],%20int)

generator/mavgen_java.py
```diff
@@ -301,13 +369,24 @@ def generate_MAVLinkMessage(directory, xml_list):
     */
     public CRC crc;
 
-    public MAVLinkPacket(int payloadLength){
+    public MAVLinkPacket(int payloadLength) {
```
WickedShell: There should be a `public MAVLinkPacket(int payloadLength, Protocol protocol)` constructor that the single arg one calls rather then calling `setProtocol(Protocol protocol)` all the time in the parser.

generator/mavgen_java.py
```diff
@@ -301,13 +369,24 @@ def generate_MAVLinkMessage(directory, xml_list):
...
         len = payloadLength;
         payload = new MAVLinkPayload(payloadLength);
+        crc_extra = -1;
     }
 
     /**
-    * Check if the size of the Payload is equal to the "len" byte
+    * Set the MAVLink protocol for this packet.
+    */
+    public void setProtocol(Protocol protocol) {
```
WickedShell: This helper isn't needed and is inconsistent with the rest of the output. All the other public fields are expected to be directly manipulated, that's why they are public.

magicrub:
Looks like we have a few merges happening in different directions. Is this being coordinated?

julianoes:
No, I just started this here but didn't have the time to spend time on it or clean it up. Not sure how to proceed.

peterbarker:
I think we really need separate PRs for separate things going into pymavlink here.

There MIGHT be some stuff in here that we want or need.  But it's all jumbled up together, and I given @WickedShell 's comments here it's certainly not going in as-is!

Are people OK if we close this in favour of pulling in things in separate PRs rather than as a conglomerate?

julianoes:
Sure, that's fine with me. As long as we can get back in sync which would be nice.

PR #136, opened 2018-01-10
Define raw_input() for Python 3

cclauss:
__raw_input()__ was removed from Python 3 in favor of __input()__.

### DIFF
--- a/tools/mavfft_int.py
+++ b/tools/mavfft_int.py
@@ -19,6 +19,11 @@ args = parser.parse_args()
 
 from pymavlink import mavutil
 
+try:
+    raw_input          # Python 2
+except NameError:
+    raw_input = input  # Python 3
+
 def plot_input(data, msg, prefix, start, end):
     preview = pylab.figure()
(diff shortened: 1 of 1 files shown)


peterbarker:
Merged, thanks!

cclauss:
@peterbarker Thanks for your help in closing these undefined name errors.  The only one that remains is https://github.com/ArduPilot/ardupilot/pull/6924

peterbarker:
On Wed, 10 Jan 2018, cclauss wrote:

> @peterbarker Thanks for your help in closing these undefined name errors.
> The only one that remains is ArduPilot/ardupilot#6924

Francisco points out the submodules probably still have issues.  Once we 
have merged this patch, do we get to add another flag to flake8 
configuration?

cclauss:
What submodules? (So that we can scan them with flake8 to see.)

ArduPilot/ardupilot does not run flake8 tests within .travis.yml like ArduPilot/pymavlink does so there is no flag (F821 undefined name) to add.

PR #137, opened 2018-01-12
str -> bytes for Python 3 support

pietrodn:
When running pymavlink in DroneKit on Python 3, pymavlink reports:

    Device /dev/ttyS0 is dead

The reason is that the `write()` function on line 835 in mavutil.py throws an exception because it is trying to encode `buf` (containing bytes) into a `str` (which in Python 3 is an Unicode string).

We can use a smarter approach: in Python 2.6 and later, [the `bytes` type is aliased to `str`](https://docs.python.org/3/whatsnew/2.6.html#pep-3112-byte-literals) (i.e., bytes), and in Python 3 it still represents bytes.
So we can cast `buf` to `bytes` both in Python 2 and 3 and obtain the correct behavior.

### DIFF
--- a/mavutil.py
+++ b/mavutil.py
@@ -834,9 +834,7 @@ class mavserial(mavfile):
 
     def write(self, buf):
         try:
-            if not isinstance(buf, str):
-                buf = str(buf)
-            return self.port.write(buf)
+            return self.port.write(bytes(buf))
         except Exception:
             if not self.portdead:
                 print("Device %s is dead" % self.device)


cclauss:
You might consider running __futurtize —stage2__ on the codebase.  Stage 1 has already been done.  See: http://python-future.org/automatic_conversion.html#stage-2-py3-style-code-with-wrappers-for-py2

pietrodn:
Superseded by #138, which includes this PR's changes and fixes other things.

dronology-outdoor:
@pietrodn It looks like you submitted a PR to get this fixed but did it get accepted?  I'm sure I've got the same issue, currently trying py3.6 with dk2.9.1 and getting the dead issue but I know I had dk working with py3 at some past point so wondering what version pairing that was...?

pietrodn:
The last DroneKit stable version on PyPI is very old and does not support Python 3. I suggest to use DroneKit's master together with the latest stable of pymavlink.

PR #139, opened 2018-01-15
Add an enumeration restriction to validate the units attribute

amilcarlucas:
It validates that the units have no typos.

### DIFF
--- a/generator/mavgen.py
+++ b/generator/mavgen.py
@@ -45,10 +45,10 @@ def mavgen(opts, args):
             from lxml import etree
             with open(schemaFile, 'r') as f:
                 xmlschema_root = etree.parse(f)
-                if opts.strict_units:
-                    # replace the generic "xs:string" type with a restricted list of known unit strings
+                if not opts.strict_units:
+                    # replace the strict "SI_Unit" list of known unit strings with a more generic "xs:string" type
                     for elem in xmlschema_root.iterfind('xs:attribute[@name="units"]', xmlschema_root.getroot().nsmap):
-                        elem.set("type", "SI_Unit")
+                        elem.set("type", "xs:string")
                 xmlschema = etree.XMLSchema(xmlschema_root)
         except:
             print("WARNING: Unable to load XML validator libraries. XML validation will not be performed", file=sys.stderr)
--- a/generator/mavgen.py
+++ b/generator/mavgen.py
@@ -25,6 +25,7 @@ DEFAULT_WIRE_PROTOCOL = mavparse.PROTOCOL_1_0
 DEFAULT_LANGUAGE = 'Python'
 DEFAULT_ERROR_LIMIT = 200
 DEFAULT_VALIDATE = True
+DEFAULT_STRICT_UNITS = False
 
 # List the supported languages. This is done globally because it's used by the GUI wrapper too
 supportedLanguages = ["C", "CS", "JavaScript", "Python", "WLua", "ObjC", "Swift", "Java", "C++11"]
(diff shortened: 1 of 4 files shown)


DonLakeFlyer:
Is it possible to update to wiki to point to this so folks know how to find the supported units?

DonLakeFlyer:
Oh, wait. Why is this not on the generic side of things?

amilcarlucas:
What do you mean with "generic side of things" ?

DonLakeFlyer:
In the mavlink repo, not the arudpilot repo.

OXINARF:
@DonLakeFlyer ArduPilot repo is the upstream, mavlink is a fork...

tridge:
only downside I see to doing this in the XML parser is that adding new units in a project means you must wait for a new pymavlink release

tridge:
perhaps we could have a --strict-units flag to the generator?

amilcarlucas:
Well .. people can:

* change the xsl
* do `setup.py --build --install`
* change the .xml
* change the ardupilot code
* recompile ardupilot

Would that not fix the problem ?

tridge:
not really, we'll just get developers like Paul and Leonard asking "why can't I build?". They would find it quite painful to do any of the things you mention (or even work out its needed)
I don't mind warnings and it being in our CI checks, but done as a default parser error I think is asking for trouble

amilcarlucas:
We already have a lot of units ... we will only seldom add units

tridge:
yes, but when we do add one we're going to cause a lot of pain for some developers. Why not emit a warning in the build, and make it compulsory in CI ?

amilcarlucas:
I did not find a way to issue a warning. But I did find a simple and ugly way to make it user configurable.

amilcarlucas:
If this one gets approved I have a followup PR to mavlink to update mavgenerate.py (the Tk GUI)

OXINARF:
Having two schemas is asking for trouble - it's changed very rarely which will make it even more probable to be forgotten.
Changing the schema on the fly seems a better option to me. You should just need to change the type attribute from `SI_Units` to `xs:string`.

amilcarlucas:
Something like:
`                if opts.strict_units:
                    for elem in tree.iterfind('xs:attribute [@name="units"]'):
                        elem.set("type", "SI_Units")`
 ??

generator/mavgen.py
```diff
@@ -44,6 +45,10 @@ def mavgen(opts, args):
             from lxml import etree
             with open(schemaFile, 'r') as f:
                 xmlschema_root = etree.parse(f)
+                if opts.strict_units:
+                    # replace the generic "xs:string" type with a restricted list of known unit strings
+                    for elem in xmlschema_root.iterfind('xs:attribute[@name="units"]'):
```
amilcarlucas: This part does not work yet :(
Anybody has any idea why ?

DonLakeFlyer:
> ArduPilot repo is the upstream, mavlink is a fork...

@OXINARF Ah, you mean for pymavlink right? Now I understand.

> perhaps we could have a --strict-units flag to the generator?

MY own personal experience with this sort of thing is that if you don't fail the build for compliance checks this will start to degrade as soon as it goes in.

amilcarlucas:
OK, it is fixed now :)

OXINARF:
@amilcarlucas 

Yes, something like that, although I would do it the other way around (have the schema file be strict and change it when strict units isn't asked in the mavgen script); this way anyone using the schema outside of mavgen will have the correct restrictions.

I think you can use XPath: `xmlschema_root.find(/xs:schema/xs:attribute[@name="units"]).set("type", "xs:string")`.

By the way, you can't use `--` inside a comment in XML like you've added in the XSD file.

@DonLakeFlyer Yes, pymavlink.

amilcarlucas:
@OXINARF I spited it into two commits for better review, and traceability.
@tridge wants the default to be "not strict" hence why it is implemented like that.

OXINARF:
@amilcarlucas tridge wants the mavgen to be not strict by default, nothing was said of the mavschema file itself 😉

amilcarlucas:
Good point

amilcarlucas:
@OXINARF I did as you asked.
If @tridge and you approve this I will squash the last two commits and then it can be merged.

tridge:
that looks much better, thanks!
we just need to fix the "b/s" failure so it passes travis

amilcarlucas:
This requires ArduPilot/mavlink#62 and mavlink/mavlink#826

amilcarlucas:
I updated ArduPilot/mavlink#62 with the remarks from @tridge and @WickedShell 
I also added two more commits here to fix/improve some stuff that I found.

tridge:
looks good, thanks Amilcar!

PR #142, opened 2018-01-24
mavutil.py : Added handling of mav_type ONBOARD_CONTROLLER to mavfile…

crichardson332:
… class, post_message method. This type previously caused errors in certain MAVProxy utilities for systems with ONBOARD_CONTROLLER components.

### DIFF
--- a/mavutil.py
+++ b/mavutil.py
@@ -265,7 +265,7 @@ class mavfile(object):
         if type == 'HEARTBEAT' and msg.get_srcComponent() != mavlink.MAV_COMP_ID_GIMBAL:
             self.target_system = msg.get_srcSystem()
             self.target_component = msg.get_srcComponent()
-            if float(mavlink.WIRE_PROTOCOL_VERSION) >= 1 and msg.type != mavlink.MAV_TYPE_GCS:
+            if float(mavlink.WIRE_PROTOCOL_VERSION) >= 1 and msg.type not in (mavlink.MAV_TYPE_GCS, mavlink.MAV_TYPE_ONBOARD_CONTROLLER):
                 self.flightmode = mode_string_v10(msg)
                 self.mav_type = msg.type
                 self.base_mode = msg.base_mode


mavutil.py
```diff
@@ -265,7 +265,7 @@ def post_message(self, msg):
         if type == 'HEARTBEAT' and msg.get_srcComponent() != mavlink.MAV_COMP_ID_GIMBAL:
             self.target_system = msg.get_srcSystem()
             self.target_component = msg.get_srcComponent()
-            if float(mavlink.WIRE_PROTOCOL_VERSION) >= 1 and msg.type != mavlink.MAV_TYPE_GCS:
+            if float(mavlink.WIRE_PROTOCOL_VERSION) >= 1 and msg.type != mavlink.MAV_TYPE_GCS and msg.type != mavlink.MAV_TYPE_ONBOARD_CONTROLLER:
```
cclauss: if float(mavlink.WIRE_PROTOCOL_VERSION) >= 1 and msg.type not in (mavlink.MAV_TYPE_GCS, mavlink.MAV_TYPE_ONBOARD_CONTROLLER):

crichardson332:
Updated (squashed and force pushed)

khancyr:
Hello, 
Thanks for your PR, but this is a duplication of https://github.com/ArduPilot/pymavlink/pull/87 . I think I will ask for a push as it was rooting a little bit as we didn't work on a better solution

crichardson332:
Ah I see, I didn't know this had been addressed before. I think it would be prudent for this simple fix to be merged, and then a better solution to be sought in the meantime. As it stands now, MAVProxy tools are broken for telemetry logs generated by systems with onboard controller type components, which is a problem...

mday299:
@magicrub Thoughts on getting this merged?

magicrub:
@mday299 this is above my pay grade. Let's ping @peterbarker and @khancyr and figure out why https://github.com/ArduPilot/pymavlink/pull/87 isn't already merged and how this variant is better/worse.

khancyr:
The solution is the same ! I rebased mine on master as I handle one more case ! I don't have merge permission, so it is up to the god hand to merge the one he want !

peterbarker:
https://github.com/ArduPilot/pymavlink/compare/master...peterbarker:heartbeat-changes?expand=1

Should I PR that or just merge it?  It seems to accomplish the same thing but consolidates the check.

khancyr:
@peterbarker just add onboard_controller and it will be good !

peterbarker:
I think what we now have in master will save people a lot of pain.  Please test :-)

PR #141, opened 2018-01-22
Python 3 fix: correct parsing of strings in parameter names

pietrodn:
In the process of rewriting a cleaner patch for [DroneKit PR #784](https://github.com/dronekit/dronekit-python/pull/784), I figured out that the byte-encoding of parameter names should be done on the pymavlink's side, not DroneKit's.

PR #30, opened 2016-11-01
DFReader.py: correct python3 support

khancyr:
Correct DFReader problem on python 3

PR #143, opened 2018-01-24
Some fixes for DFReader on Python 3

pietrodn:
This should solve the [mavlogdump.py crash](https://github.com/ArduPilot/pymavlink/pull/141#issuecomment-359960018) under Python 3.

Probably more work is needed, but I checked that these fixes allow the parsing of a .px4log file from Python 2 and Python 3.

### DIFF
--- a/DFReader.py
+++ b/DFReader.py
@@ -93,12 +93,9 @@ class DFFormat(object):
                 (self.type, self.name, self.format, self.columns))
 
 
-def null_term(str):
+def null_term(s):
     '''null terminate a string'''
-    idx = str.find("\0")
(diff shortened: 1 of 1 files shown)


pietrodn:
Related: #30

peterbarker:
@pietrodn I merged in @khancyr 's DFReader fix, and it appears to all be fixed in master.

I think I missed any answer to my query on whether we should be using `byte` throughout `pymavlink`.

Please reopen this if you think it will fix more than what's fixed in master already.

pietrodn:
I just filed #146 for a minor issue related to that PR.

PR #144, opened 2018-01-29
Filter HEARTBEAT sources

khancyr:


khancyr:
overseed https://github.com/ArduPilot/pymavlink/pull/142 and https://github.com/ArduPilot/pymavlink/pull/87

PR #146, opened 2018-02-05
Remove unneeded .close() statements

pietrodn:
This is a follow-up on #30 to correct a small bug introduced in there.
When using context managers and the [`with` statement](https://www.python.org/dev/peps/pep-0343/), it's superfluous to call `.close()` on the file-like object.


peterbarker:
Merged, thanks.

Issue #23, opened 2016-10-12
setup.py: Dictionary key 'pymavlink' is repeated with different values


cclauss:
https://github.com/ArduPilot/pymavlink/blob/master/setup.py#L129-L130

Was one of these two keys meant to be changed to something else?

pietrodn:
Fixed. Can be closed.

PR #149, opened 2018-02-18
Fix duplicate key in package_data (setup.py)

pietrodn:
The `package_data` option in `setup.py` contains a duplicate key.
This fixes issue #23.


peterbarker:
Merged, thanks.

PR #152, opened 2018-02-22
Fix missing x25crc module in python mavlink generation

jxltom:
Since only a single file will be generated in python's mavlink generation, ```x25crc``` can not be found.

Import ```x25crc``` from ```pymavlink``` directly will solve this problem.

Note that when using python's mavlink generation, ```pymavlink``` package will be required.

### DIFF
--- a/generator/mavgen_python.py
+++ b/generator/mavgen_python.py
@@ -30,7 +30,7 @@ from builtins import range
 from builtins import object
 import struct, array, time, json, os, sys, platform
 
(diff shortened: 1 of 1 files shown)


jxltom:
ping?

tridge:
this would mean that we'd use the installed version not the current version when running setup.py. I don't think thats a good idea

jxltom:
@tridge But in this way, the generated python module has to be placed within this project's directory otherthan any user specified directory. Maybe provides an option?

PR #159, opened 2018-03-12
mavlogdump: pack "Q" argument need to be int

patrickelectric:
[Format characters](https://docs.python.org/2/library/struct.html#format-characters)

### DIFF
--- a/tools/mavlogdump.py
+++ b/tools/mavlogdump.py
@@ -140,7 +140,7 @@ while True:
             continue
         if m.get_type() == 'PARAM_VALUE' and args.parms:
             timestamp = getattr(m, '_timestamp', None)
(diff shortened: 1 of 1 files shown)


peterbarker:
What breaks without this change?  I've no problem imagining it breaks *something* :-)

patrickelectric:
@peterbarker 
```
$ mavlogdump.py david_bowie.tlog --types SYS_STATUS --format csv -o file.csv
  File "/usr/bin/mavlogdump.py", line 167, in <module>
    output.write(struct.pack('>Q', timestamp*1.0e6))
struct.error: required argument is not an integer
```
`Python 3.6.4`

peterbarker:
Replicated, tested, squashed and merged.

Thanks!

PR #162, opened 2018-03-18
mavgen_javascript: pass file in MAVLink creator

doojinkang:
1. Pass file in MAVLink creator function, so that it can be used in send.
2. MAVLink can be used as 'mavlink.MAVLink'


WickedShell:
@magicrub I have nothing to do with JavaScript :)

davidbuzz:
this is a very old PR against the javascript/node bindings that i think is no longer relevant, due to https://github.com/ArduPilot/pymavlink/pull/431 and friends , so closing it.     if closed in error, please re-test agains master and open a new one .

PR #166, opened 2018-04-10
mavlink 2.0 lua support

gothame:


okalachev:
Does it work? It doesn't for me, showing only (broken) `HEARTBEAT` messages.



Using the last version of WireShark, PX4, QGroundControl.

PR #168, opened 2018-04-10
Add thread safe checking for the signature (downstream merge)

amilcarlucas:


generator/C/include_v2.0/mavlink_types.h
```diff
@@ -238,15 +238,18 @@ typedef bool (*mavlink_accept_unsigned_t)(const mavlink_status_t *status, uint32
 /*
   flags controlling signing
  */
-#define MAVLINK_SIGNING_FLAG_SIGN_OUTGOING 1
+#define MAVLINK_SIGNING_FLAG_SIGN_OUTGOING 1    ///< Enable outgoing signing
+// This disables protection against replay attacks, but can be necessary for systems without a proper time base
+#define MAVLINK_SIGNING_FLAG_NO_TIMESTAMPS 2    ///< Ignore timestamp mismatches
```
WickedShell: This is still an illogical feature that shouldn't be brought in.

amilcarlucas:
This was the issue discussed on the call ?
If this accomplishes the same thing as disable signing, then I will just close this.
Disabling signing is a simpler, non-half-backed solution, where the users then know and understand that they are disabling security. This PR just gives a false sense of security.

If some of my conclusions are wrong, feel free to re-open this.

PR #170, opened 2018-04-11
Raise bad_crc flag when return

rolandash:
Better to raise a BAD_CRC flag to distinguish from INCOMPLETE packet, so that app can tell the difference and handle it more precisely.

More return code are actually preferred here, however that will rely on the definition of new codes to be done first.


peterbarker:
Sorry, this is a change which will break basically every user of this library.  That's a boolean return value - 1 for "I got a message" and 0 for "I didn't get a message".

To get the current state (on 0 being returned) you'd check the status object.

PR #173, opened 2018-04-26
Update with ArduCopter 3.6.0 streams

amilcarlucas:


peterbarker:
This looks OK.  But it occurs to me that this patch shows the tool is closely tied to the firmware version.

Perhaps we should move this script into the ArduPilot repo in Tools somewhere?

amilcarlucas:
Very good point Peter. I'll prepare a PR sooninsh. Meanwhile can this be merged ?

PR #176, opened 2018-05-09
protocol.h: fix GCC warning: type qualifiers ignored on cast result type

bkueng:
GCC 8.0.1 warning/error that I got:
```
../../mavlink/include/mavlink/v2.0/common/./mavlink_msg_heartbeat.h: In function ‘uint8_t mavlink_msg_heartbeat_get_type(const mavlink_message_t*)’:
../../mavlink/include/mavlink/v2.0/common/../protocol.h:247:91: error: type qualifiers ignored on cast result type [-Werror=ignored-qualifiers]
 #define _MAV_RETURN_uint8_t(msg, wire_offset) (const uint8_t)_MAV_PAYLOAD(msg)[wire_offset]
```

### DIFF
--- a/generator/C/include_v0.9/protocol.h
+++ b/generator/C/include_v0.9/protocol.h
@@ -223,9 +223,9 @@ _MAV_PUT_ARRAY(int64_t,  i64)
 _MAV_PUT_ARRAY(float,    f)
 _MAV_PUT_ARRAY(double,   d)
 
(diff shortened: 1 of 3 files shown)


bkueng:
> I wonder if it is going to complain with line 276 too.

Interestingly it did not. But I can change it there as well if you want.

> Also, should v0.9 be fixed too?

Done

PR #179, opened 2018-05-14
bad CRC message on arduino compiler

Ahrovan:
missing pointer address on arduino compiler, and
case MAVLINK_PARSE_STATE_GOT_PAYLOAD: {
		const mavlink_msg_entry_t *e = mavlink_get_msg_entry(rxmsg->msgid);
		uint8_t crc_extra = e?e->crc_extra:0;
always set crc_extra to zero.
		mavlink_update_checksum(rxmsg, crc_extra);
then bad CRC occur


tridge:
the existing code looks correct, and with the change we risk a large array ending up on the stack, this doesn't look like a good idea

Ahrovan:
I hope you have an idea to solve this problem in the Arduino compiler because you did not accept my solution.

peterbarker:
@Ahrovan can you provide the compiler output and, if possible, steps to reproduce?

Please understand what tridge said - that structure is massive for a platform like Arduino, and shoving one onto the stack is a recipe for random run-time failure.

This code is actually a little odd to be enabling on a tiny platform.

PR #180, opened 2018-05-15
tools: fixed mavlogdump for binary output

tridge:


khancyr:
git submodule update --init --recursive should work

PR #182, opened 2018-05-18
Add support for deprecation of messages

amilcarlucas:
This is work in progress, I can not test it on a windows machine :(


amilcarlucas:
Tested on a windows machine using mavlink/mavlink#910

LorenzMeier:
Would be great to get this in.

PR #186, opened 2018-05-24
Try multiple XML etree implementations before giving up on validation

amilcarlucas:


amilcarlucas:
A better implementation is provided by #185

PR #193, opened 2018-07-09
Replace dpixels with dpix for consistency with already existing pix units

amilcarlucas:


amilcarlucas:
The tests can not pass because they require mavlink changes ( ArduPilot/mavlink#72 )

PR #194, opened 2018-07-09
Correct kibibyte and mebibyte units

amilcarlucas:


amilcarlucas:
The tests can not pass because they require mavlink changes ( Ardupilot/mavlink#73 )

dagar:
What's the status here? I don't think circular dependencies like this are workable. 
https://github.com/mavlink/mavlink/pull/962

PR #205, opened 2018-07-26
Mark WIP messages using XML markup

amilcarlucas:
Mark WIP messages using proper XML markup instead of having that information in some comments
As a side effect it improves the validation of Mavlink 2.0 message extensions

### DIFF
--- a/generator/mavschema.xsd
+++ b/generator/mavschema.xsd
@@ -150,6 +150,14 @@
     </xs:complexType>
 </xs:element>
 
+<xs:element name="wip">
+    <xs:complexType mixed="true">
+        <xs:sequence>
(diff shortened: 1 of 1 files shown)


dagar:
I'm not that familiar with the generator, but this looks correct to me. Let's give it a try on common and compare the output.

amilcarlucas:
The old code does not correctly detect malformed XML files. For example it does not detect that two description fields are not allowed in one message.

hamishwillee:
Yes, I saw that @amilcarlucas - good stuff. I assume you ran the generator on old and new versions, and it doesn't choke on anything and creates the same output? Or was there actually some errors in our common.xml definition?

PR #206, opened 2018-07-26
mavnative: fix compiler warning on missing case

khancyr:


amilcarlucas:
Shouldn't it print an error message, or increment an error counter?

khancyr:
@amilcarlucas no need ... in mavnative nothing use this flag

PR #210, opened 2018-08-08
Update schema for MAV_CMD params

amilcarlucas:
This provides detailed  information such that a ground station can
build ui from it

It builds upon #7

### DIFF
--- a/generator/mavschema.xsd
+++ b/generator/mavschema.xsd
@@ -11,7 +11,6 @@
 <!-- definition of attributes -->
 <xs:attribute name="name"> <!-- enum,entry,message,field -->
 </xs:attribute>
-<xs:attribute name="index" type="xs:unsignedByte"/> <!-- param -->
 <xs:attribute name="id" type="xs:unsignedShort"/> <!-- message -->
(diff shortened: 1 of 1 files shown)


generator/mavschema.xsd
```diff
@@ -135,10 +134,29 @@
...
     <xs:complexType mixed="true">
         <xs:attribute ref="index" use="required"/>
+        <xs:attribute ref="label" use="required"/>
+        <xs:attribute ref="units"/>
+        <xs:attribute ref="enum" />
+        <xs:attribute ref="decimalPlaces"/>
+        <xs:attribute ref="increment"/>
+        <xs:attribute ref="minValue"/>
+        <xs:attribute ref="maxValue"/>
+        <xs:attribute ref="hasLocation"/>
```
amilcarlucas: Should hasLocation be an attribute of the <entry> element instead ?

generator/mavschema.xsd
```diff
@@ -135,10 +134,29 @@
...
+        <xs:attribute ref="isDestination"/>
```
amilcarlucas: Should isDestinationbe an attribute of the <entry> element instead ?

PR #211, opened 2018-08-08
Improve MAV_CMD parameter information

amilcarlucas:
This is a followup of #7 and goes together with mavlink/mavlink#623

### DIFF
--- a/generator/mavschema.xsd
+++ b/generator/mavschema.xsd
@@ -141,8 +141,10 @@
 <xs:attribute name="increment" type="xs:float"/>                       <!-- parameter increment -->
 <xs:attribute name="minValue" type="xs:float"/>                        <!-- parameter minimum value -->
 <xs:attribute name="maxValue" type="xs:float"/>                        <!-- parameter maximum value -->
-<xs:attribute name="hasLocation" type="xs:boolean" default="true"/>    <!-- parameter has lat/lon location -->
-<xs:attribute name="isDestination" type="xs:boolean" default="true"/>  <!-- parameter is a destination -->
+
+<!-- definition entry elements attributes (like the ones used on MAV_CMD for example) -->
+<xs:attribute name="hasLocation" type="xs:boolean" default="true"/>    <!-- entry has lat/lon/alt location -->
+<xs:attribute name="isDestination" type="xs:boolean" default="true"/>  <!-- entry is a destination -->
 
 <!-- definition of complex elements -->
 <xs:element name="param">
@@ -155,8 +157,6 @@
         <xs:attribute ref="increment"/>
         <xs:attribute ref="minValue"/>
         <xs:attribute ref="maxValue"/>
-        <xs:attribute ref="hasLocation"/>
-        <xs:attribute ref="isDestination"/>
     </xs:complexType>
 </xs:element>
 
@@ -208,6 +208,8 @@
         </xs:sequence>
         <xs:attribute ref="value" />
         <xs:attribute ref="name" use="required"/>
+        <xs:attribute ref="hasLocation"/>
+        <xs:attribute ref="isDestination"/>
     </xs:complexType>
 </xs:element>
 
--- a/generator/mavschema.xsd
+++ b/generator/mavschema.xsd
@@ -11,7 +11,6 @@
 <!-- definition of attributes -->
 <xs:attribute name="name"> <!-- enum,entry,message,field -->
 </xs:attribute>
-<xs:attribute name="index" type="xs:unsignedByte"/> <!-- param -->
 <xs:attribute name="id" type="xs:unsignedShort"/> <!-- message -->
 <xs:attribute name="print_format" type="xs:string"/> <!-- field -->
 <xs:attribute name="enum" type="xs:string"/>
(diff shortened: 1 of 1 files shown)


amilcarlucas:
CI tests pass, this does not change the generated code, nor does it change the generated documentation.
I would really like to get feedback on this one. @DonLakeFlyer @LorenzMeier @dagar

dagar:
Overall looks good to me. From here I'd want to go through a few examples of updating common to see how well it works in practice.

Could you elaborate on the intended use of hasLocation and isDestination? What about other coordinate frames? I find the use of location in MAV_CMDs pretty confusing. With COMMAND_INT there's a field for the MAV_FRAME (https://mavlink.io/en/messages/common.html#COMMAND_INT), but I don't know how widespread the usage is.

DonLakeFlyer:
> Could you elaborate on the intended use of hasLocation and isDestination?

hasLocation - Command specifies a lat/lon/alt
isDestination - Command specified a lat/lon/alt and the vehicle flies through the point

@amilcarlucas QGC also has the concept of specifiesAltitudeOnly for a command. In this case only param 7 has altitude, but 6/7 are NOT lat/lon.

dagar:
What about different frames? There are already some that could be either global or local, some that are local only, relative alt, etc.

DonLakeFlyer:
> What about different frames? There are already some that could be either global or local, some that are local only, relative alt, etc.

The frame used when COMMAND_LONG is used should be in there somewhere.

peterbarker:
On Tue, 14 Aug 2018, Don Gagne wrote:

>
>       Could you elaborate on the intended use of hasLocation and isDestination?
> 
> hasLocation - Command specifies a lat/lon/alt
> isDestination - Command specified a lat/lon/alt and the vehicle flies through the point

What about the "kinda specifies location" commands like NAV_CMD_NAV_LAND - 
where ArduCopter will wander over to a location if it is specified before 
landing, but if one isn't specified it just lands?

DonLakeFlyer:
> What about the "kinda specifies location" commands like NAV_CMD_NAV_LAND -
> where ArduCopter will wander over to a location if it is specified before
> landing, but if one isn't specified it just lands?

Is that the lat/lon = 0 special case stuff. From a QGC standpoint I have no idea how to handle that sort of thing with a decent ui.

peterbarker:
On Tue, 14 Aug 2018, Don Gagne wrote:

> Is that the lat/lon = 0 special case stuff. From a QGC standpoint I have no idea how to handle that sort of thing with a decent ui.

Yes it is.  I'm not saying it's a good thing, but it is a thing :-)

I guess the markup could be "optional_location" or something?

generator/mavschema.xsd
```diff
@@ -9,15 +9,14 @@
...
 <!-- definition of attributes -->
-<xs:attribute name="name"> <!-- enum,entry,message,field -->
+<xs:attribute name="name"> <!-- used in enum,entry,message,field elements -->
 </xs:attribute>
-<xs:attribute name="index" type="xs:unsignedByte"/> <!-- param -->
-<xs:attribute name="id" type="xs:unsignedShort"/> <!-- message -->
-<xs:attribute name="print_format" type="xs:string"/> <!-- field -->
-<xs:attribute name="enum" type="xs:string"/>
-<xs:attribute name="display" type="xs:string"/>
+<xs:attribute name="id" type="xs:unsignedShort"/> <!-- used in message elements -->
```
hamishwillee: Can't a message id be up to 24 bits? This is 16 bits.

DonLakeFlyer:
> Should the "vehicle flies through the point" really be part of this? ie what about case where vehicle doesn't fly through point.

The point of this information is so that a ground station can use it to draw waypoint lines on a map. For that it needs to know whether an item specifies a location, and if the vehicle does or does not fly through that location.

DonLakeFlyer:
> I guess the markup could be "optional_location" or something?

I guess. But making that part of the standard mavlink spec makes my skin crawl a bit. I wouldn't want new examples of this to show up if at all possible.

amilcarlucas:
@hamishwillee most of the discussion on the attributes of this PR are on mavlink/mavlink#623

generator/mavschema.xsd
hamishwillee: @amilcarlucas ?

amilcarlucas: Very good point. As you can see that has been a "xs:unsignedShort" even before my changes.

So ... That is something that should be fixed by a future PR.

hamishwillee: Fair enough! Let's try remember to post one when this has gone in :-).

hamishwillee: Created https://github.com/ArduPilot/pymavlink/issues/216 for tracking

amilcarlucas:
@tridge  can I merge this ?

generator/mavschema.xsd
```diff
@@ -135,10 +134,29 @@
   </xs:restriction>
 </xs:simpleType>
 
+<!-- definition param elements attributes (like the ones used on MAV_CMD for example) -->
+<xs:attribute name="index" type="xs:unsignedByte"/>                    <!-- parameter index -->
+<xs:attribute name="label" type="xs:string"/>                          <!-- parameter label (name) -->
+<xs:attribute name="decimalPlaces" type="xs:unsignedByte"/>            <!-- parameter number of decimal places to be displayed in the GUI -->
```
amilcarlucas: I guess this should be replaced with `print_format`. That way it reuses an existing concept. There is no need to introduce a new one IMHO.

generator/mavschema.xsd
```diff
@@ -135,10 +134,29 @@
...
+<xs:attribute name="isDestination" type="xs:boolean" default="true"/>  <!-- entry is a destination -->
+
 <!-- definition of complex elements -->
 <xs:element name="param">
     <xs:complexType mixed="true">
         <xs:attribute ref="index" use="required"/>
+        <xs:attribute ref="label"/>
+        <xs:attribute ref="units"/>
+        <xs:attribute ref="enum" />
+        <xs:attribute ref="decimalPlaces"/>
```
amilcarlucas: I guess this should be replaced with `print_format`. That way it reuses an existing concept. There is no need to introduce a new one IMHO.

DonLakeFlyer:
In QGC the concept of decimal places is not just used to display values to the user but there are various places in the code where the actual precision of the value is needed to be known. For those the number of decimal places is required. This usage is currently only for parameter values not MAV_CMD, but I could see this coming up for that as well.

The specific case is for these slider controls:
screen shot 2018-09-29 at 10 07 37 am

In order to build a set of value before/after the current value the precision of the value being modified needs to be known such that those values match the precision. I could see the same thing coming up for a UI which is used to build up COMMAND_LONG calls.

hamishwillee:
@amilcarlucas Can you chase @tridge about this one in the ArduPilot dev call or recommend someone else with the authority and background to approve this?

amilcarlucas:
@DonLakeFlyer well, print_format can be used to achieve the same thing.
@hamishwillee yes, I will do that.

DonLakeFlyer:
What's the definition of print_format and how would it support the same thing?

DonLakeFlyer:
Having to write a generic print format parser is not something anyone would really want to have to implement.

amilcarlucas:
The point is that the mavlink spec already uses `print_format` since a couple of years.
So you (and/or other GCSs) probably have that implemented already anyways.

PR #226, opened 2018-10-08
Tools: allow to save extracted parameter with mavparms.py

khancyr:
add cmdline parameter to save parameter extracted !

### DIFF
--- a/tools/mavparms.py
+++ b/tools/mavparms.py
@@ -1,26 +1,25 @@
 #!/usr/bin/env python
 
-'''
-extract mavlink parameter values
-'''
+"""
+Extract mavlink parameter values from log files.
+Its works both on dataflashlog (.BIN) and (.log) or telemetry log (.tlog).
+"""
 from __future__ import print_function
-
-
 import time
+from pymavlink import mavutil
 
 from argparse import ArgumentParser
 parser = ArgumentParser(description=__doc__)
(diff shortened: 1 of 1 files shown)


peterbarker:
What does this gain us vs just redirecting the output of the script?

I note that this *does* kill that functionality because of the print() messages present.

khancyr:
@peterbarker, you made the assumption that everybody know how to redirect the output ! that is the difference ! Moreover, I am preparing a small GUI for Pymavlink tool so having parameter option is simpler. This is really intend to people not use to cmdline control  (and that a big part of user)
I will add # on each print so it won't break old behaviour

peterbarker:
On Tue, 9 Oct 2018, Pierre Kancir wrote:

> @peterbarker, you made the assumption that everybody know how to redirect the output ! that is the difference ! Moreover, I am preparing a small GUI for
> Pymavlink tool so having parameter option is simpler. This is really intend to people not use to cmdline control (and that a big part of user)

In that case - shelling out to mavparms.py is probably the wrong approach 
:-)

khancyr:
surely, but it will does the job for now ! It is not like I will get the time to do the GUI soon ^^ .
Are you good with the new option ?

peterbarker:
On Wed, 10 Oct 2018, Pierre Kancir wrote:

> surely, but it will does the job for now ! It is not like I will get the time to do the GUI soon ^^ .
> Are you good with the new option ?

Sorry, I'm not in favour of this option - it's redundant with redirection.

khancyr:
Ok then I close this

PR #227, opened 2018-10-15
mavgen_python: correct initialisation of string fields

peterbarker:


tridge:
another approach would be a self.fielddefaults[] list? Slightly less code

peterbarker:
On Mon, 15 Oct 2018, Andrew Tridgell wrote:

> another approach would be a self.fielddefaults[] list? Slightly less code

I considered it :-)

But added the duplicate code anyway.

Just lazy.

PR #232, opened 2018-10-23
Use struct.Struct some more and reuse unpacker to pack

peterbarker:


peterbarker:
This one just got a little bigger.

It's somewhat faster than master, taking 13.8 seconds to parse a log where master takes 19.4

generator/mavgen_python.py
```diff
@@ -781,11 +800,10 @@ def decode(self, msgbuf):
...
-                for i in range(0, len(tlist)):
-                    if type.fieldtypes[i] == 'char':
-                        if sys.version_info.major >= 3:
-                            tlist[i] = tlist[i].decode('utf-8')
-                        tlist[i] = str(MAVString(tlist[i]))
+                for i in type.char_fields:
+                    if sys.version_info.major >= 3:
+                        tlist[i] = tlist[i].decode('utf-8')
+                    tlist[i] = str(MAVString(tlist[i]))
                 t = tuple(tlist)
```
pietrodn: Another idea for optimization: re-assigning `tlist` to `t` as a tuple is useless, since it is possible to directly unpack `tlist`: `m = type(*tlist)`.

peterbarker: Done

peterbarker:
Testing shows this to be about 4.5% faster.

PR #235, opened 2018-10-29
Simplify comparisons with None, True, False

pietrodn:
PEP8 [suggests](https://docs.quantifiedcode.com/python-anti-patterns/readability/comparison_to_none.html) to use `is None` to compare a value to `None`, instead of the equality operator.

This has an important consequence in the `parse_char` function in `mavgen_python.py`. When profiling a DroneKit-based application with yappi, I noticed that a lot of computation time (8.7%) was spent in the `__eq__` method of MAVLink message object.
After running my application for a few minutes, the profiler recorded 1.9 million calls to `MAVLink_attitude_message.__eq__`.
That was weird!

The problematic code resides in the [`parse_char` function](https://github.com/ArduPilot/pymavlink/blob/b4d048f16422ebe482593ff7ad0e35baaa92b833/generator/mavgen_python.py#L558):

```python
            if m != None:
                self.total_packets_received += 1
                self.__callbacks(m)
```

If `m` is not None, the inequality check results in a call to the [`__ne__()` method](https://github.com/ArduPilot/pymavlink/blob/b4d048f16422ebe482593ff7ad0e35baaa92b833/generator/mavgen_python.py#L162) of the `MAVLink_message` class, which in turn calls `__eq__`:

```python
    def __ne__(self, other):
        return not self.__eq__(other)
    def __eq__(self, other):
        if other == None:
            return False
```

By the way, if `other` is in turn a `MAVLink_message`, the equality operator would call `__eq__` again: **infinite loop**!

We can avoid the two function calls and the potential infinite loop by using the `is` operator to compare `m` with `None`.
I went ahead and searched for the same pattern in all the code.

This PR also simplifies [redundant binary comparisons](https://docs.quantifiedcode.com/python-anti-patterns/readability/comparison_to_true.html) (e.g. `== True`).

### DIFF
--- a/examples/bwtest.py
+++ b/examples/bwtest.py
@@ -44,7 +44,7 @@ while True:
         #and return a new message if available
         m = master.recv_msg()
 
-        if m == None: break  #No new message
+        if m is None: break  #No new message
         
         if m.get_type() not in counts:
         #if no messages of this type received, add this type to the counts dict
--- a/examples/mav2pcap.py
+++ b/examples/mav2pcap.py
@@ -159,7 +159,7 @@ def convert_file(mavlink_file, pcap_file):
         if next_sof > 0:
             print("skipped " + str(next_sof) + " bytes")
             if write_junk:
-                if skipped_char != None:
+                if skipped_char is not None:
                     junk = skipped_char + data[:next_sof]
                     skipped_char = None
                 write_packet(i, junk, 0x03, len(junk))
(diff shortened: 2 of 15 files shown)


pietrodn:
Actually I was wrong about the numbers... the 1.9 M calls were DroneKit's fault ([this line of code](https://github.com/dronekit/dronekit-python/blob/master/dronekit/__init__.py#L1285)).

From pymavlink, I registered "only" 13.7k calls. Still, it's worth fixing :-)

peterbarker:
Merged, thanks!

PR #239, opened 2018-11-08
Generator: add mm/h units for weather station

khancyr:
needed for weather station to report rain


karthikdesai:
There's a weather station that speaks mavlink? 😮

PR #240, opened 2018-11-13
generate WIP warnings on the use of WIP messages in C code

tridge:


hamishwillee:
@tridge 
1. So how do you expect this to be used? I think you are suggesting that a flight stack that wants to prevent WIP messages in their environment would add the following line to whatever code they have that includes the library?
```
#include <mavlink/mavlink.h>
#define MAVLINK_WIP __attribute__((warning("MAVLink work in progress")))
```
2. FMI, I'd like to run the tests with this active - ie use `./test_generator.sh`. How do I inject this into the test suite?

tridge:
@hamishwillee yes, that's how it would be used

on (2), the problem is we run tests with -Werror, so any warnings would make the test fail

hamishwillee:
> on (2), the problem is we run tests with -Werror, so any warnings would make the test fail

Yes, I want to run the tests, watch them fail, and see the message :-)

It was more FMI than anything else, because it isn't obvious (to me) how you could do this if you wanted given that most of the test is autogenerated. 

If it were me I would additionally add a file level comment about the fact the message is WIP. My concern is users who see a message and try to use it without understanding the implications. You might not consider that a high risk.

I'm OK with this fix otherwise. I don't have any magic to merge or approve this.

hamishwillee:
When we use similar approach for deprecated tag, wouldn't it be better if that was "default on"?

amilcarlucas:
yes, I guess so.

hamishwillee:
@peterbarker @auturgy Should this be merged? 

I don't love the implementation in that the whole idea was that WIP stuff should not be used in production.
tridge has his reasons, but I doubt that people will think to enable this, which makes it pointless.

I would have preferred perhaps that the library was **built** without WIP messages by default and you had to specifically enable them using a build time flag.
That of course would be a bit frustrating because it would be all or nothing - you want a WIP thing in ardupilot you have to take the ones in common too.

We are moving away from using WIP in common, but I suspect that dialects may still find it useful to keep WIP, and therefore possibly to have this warning.

auturgy:
Ideally we’d build with WIP by default in Master/Dev branches but not in production/release branches.

hamishwillee:
> Ideally we’d build with WIP by default in Master/Dev branches but not in production/release branches.

@auturgy Yes, though I guess that would have its own challenges when you cut the release and realize you actually rely on something WIP.
Obviously we can do it, but not without development effort. What would be good is if we could agree how we would like this to work, so at least someone writing a generator can do the right thing here.

Is there any strong feeling on what you want ArduPilot side?

amilcarlucas:
I also vote to remove WIP items from stable releases. What do you think @tridge?

python36:
A new PR needs to be created because after the https://github.com/ArduPilot/pymavlink/pull/592 merge, the WIP attribute needs to be added to the enumeration. 

I think that elements with the WIP tag should be ignored by the parser (mavparse.py), not every generator.

peterbarker:
> @peterbarker @auturgy Should this be merged?

I think so.

> I don't love the implementation in that the whole idea was that WIP stuff should not be used in production. tridge has his reasons, but I doubt that people will think to enable this, which makes it pointless.
> 
> I would have preferred perhaps that the library was **built** without WIP messages by default and you had to specifically enable them using a build time flag. That of course would be a bit frustrating because it would be all or nothing - you want a WIP thing in ardupilot you have to take the ones in common too.

I think that's acceptable.

> We are moving away from using WIP in common, but I suspect that dialects may still find it useful to keep WIP, and therefore possibly to have this warning.

It's kind of neither here nor there.  It's not a great look from a standards-perspective having things in `common.xml` which are WIP.... if we can move everyone to using `all.xml` then we can gradually move to only having them in `development.xml`.

peterbarker:
> I also vote to remove WIP items from stable releases. What do you think @tridge?

We need to discuss this in the context of the generated C mavlink headers here: https://github.com/mavlink/c_library_v2

If we were to add an option to `mavgen.py` to omit `<wip/>` tags - would that repository start to use that option?  The implication there is that suddenly anything marked `<wip/>` would disappear from any project using those headers - so, for example, qgroundcontrol (https://github.com/mavlink/qgroundcontrol/blob/master/.gitmodules#L4)

That could be a lot of breakage in a short amount of time.  This might be considered a transition issue?  Should the pre-compiled headers come in two flavours, with-WIP and without?

Problem then becomes "what gets built with WIP, what doesn't?".  So, for example, do QGC's daily builds ship with WIP messages and the releases not?  ArduPilot's "latest" releases with-WIP and "stable" releases without?

One big advantage of not including WIP messages is that field definitions can change semantics during development *without* changing the checksum of the message.  Might this lead to dangerous situations?

hamishwillee:
@peterbarker `wip` was our first shot at trying to make it easier to develop new things without committing users to having to take them or developers to compatibility. I "think" the original idea was that they would always be present and always warn unless you turned them off. Had we decided to build without the WIP stuff at the beginning it might have worked, but I think you're right and it might create a lot of breaks now.

In particular I suspect PX4 is dependent on some of that stuff, so I doubt the generated libraries would follow, at least immediately. 
We're trying to work to where there is little or no `wip` in common. Either things are in development.xml or they are actually common, or they are in a dialect. It is slow going.

Anyway, my leaning is to go with this PR. We're hoping to not use it ourselves long term, but others might like the regular warning.

hamishwillee:
@peterbarker Did you have a chance to discuss this in call?

generator/C/include_v0.9/protocol.h
```diff
@@ -34,6 +34,14 @@
 #define MAVLINK_END_UART_SEND(chan, length)
 #endif
 
+/*
+  to get warnings when any WIP message is used, add this:
+  #define MAVLINK_WIP __attribute__((warning("MAVLink work in progress")))
+*/
+#ifndef MAVLINK_WIP
+#define MAVLINK_WIP
+#endif
```
julianoes, line 43: Shouldn't the default be to print the warning and a custom define can be used to silence it?

hamishwillee: Absolutely, but see my question and the response from @tridge above https://github.com/ArduPilot/pymavlink/pull/240#issuecomment-438557184

Do you think we could make the behaviour as we expect but explicitly disable the warning in tests? We don't particularly want to have warnings in these tests - we want to have them in live environments.

hamishwillee:
@peterbarker We closed #1924 in December 2022 in favour of this PR added in 2018 on your suggestion.
You're right - this is a better technical solution than constructing XML definition include files to omit development.xml. Or at least it is a better solution for C which is our library of most concern.

This has basically been blocked on the request to enable the warning by default. I think after 5 years we can agree that the Pymavlink dev team can block doing the "bleedingly obvious correct thing" forever right?

How about you update this to fix the merge conflicts and merge as is? That will be very annoying, and mean that that we will continue to have WIP leaking into releases. BUT it does mean that it is possible for developers to more easily get these warnings.

Thoughts?

peterbarker:
Force-pushed to fix conflicts with has_location changes.

hamishwillee:
Thanks very much @peterbarker .

What needs to happen to get this in? Do you need more testing, or something else? I can help.

hamishwillee:
Shall we wait another 5 years?

hamishwillee:
FYI @auturgy - this is the one discussed in the call.

tridge:
@hamishwillee we would need to remove the wip tags on the opendroneid messages first, or all builds that use -Werror (including ArduPilot) will fail

tridge:
there is also the basic problem that we use -Werror in CI in ArduPilot. So as soon as a new WIP message gets added that we want to use in ArduPilot master (which is quite often) we would fail CI

tridge:
we can bring it in default off, just not default on

hamishwillee:
@tridge Thank you. So can't you adjust CI to turn this default off _in CI_ - default on everywhere else?

If this is not possible then as above, we'd rather get this in default off than have no way to check for WIP.

> @hamishwillee we would need to remove the wip tags on the opendroneid messages first,

W.r.t. ODID none of those messages have WIP. There are WIP things though, such as https://mavlink.io/en/messages/common.html#MAV_CMD_DO_ORBIT - I'm happy to do another round of accepting/removing if that is needed.

tridge:
> So can't you adjust CI to turn this default off _in CI_ - default on everywhere else?

we don't want it for developers doing local builds, or in the custom build server.
The point is I don't want lots of warnings being shown to people who can't do anything about the warnings.

hamishwillee:
OK - thanks. Then let's get it in disabled by default.

hamishwillee:
So can we merge this please @tridge / @peterbarker ?

julianoes:
Any chance we could merge it? It's been 6 years and it would still be good.

hamishwillee:
Yes it would!

amilcarlucas:
It only generates warnings when explicitly asked to. I think it should be merged.

PR #242, opened 2018-11-15
Add centiseconds as SI_Unit

eyeam3:
As discussed in mavlink/mavlink#1009 we have to add centiseconds as SI unit.

@amilcarlucas Is this all I had to change?


amilcarlucas:
Thanks

PR #246, opened 2018-11-25
mavparse.py: understand MANUAL_CONTROL target system field

peterbarker:
Currently mavlink routing doesn't work on this field so everyone sees it.


OXINARF:
Is there a reason not to fix the field name?

tridge:
@OXINARF it would break hash and api

PR #247, opened 2018-11-26
mavparmdiff.py: add option to not exclude some parameters from compar…

peterbarker:
…ison

### DIFF
--- a/tools/mavparmdiff.py
+++ b/tools/mavparmdiff.py
@@ -9,6 +9,16 @@ from argparse import ArgumentParser
 parser = ArgumentParser(description=__doc__)
 parser.add_argument("file1", metavar="FILE1")
 parser.add_argument("file2", metavar="FILE2")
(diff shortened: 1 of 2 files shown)


tools/mavparmdiff.py
```diff
@@ -9,13 +9,23 @@
...
 parser.add_argument("file1", metavar="FILE1")
 parser.add_argument("file2", metavar="FILE2")
+parser.add_argument("--use-excludes",
+                    help="exclude volatile and similar parameters",
+                    default=True,
+                    action='store_true',
+                    dest='use_excludes')
+parser.add_argument("--no-excludes",
+                    help="include volatile and similar parameters",
+                    default=True,
```
OXINARF: I think I've read that putting default in both options isn't recommended when the destination is the same.

tridge: I agree

peterbarker:
Merged this.

PR #251, opened 2018-12-07
support multiple vehicle operation for mavutil

tridge:
this separates out parameters and messages per source system, while
maintaining API compatibility with previous versions using attribute
properties

### DIFF
--- a/mavutil.py
+++ b/mavutil.py
@@ -109,6 +109,22 @@ def set_dialect(dialect):
 # Set the default dialect. This is done here as it needs to be after the function declaration
 set_dialect(os.environ['MAVLINK_DIALECT'])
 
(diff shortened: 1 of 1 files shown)


mavutil.py
```diff
@@ -134,29 +153,78 @@ def __init__(self, fd, address, source_system=255, source_component=0, notimesta
         self.mav.robust_parsing = self.robust_parsing
         self.logfile = None
         self.logfile_raw = None
-        self.param_fetch_in_progress = False
```
jaxxzer, line 137: I have a script that was checking this attribute to wait for all of the parameters to be fetched. How should I check if the param fetch is complete after this patch?

PR #258, opened 2019-01-16
mavschema: Add default and restricted attributes to params

hamishwillee:
MAV_CMD param values are current indicated as reserved using a tag content of "Reserved" or "Empty" (if no param is defined, this is also assumed to be reserved). The default value is typically default set by Mission Planner and QGC as 0, though sometimes a user will indicate default value of NaN in the description. 

> **Note** The default value is a bit of a pain for params, because once it is set to 0 or NaN the param can't be reused for a real param where the default value means anything other than "unchanged". Big discussion of reasons [here](https://github.com/mavlink/mavlink/issues/1050).

This PR adds `reserved` attribute to allow a param to explicitly be marked as reserved, and `default` to allow a default value (along with enabling `NaN` as a possible default).

@amilcarlucas @peterbarker I would love to further add tests to assert that NaN cannot be assigned to param 5, 6 (which can be sent in COMMAND_INT), and also to ensure that a unique index must be applied for every param. However when I try add xs:assertion I get error `WARNING: Unable to load XML validator libraries. XML validation will not be performed`. Any ideas on how to do this would be great, but otherwise happy for this to go in/be reviewed.

### DIFF
--- a/generator/mavschema.xsd
+++ b/generator/mavschema.xsd
@@ -31,6 +31,7 @@
         <xs:pattern value="\d{1,10}"/> <!-- base 10 int -->
         <xs:pattern value="0[xX][0-9a-fA-F]{1,8}"/> <!-- base 16 -->
         <xs:pattern value="0[bB][0-1]{1,32}"/> <!-- base 1 -->
+        <xs:pattern value="NaN"/> <!-- Allow not-a-number as a default (for params) -->
     </xs:restriction>
   </xs:simpleType>
 </xs:attribute>
@@ -145,6 +146,7 @@
 <xs:attribute name="increment" type="xs:float"/>                       <!-- parameter increment -->
 <xs:attribute name="minValue" type="xs:float"/>                        <!-- parameter minimum value -->
 <xs:attribute name="maxValue" type="xs:float"/>                        <!-- parameter maximum value -->
+<xs:attribute name="reserved" type="xs:boolean" default="false"/>      <!-- parameter is reserved -->
 
 <!-- definition entry elements attributes (like the ones used on MAV_CMD for example) -->
 <xs:attribute name="hasLocation" type="xs:boolean" default="true"/>    <!-- entry has lat/lon/alt location -->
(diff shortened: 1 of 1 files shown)


peterbarker:
On Tue, 15 Jan 2019, Hamish Willee wrote:

> when I try add xs:assertion I get error WARNING: Unable to load XML
> validator libraries. XML validation will not be performed. Any ideas on how
> to do this would be great, but otherwise happy for this to go in/be
> reviewed.

pip install --user lxml ?

khancyr:
hello, you need libxslt1.1 to have the validator !

hamishwillee:
@khancyr Thank you! Unfortunately that is already installed and I still get the error. I thought perhaps the schema definition up the top might be wrong `<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema">` but alternatives like `<xs:schema xmlns:xs="http://www.w3.org/2001/XMLSchema-instance">` are invalid.

Any other ideas?

amilcarlucas:
on windows you also need to install xslt-dvel cygwin package.

hamishwillee:
Hi @amilcarlucas 

> on windows you also need to install xslt-dvel cygwin package.

Really? We don't run this in cygwin on Windows. In any case, not sure this would help because the validation code uses lxml (the python dependency) and as above https://github.com/ArduPilot/pymavlink/pull/258#issuecomment-455389196 I'm pretty sure that we're limited by the toolchain to XSL 1.0. 

So we either change our toolchain or our validator or accept we'll never be able to do full validation.

khancyr:
Could you provide a file or a schema that we aren't able to validate?

khancyr:
Ok I see what I didn't understand previously ... yes lmxl don't support xsd 1.1 ... in fact only few parser support it.
What I wouldn't do is to change the parser, as it will be painfull for long time for everybody. Lxml also support parsing schematron and RelaxNG schema, would you found what you want in it ? Otherwise a full conversion to json could be a better move

hamishwillee:
@khancyr 

> What I wouldn't do is to change the parser, as it will be painfull for long time for everybody. 

Absolutely!

Probably Schematron does what we want. The lowest cost effort is to continue to use XSD validation as we do now, and run Schematron as a second parser for any additional validation that isn't supported by that toolchain. I think that can probably done with minimal change.

hamishwillee:
@amilcarlucas I don't have merge rights here - so if you are able to push this through as well that would be good.

PR #263, opened 2019-01-31
Working on updating to python 3

IAmMadfly:
Moving one line as allowed it to work (Windows 10). ".generator" is not declared in the function used. Replaced line to be global.

### DIFF
--- a/mavutil.py
+++ b/mavutil.py
@@ -21,6 +21,8 @@ try:
 except Exception:
     pass
 
+from .generator import mavparse
+
 # maximum packet length for a single receive call - use the UDP limit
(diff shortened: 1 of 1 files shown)


peterbarker:
Please paste in the exception you're fixing with this one.

Moving the import line may require us to change the dependencies for pymavlink, so it's worth considering the change, and whether we can/should fix it some other way.  I mean - having the imports in the middle isn't in good taste, so it must have been done that way for a reason....

KarlRun:
Very sorry. It seems as though I had an environmental error previously. My program is working correctly now.
I apologize for the inconvenience.

PR #265, opened 2019-02-05
Add grams to SI_Units

hamishwillee:
Required for https://github.com/mavlink/mavlink/pull/1072


hamishwillee:
@amilcarlucas Could you please push this in.

hamishwillee:
> technically kg is the SI unit, right?

I *think* the "base" unit is the kg, and SI refer to gram and milligram as "derived units". Either way though, we have other "non SI" units in there where it is pragmatic to do so - as in this case.

amilcarlucas:
done, thanks

hamishwillee:
Thanks @amilcarlucas ! appreciate this.

PR #268, opened 2019-02-09
mavlink_start_checksum() eliminate possibility of unaligned memory access

dagar:
@peterbarker here's another one. https://github.com/ArduPilot/pymavlink/pull/267

image

### DIFF
--- a/generator/C/include_v2.0/mavlink_helpers.h
+++ b/generator/C/include_v2.0/mavlink_helpers.h
@@ -470,12 +470,16 @@ union __mavlink_bitfield {
 
 MAVLINK_HELPER void mavlink_start_checksum(mavlink_message_t* msg)
 {
-	crc_init(&msg->checksum);
+	uint16_t crcTmp = 0;
+	crc_init(&crcTmp);
+	msg->checksum = crcTmp;
 }
 
 MAVLINK_HELPER void mavlink_update_checksum(mavlink_message_t* msg, uint8_t c)
 {
-	crc_accumulate(c, &msg->checksum);
+	uint16_t checksum = msg->checksum;
+	crc_accumulate(c, &checksum);
+	msg->checksum = checksum;
 }
 
 /*


peterbarker:
@dagar Have to check this one with tridge.  Looks logically correct, I'd just like to check for performance concerns.  Correctness can wait ;-)

peterbarker:
Sorry, should provide an update here.

I asked tridge to look at this and he decided to check the root cause of the problem - packing these structures.  He was generally confident that with a little finessing we could stop packing these structures at all without impact, negating the need to create stack variables here.

Brooce:
I just wanted to confirm that the bug exists (caused a HardFault handler ISR on my custom STM32 application) and the proposed solution solved it for Mavlink1 libraries.  For Mavlink2 however, the problem occurred in another place - in "mavlink_finalize_message_buffer" function. I have fixed it with analogous approach at the topmost level.

I'm not sure how to PR into a submodule and don't have the time to deal with it now, so here's just a quick little patch if anyone wants to make use of it:
[0001-Mavlink_helpers-Added-separate-variable-for-checksum.patch.txt](https://github.com/ArduPilot/pymavlink/files/2956520/0001-Mavlink_helpers-Added-separate-variable-for-checksum.patch.txt)

dagar:
@Brooce that one should have been fixed in https://github.com/ArduPilot/pymavlink/pull/267.
I've rebased this PR on current master, could you try again?

dagar:
> I asked tridge to look at this and he decided to check the root cause of the problem - packing these structures. He was generally confident that with a little finessing we could stop packing these structures at all without impact, negating the need to create stack variables here.

@peterbarker any update on this effort? Unless it's currently underway I would propose merging this in the meantime. The performance difference seemed to be negligible.

tridge:
I do want to switch to unpacked structures in the future, but I agree with this as a quick fix. Merging

dagar:
Thanks, I'll update https://github.com/mavlink/mavlink now.

julianoes:
Related to #5.

PR #269, opened 2019-02-13
Support reserved/default params in generated libraries

hamishwillee:
XML definition files can now declare command params as reserved and define a default value of 0 or NaN. If a param is not declared it is assumed to be reserved with default of zero.

What this PR does is implement that behaviour and generate improved docs:
- adds a numeric prefix (e.g. **3. Param description**) for all `param` value descriptions. This makes it much easier to work out what text maps to which param, and makes it clear that you're in a param even if it has no description.
- adds standard text for reserved values
- Creates "reserved" params for index values that were not declared in XML. This improves docs, but also means we can remove all reserved params from XML if we want. 

The change is in the stored description in the parser, so will flow to all outputs.

It doesn't prevent someone declaring param 6 or 5 as reserved values with NaN default value (which is invalid when sending in COMMAND_INT). My understanding is that the schema should prevent that in validation, but the parser should respect XML  if validation is not run (?)

### DIFF
--- a/generator/mavparse.py
+++ b/generator/mavparse.py
@@ -316,6 +316,26 @@ class MAVXML(object):
         f.close()
    
 
+        #Post process to add reserved params (for docs)
+        for current_enum in self.enum:
+            if not 'MAV_CMD' in current_enum.name:
+                continue
+            print(current_enum.name)
+            for enum_entry in current_enum.entry:
+                print(enum_entry.name)
+                if len(enum_entry.param) == 7:
+                    continue
+                params_dict=dict()
+                for param_index in range (1,8):
+                    params_dict[param_index] = MAVEnumParam(param_index, label='', units='', enum='', increment='', 
+                                                        minValue='', maxValue='', default='0', reserved='True')
+
+                for a_param in enum_entry.param:
+                    params_dict[int(a_param.index)] = a_param
+                enum_entry.param=params_dict.values()
+                
+
+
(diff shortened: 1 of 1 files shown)


WickedShell:
I'm not a huge fan of this. Because these aren't really messages there is no way to enforce this at a library level, which means that NaN's are still easily not emitted and still require the implementor of sending the message to insert the NaN's correctly. The attribute makes sense on a top level message, but on something that is encoded into COMMAND_LONG, COMMAND_INT, MISSION_ITEM, MISSION_ITEM_INT among others I think there are to many ways for that to not reflect the default of NaN.

hamishwillee:
@WickedShell I _am_ convinced this is a good idea for exactly the reason you are not.

Because there is no way to enforce this in (all) the API(s) the best thing you can do is add the documentation at point-of-use (in generated library) - to increase the chance that the end user will see the info and "do the right thing". 

In addition (separate from the NaN issue), this ensures that the in-source documentation is actually usable by adding visible index. Currently if someone omits a param definition it is impossible to tell what definition applies to what param in the docs). So even if we were to remove the "default value" from this, I'd still say this is real improvement.

hamishwillee:
@peterbarker @tridge Can I please get this reviewed? 

I spoke to @WickedShell  and he is conceptually OK with this - his review did not understand that this change is about in-source docs more than anything else.

peterbarker:
Merged, thanks!

PR #271, opened 2019-02-18
mavlink_helpers: Tidy parser example

hamishwillee:
This partially fixes the in-source docs for `mavlink_parse_char()` and friends in the 1.0 and 2.0 generator - the example used incorrect names for parameters  and did not pass in the status variable. 

The description of the problem is in https://github.com/mavlink/mavlink/issues/1086

The issue has a separate comment below which I believe is invalid - @peterbarker can you confirm that the assignments of these are in `mavlink_frame_char_buffer()` (which is called by  `mavlink_parse_char()`).

> Some members of mavlink_status_t are never assigned - for example `msg_received`, `buffer_overrun`, `parse_error` (this is very misleading!). Value from `parse_error` is assigned to `packet_rx_drop_count` - https://github.com/mavlink/c_library_v2/blob/d240d0986710045663894aebcea89e71ce981ee4/mavlink_helpers.h#L832

### DIFF
--- a/generator/C/include_v1.0/mavlink_helpers.h
+++ b/generator/C/include_v1.0/mavlink_helpers.h
@@ -211,34 +211,12 @@ MAVLINK_HELPER void mavlink_update_checksum(mavlink_message_t* msg, uint8_t c)
  * parser in a library that doesn't use any global variables
  *
  * @param rxmsg    parsing message buffer
- * @param status   parsing starus buffer 
+ * @param status   parsing status buffer 
  * @param c        The char to parse
  *
  * @param r_message NULL if no message could be decoded, otherwise the message data
  * @param r_mavlink_status if a message was decoded, this is filled with the channel's stats
  * @return 0 if no message could be decoded, 1 on good message and CRC, 2 on bad CRC
- *
- * A typical use scenario of this function call is:
- *
- * @code
- * #include <mavlink.h>
- *
- * mavlink_status_t status;
- * mavlink_message_t msg;
- * int chan = 0;
- *
- *
- * while(serial.bytesAvailable > 0)
- * {
- *   uint8_t byte = serial.getNextByte();
- *   if (mavlink_frame_char(chan, byte, &msg, &status) != MAVLINK_FRAMING_INCOMPLETE)
- *     {
- *     printf("Received message with ID %d, sequence: %d from component %d of system %d", msg.msgid, msg.seq, msg.compid, msg.sysid);
- *     }
- * }
- *
- *
- * @endcode
  */
 MAVLINK_HELPER uint8_t mavlink_frame_char_buffer(mavlink_message_t* rxmsg, 
                                                  mavlink_status_t* status,
@@ -434,9 +412,9 @@ MAVLINK_HELPER uint8_t mavlink_frame_char_buffer(mavlink_message_t* rxmsg,
  * message is received it is copies into *r_message and the channel's status is
  * copied into *r_mavlink_status.
  *
- * @param chan     ID of the current channel. This allows to parse different channels with this function.
- *                 a channel is not a physical message channel like a serial port, but a logic partition of
- *                 the communication streams in this case. COMM_NB is the limit for the number of channels
+ * @param chan     ID of the channel to be parsed.
+ *                 A channel is not a physical message channel like a serial port, but a logical partition of
+ *                 the communication streams. COMM_NB is the limit for the number of channels
  *                 on MCU (e.g. ARM7), while COMM_NB_HIGH is the limit for the number of channels in Linux/Windows
  * @param c        The char to parse
  *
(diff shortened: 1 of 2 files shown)


generator/C/include_v1.0/mavlink_helpers.h
```diff
@@ -206,31 +206,32 @@ MAVLINK_HELPER void mavlink_update_checksum(mavlink_message_t* msg, uint8_t c)
...
  * @param status   parsing starus buffer 
  * @param c        The char to parse
  *
- * @param returnMsg NULL if no message could be decoded, the message data else
- * @param returnStats if a message was decoded, this is filled with the channel's stats
+ * @param r_message NULL if no message could be decoded, otherwise the message data
+ * @param r_mavlink_status if a message was decoded, this is filled with the channel's stats
  * @return 0 if no message could be decoded, 1 on good message and CRC, 2 on bad CRC
  *
  * A typical use scenario of this function call is:
```
peterbarker: No, this is a typical use scenario for a completely different function call :-)

generator/C/include_v1.0/mavlink_helpers.h
```diff
@@ -206,31 +206,32 @@ MAVLINK_HELPER void mavlink_update_checksum(mavlink_message_t* msg, uint8_t c)
 }
 
 /**
- * This is a varient of mavlink_frame_char() but with caller supplied
+ * This is a variant of mavlink_frame_char() but with caller supplied
  * parsing buffers. It is useful when you want to create a MAVLink
  * parser in a library that doesn't use any global variables
  *
  * @param rxmsg    parsing message buffer
  * @param status   parsing starus buffer
```
peterbarker: More bugs....

generator/C/include_v1.0/mavlink_helpers.h
```diff
@@ -430,32 +431,33 @@ MAVLINK_HELPER uint8_t mavlink_frame_char_buffer(mavlink_message_t* rxmsg,
  * 2 (MAVLINK_FRAMING_INCOMPLETE, MAVLINK_FRAMING_OK or MAVLINK_FRAMING_BAD_CRC)
  *
  * Messages are parsed into an internal buffer (one for each channel). When a complete
- * message is received it is copies into *returnMsg and the channel's status is
- * copied into *returnStats.
+ * message is received it is copies into *r_message and the channel's status is
+ * copied into *r_mavlink_status.
  *
  * @param chan     ID of the current channel. This allows to parse different channels with this function.
```
peterbarker: More language bugs...

generator/C/include_v1.0/mavlink_helpers.h
hamishwillee: Fixed

generator/C/include_v1.0/mavlink_helpers.h
hamishwillee: I've tidied the wording, but I still don't find the explanation _useful_.

What I "think" is that a channel represents a logical connection to an desired endpoint - e.g. a particular GCS or a mavlink camera. Basically anything that needs a specific set of messages streamed to it.  The channel is then just a handle to that logical "thing" for stats and the message signature. 
The text says that it isn't about serial port, but generally I can't see how you would be able to differentiate any incoming information from multiple channels unless they were separated by physical or logical port.

If you think that's about right, please let me know. Shouldn't stop this merging though.

hamishwillee:
@peterbarker Thanks for the review. I "fixed" those bugs - if you're OK with what I did, I'm good for you to merge because as you say, it is better than what came before.

generator/C/include_v1.0/mavlink_helpers.h
peterbarker: channels aren't necessarily *to* endpoints.  Sometimes channels aren't even used for actual communication, just fiddling with messages.  And often there are multiple system IDs that a channel may correspond to.

I usually think about channels as "parallel channels of communication".  Each channel has separate space for incoming and outgoing stuff.  chan just switches between them (it really is just an offset at heart....)

hamishwillee: Hi @peterbarker ,

> channels aren't necessarily _to_ endpoints. Sometimes channels aren't even used for actual communication, just fiddling with messages.

Can you give me an example?

> And often there are multiple system IDs that a channel may correspond to.

What do you mean. Shouldn't each physical system have just one id? Or are you talking about destinations - ie a channel might talk to multiple endpoints specifically (in addition to broadcasting). 

The [official docs](https://mavlink.io/en/mavgen_c/#streams) throw in the term "stream", but seem to mean the same thing as you indicated.

> The C MAVLink library utilizes a "channel" metaphor to allow for simultaneous processing of multiple, independent MAVLink streams in the same program. It is therefore important to use the correct channel for each operation as all receiving and transmitting functions provided by MAVLink require a channel.
>
> If only one MAVLink stream exists, channel 0 should be used by specifying the MAVLINK_COMM_0 constant.

So what would you see as a logical stream? Can you help me with some examples of things that it makes sense to split up into logical independent handling. And the converse, why not just use one channel?

peterbarker:
On Sun, 24 Feb 2019, Hamish Willee wrote:

>       channels aren't necessarily to endpoints. Sometimes channels aren't even used for actual communication, just fiddling with messages.
> Can you give me an example?

Anything which finalizes a message but doesn't send it using a _send 
function.

>       And often there are multiple system IDs that a channel may correspond to.
> What do you mean. Shouldn't each physical system have just one id? Or are you talking about destinations - ie a channel might talk to multiple endpoints
> specifically (in addition to broadcasting).

destinations.

>       The C MAVLink library utilizes a "channel" metaphor to allow for simultaneous processing of multiple, independent MAVLink streams in the same
>       program. It is therefore important to use the correct channel for each operation as all receiving and transmitting functions provided by
>       MAVLink require a channel.
>
>       If only one MAVLink stream exists, channel 0 should be used by specifying the MAVLINK_COMM_0 constant.
> 
> So what would you see as a logical stream? Can you help me with some examples of things that it makes sense to split up into logical independent handling.
> And the converse, why not just use one channel?

For receiving on different links you must use multiple channels.  Each 
uart gets a channel, for example.

PR #272, opened 2019-02-20
Null checks on C helper function mavlink_frame_char_buffer

guyf2010:
In issue #169 a couple of NULL checks were added preventing a second message or status object from being required when running mavlink_frame_char_buffer in the C library. In two places a memcpy was run with the second message as a destination without checking it exists.

### DIFF
--- a/generator/C/include_v2.0/mavlink_helpers.h
+++ b/generator/C/include_v2.0/mavlink_helpers.h
@@ -779,7 +779,9 @@ MAVLINK_HELPER uint8_t mavlink_frame_char_buffer(mavlink_message_t* rxmsg,
 				}
 			}
 			status->parse_state = MAVLINK_PARSE_STATE_IDLE;
-			memcpy(r_message, rxmsg, sizeof(mavlink_message_t));
+			if(NULL != r_message){
+				memcpy(r_message, rxmsg, sizeof(mavlink_message_t));
+			}
 		}
 		break;
 	case MAVLINK_PARSE_STATE_SIGNATURE_WAIT:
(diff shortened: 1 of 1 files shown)


generator/C/include_v2.0/mavlink_helpers.h
```diff
@@ -779,7 +779,9 @@ MAVLINK_HELPER uint8_t mavlink_frame_char_buffer(mavlink_message_t* rxmsg,
 				}
 			}
 			status->parse_state = MAVLINK_PARSE_STATE_IDLE;
-			memcpy(r_message, rxmsg, sizeof(mavlink_message_t));
+			if(NULL != r_message){
```
peterbarker: Please don't reverse the order of comparisons.

Also, a space between the if and opening pren...

... and between the closing pron and opening brace.

hamishwillee:
@peterbarker Does this change also need to happen in https://github.com/ArduPilot/pymavlink/tree/master/generator/C/include_v1.0 , 

> The parameter description for r_message is just completely bogus, too

Thanks. I'm aware we need to come back to this. Created https://github.com/ArduPilot/pymavlink/issues/289 as a reminder.

generator/C/include_v2.0/mavlink_helpers.h
guyf2010: Implemented the suggested changes in commit 10b6e1fe96550e4cd7655f04c72618deedb47227

(It appears the date on my VM was wrong after resuming, so the commit shows up as from before this comment)

The order of comparisons was done to be consistent with the other null checks. In this commit, I updated the other null comparisons to not be reversed either.

magicrub:
lgtm

peterbarker:
Merged, thanks!

PR #280, opened 2019-03-02
mavgen_wlua: add missing v2.0 fields

gswly:
before:

after:

This PR allows the Wireshark plugin to parse and display two fields that were added in Mavlink v2.0:
* incompatibility flag
* compatibility flag

These are currently skipped. In particular, adding the incompatibility flag is the first step in implementing signature parsing (which is also missing).


gswly:
@amilcarlucas hi, if you merge this trivial PR i (or anyone else) can implement signature parsing too, the last 2.0 feature missing from the Wireshark dissector.

amilcarlucas:
Thanks, merged.

PR #281, opened 2019-03-02
mavgen_wlua: fix parsing of v2.0 message ids

gswly:
before:

after:

Message ids of v2.0 frames are parsed in a completely wrong way, consequently payloads are not identified and frames are not displayed correctly.

Message IDs are saved in up to 3 bytes (24 bits) (1 byte in v1.0, 3 bytes in v2.0).

What the original author did:
* take the 1st byte, right shift by 8, i.e. 255 >> 8, always resulting in zero.
* take the 2nd byte, right shift by 16, i.e. 255 >> 16, always resulting in zero.
* take the 3rd byte, summing all three up and putting the result in a byte (mess is complete).

What i did:
* take all 3 bytes together, parse them as a single 24bits, little endian, unsigned number
* put the result in a uint24.

PS: the Wireshark Lua dissector has a huge quantity of bugs, and it's a pity since it could be a good entry point for people who wants to approach the protocol and understand how it works.


peterbarker:
Merged, thanks.  In terms of using this as a tool - do we have modern docs on how to take these LUA bindings and use them in wireshark?

PR #284, opened 2019-03-02
mavgen_wlua: link message id to corresponding hex dump

gswly:
same as #279 but for the message id field.


peterbarker:
merged, thanks.

PR #285, opened 2019-03-04
fix MAVLink_bad_data being initialized with str instead of bytearray

jo-m:
This broke MAVProxy under Python3, see this issue: https://github.com/ArduPilot/MAVProxy/issues/600


peterbarker:
Merged, thanks!

jo-m:
@peterbarker Hi
Currently MAVproxy still has this bug, as the fixed pymavlink (2.3.5) has not yet been released on pypi. Can you tell me when a release is scheduled?

Thanks

PR #288, opened 2019-03-14
Create a regression test for mavlogdump.py

peterbarker:
Sadly, running this shows me that the Python3 dump is an even 18 hours off from correct - or at least what Python2 and my memory thinks is correct.


peterbarker:
The time thing has been fixed.  But now we have precision problems between Python2 and Python3 float outputs.

PR #296, opened 2019-03-29
mavprase.py: stop magically prefixing descriptions with their index

peterbarker:
This breaks MAVProxy's waypoint display as it string-matches against
this description.

The value should be available as separate metadata anyway so any user of
the metadata should be able to synthesise this themselves.

@hamishwillee https://github.com/ArduPilot/pymavlink/pull/269/files#diff-47dc7d87fdc00f28b7816950fba41176R162 added this line.  I wish to remove it.  Was this vital for something you were doing?

### DIFF
--- a/generator/mavparse.py
+++ b/generator/mavparse.py
@@ -159,7 +159,6 @@ class MAVEnumParam(object):
             self.description = 'Reserved (default:%s)' % self.default
         else:
             self.description = description
-        self.description = '%s: %s' % (self.index, self.description)
 
 class MAVEnumEntry(object):
     def __init__(self, name, value, description='', end_marker=False, autovalue=False, origin_file='', origin_line=0):


tridge:
good catch!

hamishwillee:
@peterbarker That line basically numbers every param in the docs for commands. That makes it a hell of a lot easier to work out what param value matches what param index. Particularly in the case where there is no description.

I did consider separately adding the index to output, but I think I decided that this was a better approach as it pushed the description into all generated param docs for all languages.

So not vital, but useful. If it can be kept that would be better but if not then nothing will fall apart.

peterbarker:
On Thu, 28 Mar 2019, Hamish Willee wrote:

> @peterbarker That line basically numbers every param in the docs for commands. That makes it a hell of a lot easier to work out what param value matches
> what param index. Particularly in the case where there is no description.
> 
> I did consider separately adding the index to output, but I think I decided that this was a better approach as it pushed the description into all
> generated param docs for all languages.
> 
> So not vital, but useful. If it can be kept that would be better but if not then nothing will fall apart.

To be clear: this is basically only supposed to appear in comments within 
the generated code but turned out to modify non-metadata (code!) instead?

The fix, then, is probably to modify each of the generators, correct?

hamishwillee:
@peterbarker No. I would have said that `self.description` IS metadata. Why is it being used in code?

peterbarker:
On Thu, 28 Mar 2019, Hamish Willee wrote:

> @peterbarker No. I would have said that self.description IS metadata. Why is it being used in code?

It is programatically accessible in both the Python code 
(mav_cmd[cmd].param.get(5,'')) and the C bindings when you ask for *all* 
of the data to be made available (rather than the limited set of 
information we normally include for embedded use).

hamishwillee:
It never occurred to me that you'd expose the description in code for actual use - "risky" to rely on pattern matching of something that can easily change - i.e. no one is protecting the descriptions in the source XML other than sanity checking that the meaning doesn't change.

I'm not precious about the line of code and you have a reasonable argument that it is part of the API. But if it were me I'd be removing the dependency in MAVProxy.

peterbarker:
On Thu, 28 Mar 2019, Hamish Willee wrote:

> It never occurred to me that you'd expose the description in code for actual use - "risky" to rely on pattern matching of something that can easily change
> - i.e. no one is protecting the descriptions in the source XML other than sanity checking that the meaning doesn't change.

It was like that when I got here :-)

> I'm not precious about the line of code and you have a reasonable argument that it is part of the API. But if it were me I'd be removing the dependency in
> MAVProxy.

That was actually the first way I attacked it.  Then I figured to get at 
the description in the code you have to go via the index anyway, so having 
it in the description wasn't good.

I'll make up some patches for at least some of the generators...

hamishwillee:
OK!

tridge:
merging now to prevent breakage of mavproxy

Issue #306, opened 2019-04-24
tcpin-created TCP socket can't be reused


alexkrob:
If I create a `mavlink_connection` with `tcpin:{localhost}:{port}`, send some data over it, call the `close()` method (which should close the socket), and attempt to create a new `mavlink_connection` with the same argument for `device` (i.e., try and use the same port), I get the following exception: `[Errno 48] Address already in use`.

alexkrob:
I created [this pull request](https://github.com/ArduPilot/pymavlink/pull/307) in order to address the issue. It's just a one-liner.

peterbarker:
Thanks!

PR #307, opened 2019-04-24
Re-use TCP ports created with tcpin.

alexkrob:
This commit is a one-line swap in order to set the socket option for address reusal before binding to the specified port. Originally this led to a bug where the port would refuse additional connections if the process that created the first connection was still active, even if the socket had been closed. By swapping these lines, the ports can be re-used. This fix was tested on OSX Mojave v10.14.4.

This is intended to address #306.


peterbarker:
Thanks!  This makes it look like the others (setsockopt before bind) in both pymavlink and MAVProxy, so it looks correct to me.

Note you can automatically close issues with the magic:

Closes #306 

text

PR #311, opened 2019-05-01
Fix TCP reconnect issues

OXINARF:
This is going to help with Azure autotest running - actually it has been running for around a week without any more errors with TCP exceptions (and reboot is always detected).


peterbarker:
I've been running this for the last couple of days.

PR #312, opened 2019-05-06
MAVLink v2 C library efficiency improvements

amilcarlucas:
**This was originaly posted by @olliw42 on mavlink/mavlink#1127 . I moved it here where it belongs**

although saying to be "highly optimized for resource-constrained systems with limited RAM and flash memory" the MAVLink C lib fails to be efficient in most aspects, flash, ram, stack, computation time, which can be relevant especially for resource constrained embeeded applications. This PR adds extensions which much improve the situation. It is offered just in case anyone might be interested.

It deals with only C and v2.0.

It is used in two applications, the STorM32 gimbal controller, which now makes heavy use of the new 'light' features, and the UC4H MavlinkBridge, which doesn't use them but uses the convenience functions (and thus kind of tests compatibility). That is, various, but by far not all aspects of the lib are tested, e.g. I don't use signatures at all, etc.

The additions/extensions are essentially non-intrusive, with maybe one exception.

I couldn't figure out how to properly deal with git submodules, so I simply created a new folder "pymavlink-light", into which I put all my changes, and added a mavgenerate_light.py which is to be used. For this PR I recreated all anew in order to achieve a series of commits which hopefully helps following.

The additions/extensions in detail:

**1) Receiving memory usage**

Currently the memory usage in receiving a message may be sketched as follows

user data -> payload_t -> msg_t of user -> buffer of user -> uart tx buf

There are several versions, pack, encode, send etc, which allow several things to do and skip/simplify some steps, e.g. the "msg_t of user" can be send directly to the "uart tx buf".

However, what is not possible is to put the provided message data directly into the "buffer of user". In many cases that's however what is desired since the "msg_t of user" typically has no further use. Msg_t is a quite big structure, and thus costly.

I thus added functions mavlink_msg_MSG_pack_txbuf() to the messages .h files. For instance, for the heartbeat there is now a function mavlink_msg_heartbeat_pack_txbuf(), which skips the need of "msg_t of user".

This is a totally non-intrusive addition, as it doesn't affect any of the other code.

**2) Parsing memory usage**

Currently the memory usage in parsing may be sketched as follows:

uart rx buf -> msg_t in COMM buffer -> msg_t of user -> payload_t

In many cases the double use of a msg_t structure is not needed, since the parsing could directly go into "msg_t of user". Msg_t is a quite big structure, and thus costly.

The new NULL check changes in mavlink_frame_char_buffer() allow to do that, however the whole thing is still quite back-through-the-door, e.g. the user needs to define his own mavlink_get_channel_buffer() to refer to his own msg_t variable, or keep using the COMM_BUF, etc..

I thus added a new function mavlink_parse_nextchar() which is much simpler to use, and also results in a logically structured API (i.e. mavlink_frame_char_buffer() is supposed to use now mavlink_parse_nextchar(), like mavlink_frame_char() uses mavlink_parse_nextchar(), and mavlink_parse_char() uses mavlink_frame_char()).

This easily could be made to be a totally non-intrusive addition, as it doesn't have to affect any of the other code. As is, it is intrusive however, since it also adds the handling of the msg_entry extension (see below). (is just one line of code, so easily removed if not wanted)

**3) FLASH usage and computation time**

MAVLINK_MESSAGE_CRCS is typically quite big, which has two implications, lots of flash used, and in addition the search by mavlink_get_msg_entry() takes long.

In many cases the application doesn't use all messages of a dialect, but just a relatively small subset, 10-20% or so. Having MAVLINK_MESSAGE_CRCS to only contain the entries for the used (incoming) messages can thus lower flash usage as well as make the search faster. With the lib as is this can be achieved in several ways, e.g. by doing one's own xml, or by copying the dialect's original MAVLINK_MESSAGE_CRCS and deleting all unused entries, etc. But all that is not very convenient and can add maintenance burdens.

I thus added a message_entries.h file to the respective dialect subfolder, which offers macros for each message entry. This allows the user to very simply build her own MAVLINK_MESSAGE_CRCS.

The gains can be quite substantial. E.g. for the STorM32, which has a decent MAVLink support, this saved 2760 bytes of flash. Also the computation time for mavlink_get_msg_entry() has become more than twice faster.

This is a totally non-intrusive addition, as it doesn't affect any of the other code.

**4) msg_entry_t, computation time**

Handling of incoming messages requires parsing the data stream, as well as some sort of identification if the received message should be handled (kind of the first step of routing). Currently, with the lib as is, the mavlink_get_msg_entry() is called at least twice: It is called in the parsing function, and then has to be called by the identification code to get the target system/component.

Note that mavlink_get_msg_entry() does a search which by principle has to be rated as costly (for as long as quantum computers have not yet become reality for us), and double-doing the search isn't clever.

I thus added a msg_entry field to the message_t structure as additional metadata. 

This allows to set it in the parsing function, and then later be reused whenever needed without any additional search. Since it's just a pointer it's insignificant memory wise in comparison to the size of the message_t structure.

This easily could be made to be totally non-intrusive, as it doesn't have to affect any of the other code. As is, it is intrusive however, since it adds this additional field to the message_t structure. I note that message_t has already some meta data, so this shouldn't be a shocking approach, and that I would not think that this is very troublesome (unless the user does "strange things" like relying explicitly on the length of the structure). Adding of the msg_entry to the message_t structure could have been avoided by returning it by the parse function (which would need double pointers, argh), but I just think that the approach I've taken is clean and simple.

Along with that change I've added helper functions to get the target system/component. 

I note that this improvement should be very significant also and especially for "big" applications, which use huge dialects and many MAVLink channels with lots of incoming traffic, and thus many search events. I thus think that this improvement should also be seriously considered for the c++ lib.

**5) MAVLINK_USE_CHAN_FUNCTIONS**

In some analogy to MAVLINK_USE_CONVENIENCE_FUNCTIONS I've added a MAVLINK_USE_CHAN_FUNCTIONS, which wraps all functions which use chan and/or the COMM buffers. 

It is defined when MAVLINK_COMM_NUM_BUFFERS is larger than zero. That way the user simply has to do a #define MAVLINK_COMM_NUM_BUFFERS 0 to ensure that none of all that is used. 

This is an essentially non-intrusive addition, since it normally doesn't affect any of the code.

**6) Formatting, few rearrangements, few renamings**

I've made plenty of changes which result in more properly formatted .h files. I've also rearranged some functions/structures to arrive at what I consider more logical and user friendly. For instance, I've moved the COMM related functions in mavlink_helpers.h to the top. Finally, I renamed some local variable names (only very few instances).

All of this is supposed to be totally non-intrusive, as it shouldn't affect the code behavior at all.

As said, I just share in case it might be of interest. Do what you want with it :).

Cheers, and as always, have fun, Olli

### DIFF
--- a/generator/C/include_v2.0/checksum.h
+++ b/generator/C/include_v2.0/checksum.h
@@ -32,17 +32,16 @@ extern "C" {
  * @param data new char to hash
  * @param crcAccum the already accumulated checksum
  **/
-static inline void crc_accumulate(uint8_t data, uint16_t *crcAccum)
+static inline void crc_accumulate(uint8_t data, uint16_t* crcAccum)
 {
-        /*Accumulate one byte of data into the CRC*/
-        uint8_t tmp;
+    /*Accumulate one byte of data into the CRC*/
+    uint8_t tmp;
 
-        tmp = data ^ (uint8_t)(*crcAccum &0xff);
-        tmp ^= (tmp<<4);
-        *crcAccum = (*crcAccum>>8) ^ (tmp<<8) ^ (tmp <<3) ^ (tmp>>4);
+    tmp = data ^ (uint8_t)(*crcAccum &0xff);
+    tmp ^= (tmp<<4);
+    *crcAccum = (*crcAccum>>8) ^ (tmp<<8) ^ (tmp <<3) ^ (tmp>>4);
 }
 #endif
-
 
 /**
  * @brief Initiliaze the buffer for the X.25 CRC
@@ -51,9 +50,8 @@ static inline void crc_accumulate(uint8_t data, uint16_t *crcAccum)
  */
 static inline void crc_init(uint16_t* crcAccum)
 {
-        *crcAccum = X25_INIT_CRC;
+    *crcAccum = X25_INIT_CRC;
 }
-
 
 /**
  * @brief Calculates the X.25 checksum on a byte buffer
@@ -64,14 +62,13 @@ static inline void crc_init(uint16_t* crcAccum)
  **/
 static inline uint16_t crc_calculate(const uint8_t* pBuffer, uint16_t length)
 {
-        uint16_t crcTmp;
-        crc_init(&crcTmp);
-	while (length--) {
-                crc_accumulate(*pBuffer++, &crcTmp);
-        }
-        return crcTmp;
+    uint16_t crcTmp;
+    crc_init(&crcTmp);
+    while (length--) {
+        crc_accumulate(*pBuffer++, &crcTmp);
+    }
+    return crcTmp;
 }
-
 
 /**
  * @brief Accumulate the X.25 CRC by adding an array of bytes
@@ -82,12 +79,12 @@ static inline uint16_t crc_calculate(const uint8_t* pBuffer, uint16_t length)
  * @param data new bytes to hash
  * @param crcAccum the already accumulated checksum
  **/
-static inline void crc_accumulate_buffer(uint16_t *crcAccum, const char *pBuffer, uint16_t length)
+static inline void crc_accumulate_buffer(uint16_t* crcAccum, const char* pBuffer, uint16_t length)
 {
-	const uint8_t *p = (const uint8_t *)pBuffer;
-	while (length--) {
-                crc_accumulate(*p++, crcAccum);
-        }
+    const uint8_t* p = (const uint8_t*)pBuffer;
+    while (length--) {
+        crc_accumulate(*p++, crcAccum);
+    }
 }
 
 #if defined(MAVLINK_USE_CXX_NAMESPACE) || defined(__cplusplus)
--- a/generator/C/include_v2.0/mavlink_helpers.h
+++ b/generator/C/include_v2.0/mavlink_helpers.h
@@ -254,6 +254,76 @@ MAVLINK_HELPER uint8_t _mav_trim_payload(const char* payload, uint8_t length)
 }
 
 /**
+ * @brief Finalize a MAVLink message into a transmit buffer
+ *
+ * This function calculates the checksum and sets length and aircraft id correctly.
+ * It assumes that the payload are already correctly set in the transmit buffer.
+ *
+ */
+MAVLINK_HELPER uint16_t mavlink_finalize_message_txbuf(char* txbuf, mavlink_status_t* status, uint8_t system_id, uint8_t component_id,
+                                uint32_t msgid, uint8_t min_length, uint8_t length, uint8_t crc_extra)
+{
+    uint8_t signature[MAVLINK_SIGNATURE_BLOCK_LEN];
+    uint8_t header_len, signature_len;
+    bool mavlink1 = (status->flags & MAVLINK_STATUS_FLAG_OUT_MAVLINK1) != 0;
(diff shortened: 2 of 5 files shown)


olliw42:
many thx, amilcarlucas :+1:

Linjieqiang:
Great.

tridge:
@Linjieqiang we can't really consider this without it at least passing CI tests. The PR doesn't even build at this stage.
Olliw is right that there is plenty of room to improve the efficiency of the generated code, but to get that accepted into this repo someone would need to put in quite a lot of effort to ensure it doesn't break existing usage.

peterbarker:
Firstly, thanks @olliw42 for these sorts of patches.  It's been quite illuminating taking a critical eye to the existing code - and these patches to it.

I understand that "working with upstream" (i.e. this pymavlink branch) can be frustrating, and that pulling patches apart into "easily reviewable" chunks is sometimes difficult and time-consuming.  IMO time spent on that is a good investment, however, because the patch-set presented here has many logical changes mixed throughout it, combined with whitespace and code re-arrangement.  Anything to be merged from this PR would probably need to be split into a separate PR first.

(1) I could really use `mavlink_msg_vision_position_estimate_pack_txbuf` , for example.  But what's this `mavlink_finalize_message_txbuf`?  How does it differ from `mavlink_finalize_message_buffer`?!

(2) I quite like the new entry point concept!  It looks to me that with ArduPilot's usage of `mavlink_parse_char` that the `status` object gets updated for every single character we process - but we only ever check that value on successfully finding a message.
I think the naming really needs looking at, however; `mavlink_parse_char` wraps `mavlink_frame_char`, so it would be very bad if we were to reverse that and have `mavlink_frame_char_buffer` wrap `mavlink_parse_char`!  A `mavlink_parse_char_buffer` wrapping `mavlink_frame_char_buffer` would also seem a logical addition.  I've created https://github.com/ArduPilot/ardupilot/pull/19746 based on this discussion.

(3) doesn't *seem* to reduce the maintenance overhead vs creating a custom XML file with only the messages you desire in it.  Perhaps this would be better done as an input to the generator - basically a whitelist of messages that should be included in the generated files?

(4) has the disadvantage that users who are packing messages will end up with a valid message whose msg_entry_t is still nullptr.  Maybe that's OK?  I don't mind this addition to the structure - but it could also just be considered a parsing artifact and go into the status structure instead?

(5) will potentially break users who are supplying their own buffers and statuses via `MAVLINK_GET_CHANNEL_STATUS` and `MAVLINK_GET_CHANNEL_BUFFER`.

(6) makes this PR a real pain to review :-)

@amilcarlucas - I really don't know whether much is going to happen in terms of review/merge of this PR.  It really needs to be pulled apart into 7 different PRs.  It is also conflicting now.

olliw42:
@peterbarker 

first, many thx for looking at this

second, you note however that this is nearly 3 years old now? And that it was basically turned down some time ago by some of your ArduPilot fellow devs?

third, very largely because there was no action but rather - let's call it - not so much positive enthusiasm, I eventually went on and have created my own library, fastMavlink it's called. I thus can't respond to any of your questions or any discussion on this old PR because all details simply have slipped my mind. The points raised here formed the basis for fastMavlink, but this null-statement is really all I can contribute now to the discussion on this PR. I would have to spend significant time to dive into it again, and - I'm sorry to say and hope you may understand - it just doesn't make any sense since I'm fully switched to fastMavlink.

sorry for that, but that's how it is :) 
windows of opportunity open and close

PR #314, opened 2019-05-08
mavwp: cast rally values to int to make mavlink message packing happy

peterbarker:


WickedShell:
Surely that's a cast to int not float?

peterbarker:
On Tue, 7 May 2019, WickedShell wrote:

> Surely that's a cast to int not float?

I was close!

Fixed.

PR #315, opened 2019-05-10
Fix occasional truncate params

hamishwillee:
This reverts part of a change my commit https://github.com/ArduPilot/pymavlink/commit/2e19ed5dd8ef60dbf07c0571d4beb8e2c552cdae 

Basically I did not understand that the parser gets the param description in chunks and assembles them. As a result a small percentage of param descriptions got truncated.

For example, in asluav.h

image

This fixes that errors.


hamishwillee:
@peterbarker Would be good if you could merge this. Reverts a bug I introduced a little while ago. Diff shows that this fixes the bug, and doesn't do anything else.

PR #290, opened 2019-03-15
Generator: Update Java MAVLink2 support

kellyschrock:
Get Java support for Mavlink2 working

PR #317, opened 2019-05-13
Generate Valid Java Files for Mavlinkv2

HefnySco:
This is heavily inspired and build on [this pull request](https://github.com/ArduPilot/pymavlink/pull/290/files).
I use this patch with dronekit-android and it works fine.

### DIFF
--- a/generator/java/lib/Messages/MAVLinkPayload.java
+++ b/generator/java/lib/Messages/MAVLinkPayload.java
@@ -25,6 +25,9 @@ public class MAVLinkPayload {
     
     public final ByteBuffer payload;
     public int index;
+    // Avoid bad index value when the previous parameter has zero.
+    // e.g ACK msg with command = 11 and all remaining butes are zeros.
+    public boolean allZeros = false;
 
     public MAVLinkPayload(int payloadSize) {
        if(payloadSize > MAX_PAYLOAD_SIZE) {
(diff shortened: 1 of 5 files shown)


magicrub:
you seem to detect the zero values, but then don't modify the length. @kellyschrock 's version doesn't either

magicrub:
parser state MAVLINK_PARSE_STATE_GOT_PAYLOAD needs the zero fill like here:
https://github.com/ArduPilot/pymavlink/blob/master/generator/C/include_v2.0/mavlink_helpers.h#L740

HefnySco:
@magicrub
In packing commands. I add zeros rather than changing the length. Not the best way to do it but it works as a first iteration of code.

HefnySco:
@magicrub
Regarding this point https://github.com/ArduPilot/pymavlink/pull/317#issuecomment-491931635
The packet is zero filled, the allZero flag in MAVLinkPayload.java returns zero when the message class tried to read remaining data.

magicrub:
Ok, so it zero fills on the input, but what about the output?

generator/mavgen_java.py
```diff
@@ -84,7 +84,7 @@ def generate_CRC(directory, xml):
         data = data & 0xff; //cast because we want an unsigned type
         int tmp = data ^ (crcValue & 0xff);
         tmp ^= (tmp << 4) & 0xff;
-        crcValue = ((crcValue >> 8) & 0xff) ^ (tmp << 8) ^ (tmp << 3) ^ ((tmp >> 4) & 0xf);
+        crcValue = ((crcValue >> 8) & 0xff) ^ (tmp << 8) ^ (tmp << 3) ^ ((tmp >> 4) & 0xff);
```
magicrub: are you sure this is supposed to be 0xFF?

magicrub:
@HefnySco Take a look at [this commit](https://github.com/aionrobotics/DroneKit-Android/commit/261223556c3d04bc0a2edbd2db935d984af8c2aa) I made. The magic is implementing [mav_trim_payload()](https://github.com/ArduPilot/pymavlink/blob/8481afb968c384fb3f260ea5fe6e0b2e0f0b1c41/generator/C/include_v2.0/mavlink_helpers.h#L327) and fixing the CRC to compensate for a different length. I haven't tested this, is that something you could try?

generator/java/lib/Messages/MAVLinkPayload.java
```diff
@@ -51,85 +57,192 @@ public void resetIndex() {
...
         short result = 0;
-        result |= payload.get(index + 0) & 0xFF;
-        index+= 1;
-        return result; 
+        try {
+            result |= payload.get(index + 0) & 0xFF;
+            index+= 1;
+
+        }
+        catch (final Exception ex)
```
WickedShell: Last I profiled this, this is a very expensive path to include. (It's been awhile since I've profiled it admittedly). Any time you hit a 0 trimmed packet you will be allocating a new exception object, and have to enter the handler, which ends up being slow and memory intensive.

It's usually easier to just either check that the index is in range first, or always ensure that the buffer has the expected minimum number of bytes allocated and filled out.

magicrub: I agree. You don't want to be throwing constant exceptions like this

generator/mavgen_java.py
HefnySco: I guess this is an error typo

generator/java/lib/Messages/MAVLinkPayload.java
HefnySco: true.
@WickedShell  this is a valid concern.
guys, please check this [update](https://github.com/ArduPilot/pymavlink/pull/317/commits/6fd6ff15eef0fcf84440ab025cabba80ad5797bc)

I added:
final byte[] array = payload.array();
            final int len = array.length;
            if (index + LENGTH > len)
            {
                allZeros = true;
                return result;
            }

where LENGTH = 1,2,4.

generator/java/lib/Messages/MAVLinkPayload.java
```diff
@@ -51,85 +57,262 @@ public void resetIndex() {
...
+            result |= (payload.get(index + 5) & 0xFF) << 40;
+            result |= (payload.get(index + 6) & 0xFF) << 48;
+            result |= (payload.get(index + 7) & 0xFF) << 56;
+            index += 8;
+        }
+        catch (final Exception ex)
+        {
+            allZeros = true;
+        }
+        Double.longBitsToDouble(result);
```
magicrub: You're forgetting the result.

magicrub:
Have you looked at my output zero padding changes?

generator/java/lib/Messages/MAVLinkPayload.java
HefnySco: I have pushed a fix... thank you

HefnySco: > Have you looked at my output zero padding changes?

not yet ... I will take a look today.
Thanks

HefnySco:
> Have you looked at my output zero padding changes?


not yet ... I will take a look today.
Thanks

HefnySco:
> @HefnySco Take a look at [this commit](https://github.com/aionrobotics/DroneKit-Android/commit/261223556c3d04bc0a2edbd2db935d984af8c2aa) I made. The magic is implementing [mav_trim_payload()](https://github.com/ArduPilot/pymavlink/blob/8481afb968c384fb3f260ea5fe6e0b2e0f0b1c41/generator/C/include_v2.0/mavlink_helpers.h#L327) and fixing the CRC to compensate for a different length. I haven't tested this, is that something you could try?


Mavlink2 trims zero parameters from commands.
i.e. if a command has 5 parameters the last 3 parameters are all zeros then only parameters 1 and 2 are sent. 

### **question**.

if I understand your code correctly you remove zero bytes what not considering parameters boundaries.
i.e. having the same example of 5 parameters command. and the 3rd,4,5 are all zeros _and the second parameter = 2  although it is non-zero, but if it is an integer parameter that means bytes 2,3,4 of the second parameter will be removed as well._

If this is valid then I need to go back [here](generator/java/lib/Messages/MAVLinkPayload.java) and modify the reading logic again. **as I assume if an integer parameter = 2 then I expect to find bytes [2,0,0,0] the last three bytes of zeros exists.**

**Kindly advise**

@magicrub  @WickedShell   guys what do you think ?

HefnySco:
> > @HefnySco Take a look at [this commit](https://github.com/aionrobotics/DroneKit-Android/commit/261223556c3d04bc0a2edbd2db935d984af8c2aa) I made. The magic is implementing [mav_trim_payload()](https://github.com/ArduPilot/pymavlink/blob/8481afb968c384fb3f260ea5fe6e0b2e0f0b1c41/generator/C/include_v2.0/mavlink_helpers.h#L327) and fixing the CRC to compensate for a different length. I haven't tested this, is that something you could try?
> 
> Mavlink2 trims zero parameters from commands.
> i.e. if a command has 5 parameters the last 3 parameters are all zeros then only parameters 1 and 2 are sent.
> 
> ### **question**.
> if I understand your code correctly you remove zero bytes what not considering parameters boundaries.
> i.e. having the same example of 5 parameters command. and the 3rd,4,5 are all zeros _and the second parameter = 2 although it is non-zero, but if it is an integer parameter that means bytes 2,3,4 of the second parameter will be removed as well._
> 
> If this is valid then I need to go back [here](generator/java/lib/Messages/MAVLinkPayload.java) and modify the reading logic again. **as I assume if an integer parameter = 2 then I expect to find bytes [2,0,0,0] the last three bytes of zeros exists.**
> 
> **Kindly advise**
> 
> @magicrub @WickedShell guys what do you think ?

I searched in C code and found that @magicrub way to trim zeros is the correct one.
I also debugged and found that partial parameters are sent i.e. int(2) = array.bytes (2) instead of array.bytes (2,0,0,0)

so I am fixing the code accordingly.

Thanks

generator/mavgen_java.py
```diff
@@ -364,24 +373,47 @@ def generate_MAVLinkMessage(directory, xml_list):
...
         }
         crc.finish_checksum(msgid);
     }
 
+    /**
+     * Return length of actual data after triming zeros at the end.
+     * @param payload
+     * @return minimum length of valid data
+     */
+    private int mavTrimPayload (final MAVLinkPayload payload)
```
HefnySco: @magicrub **mavTrimPayload** has been implemented here.

generator/mavgen_java.py
```diff
@@ -364,24 +373,47 @@ def generate_MAVLinkMessage(directory, xml_list):
...
+    
+
     /**
     * Encode this packet for transmission.
     *
     * @return Array with bytes to be transmitted
     */
     public byte[] encodePacket() {
-        byte[] buffer = new byte[10 + len + 2];
+        final int lenTrimmed = mavTrimPayload (payload);
```
HefnySco: **lenTrimmed = mavTrimPayload (payload);**   is used instead of len in **encodePacket()**

magicrub:
looking good! Have you tested it? I found that SITL vibration msg is an easy one to see the "clipping" values become zero

HefnySco:
> looking good! Have you tested it? I found that SITL vibration msg is an easy one to see the "clipping" values become zero

yes I use it with Andruav telemetry & QGroundControl.

magicrub:
I'm not seeing any packets with id > 255 getting through. The zero padding seems to work though

HefnySco:
My test is that I add the lib to dronekit fir android and run Andruav enabled telemetry then connected from otherside using qground via Andruav web. That ensures that all traffic that qground handles goes via dronekit avd the updated Mavlink that is all test I have done.
I am open to any suggestions to finalize this  update.

magicrub:
I'm unable to receive a packet with msgid > 255. Can you test that now?

rrr6399:
I submitted this pull request for an updated version of Hunter's mods just for comparison in case there is anything useful in it to use:
https://github.com/ArduPilot/pymavlink/pull/320
I've tested it in the field with MAVLink v1 and v2

HefnySco:
> I'm unable to receive a packet with msgid > 255. Can you test that now?

I ran the following sample:

                    msg_esc_telemetry_1_to_4 m = new msg_esc_telemetry_1_to_4();
                    m.count =new int[] {1,2,3,4};
                    m.voltage=new int[] {21,22,23,24};
                    m.totalcurrent = new int[] {10,20,30,40};
                    m.rpm = new int[] {11,12,13,14};
                    MAVLinkPacket mp = m.pack();
                    byte [] b = mp.encodePacket();

                    Parser p = new Parser();

                    for (int i=0;i<b.length;++i)
                    {
                        MAVLinkPacket pk = p.mavlink_parse_char( b[i] & 0xff);
                        if (pk != null)
                        {
                            MAVLinkMessage msg = pk.unpack();
                            Log.d("msg","reconstructed");
                        }
                    }

and the messages reconstructed again with the same input values and zero values at the end were truncated.

magicrub:
I'm favoring https://github.com/ArduPilot/pymavlink/pull/320 because it handles the short incoming packets much better and also does extensions

magicrub:

PR #318, opened 2019-05-15
Fix up CS generator file paths

### LINKED ISSUE #151: C# generator problem
Generating C# source via mavlink generators gives the following errors:
```
This compiler is provided as part of the Microsoft (R) .NET Framework, but only supports language versions up to C# 5, which is no longer the latest version. For compilers that support newer versions of the C# programming language, see http://go.microsoft.com/fwlink/?LinkID=533240

error CS2001: Source file 'CS\common\ByteArrayUtil.cs' could not be found
error CS2001: Source file 'CS\common\FrameworkBitConverter.cs' could not be found
error CS2001: Source file 'CS\common\Mavlink.cs' could not be found
Error
Error: Compilation failed. (1)
```
This was reported in mavlink/mavlink:
* https://github.com/mavlink/mavlink/issues/813
* https://github.com/mavlink/mavlink/issues/482

hamishwillee:
Fixes #151 

This just fixes up the generator so that it searches for file paths of include files relative to the current directory (originally suggested by @iliasam.

Testing shows that this allows the libraries to be built for both mavlink 2 and 1. Note that the main output files generated before this fix at the same - but you now get new files mavlink.dll, mavlink.pdb, mavlink.xml.
My assumption is that the defect meant that you couldn't link and debug into the library, but you could still use it. 

@peterbarker @amilcarlucas Can you please review.

### DIFF
--- a/generator/mavgen_cs.py
+++ b/generator/mavgen_cs.py
@@ -328,7 +328,8 @@ using System.Reflection;
     
     generatedCsFiles = [ serfilename, structsfilename]
     
-    includedCsFiles =  [ 'CS/common/ByteArrayUtil.cs', 'CS/common/FrameworkBitConverter.cs', 'CS/common/Mavlink.cs'  ]
+    cur_dir = os.path.dirname(os.path.abspath(__file__)) + '/'
+    includedCsFiles = [cur_dir + 'CS/common/ByteArrayUtil.cs', cur_dir + 'CS/common/FrameworkBitConverter.cs', cur_dir + 'CS/common/Mavlink.cs' ]
     
     outputLibraryPath = os.path.normpath(dir + "/mavlink.dll")
     


OXINARF:
@hamishwillee I don't have any comment for the PR itself (I think C# support is terrible out of date and generating a compiled library instead of headers like C/C++ does), but I think you are making some confusion regarding MAVLink2 support: neither C# nor Java support it; they might not break when built with `--wire-protocol=2.0` but that doesn't mean they have support for the protocol.

hamishwillee:
Hi @OXINARF,

Thanks! I based my assumption of MAVLink 2 on:
- Presence of the MAVLink 2 messages in the generated output - which I now realise might just appear because of the parsing phase
- presence of mavlink 2 start byte and version in generated output 
- @auturgy told me that Mission Planner supports MAVLink 2 and uses CS.

However looking at the code I can't see anything to indicate that MAVLink 2 packet format is understood - or anything about signing, so my feeling is you are probably right. Any idea how Mission Planner might be doing this/where the confusion comes?

> and generating a compiled library instead of headers like C/C++ does

The compiled library only gets generated with this fix. Can this be used without that library?

Upshot, this can still merge, but I think I should make sure the docs say CS does not support MAVLink 2 - at least until I understand how Mission Planner can possibly work with this

Cheers
Hamish

auturgy:
@meee1 can you provide some input here?

meee1:
my library supports mavlink2, and is inspired by the c version. my version has never been part of pymavlink or the generator in this repo. it does however use the generator with changes that suited my needs.

hamishwillee:
@meee1 Thanks for the clarification. Any chance of contributing that back into Pymavlink/generator?

hamishwillee:
@peterbarker Can you review this. It is a trivial change to fix up a bug in MAVLink 1 C# generation. This library is apparently out of date etc, but if we're not going to remove it, then this fix is pretty much a no-brainer.

hamishwillee:
@peterbarker ?

generator/mavgen_cs.py
```diff
@@ -328,7 +328,8 @@ def generate(basename, xml):
     
     generatedCsFiles = [ serfilename, structsfilename]
     
-    includedCsFiles =  [ 'CS/common/ByteArrayUtil.cs', 'CS/common/FrameworkBitConverter.cs', 'CS/common/Mavlink.cs'  ]
+    cur_dir = os.path.dirname(os.path.abspath(__file__)) + '/'
```
peterbarker: This should be `os.path.join`

This should also look something like this:
```
     includedCsFilesCommon =  [ 'ByteArrayUtil.cs', 'FrameworkBitConverter.cs', 'Mavlink.cs'  ]
     cur_dir = os.path.dirname(os.path.abspath(__file__))
     incudeCSFiles = [ os.path.join([cur_dir, "CS", "Common", x]) for x in includedCsFilesCommon ]
```

... but I'm only really fussed with using the appropriate join function.

CornerOfSkyline:
This issue still in master, this pr should be merge.@peterbarker

peterbarker:
@CornerOfSkyline does this PR work for you?

peterbarker:
@hamishwillee it looks like Mr Oborne fixed this as part of adding mavlink2 support.

hamishwillee:
I think that very likely. Closing as fixed. Can be reopened if someone can replicate it against current CS generator.

PR #320, opened 2019-05-17
Java generated compatible with MAVLink V1 and V2

rrr6399:
I've tested this generator with MAVLink V1 and V2 messaging.


magicrub:
This is now the third Java MAVLink v2 PR. Please take a look at PR https://github.com/ArduPilot/pymavlink/pull/317

magicrub:
does this handle zero padding and short packets?

rrr6399:
It should handle truncated incoming packets since it allocates a ByteBuffer that is the maximum size of the MAVLink 2 payload. It doesn't appear that there is any logic to truncate the outgoing packets to the last non-zero value. It just sends the entire set of bytes for the given message.  I'll see what'd take to support that.

I'll compare with the other pull request as well.

generator/java/lib/Messages/MAVLinkPayload.java
```diff
@@ -72,7 +77,7 @@ public short getShort() {
         return result;
     }
 
-    public int getUnsignedShort(){
+    public int getUnsignedShort() {
```
HefnySco: I think these functions does not handle truncated values.
payload.get(index) will give exception if index is out of range.

rrr6399: The payload buffer is sized to the largest possible payload so those functions will just return 0.

rrr6399: This is what I did to deal with truncated payloads:
https://github.com/ArduPilot/pymavlink/blob/4935e9cf039050093467e8b2670c0fb1b836e650/generator/java/lib/Messages/MAVLinkPayload.java#L34

HefnySco: oic you preallocated with full size so it will never get out of index.

magicrub:
I'm favoring this PR, but it would be nice if the incoming parsing state machine was wasn't two totally separate copies. Harder to maintain that way. Just use the single one and check the isMavlink2 flag to jump to the new states

magicrub:
I'm working on combining the parser state machine

magicrub:
Almost there.. @rrr6399 can you take a look [at this commit](https://github.com/magicrub/pymavlink/commits/java-mavlink2) and cherry pick it into your branch?
- adds common mavlink 1 and 2 inbound parsing (copies c version style)
- adds output zero trim

magicrub:
I also rebased your branch onto master

magicrub:

PR #324, opened 2019-06-04
mavparm: exclude MIS_TOTAL from being loaded from files by default

peterbarker:
Also tidy and sort the list


magicrub:
how does this work?

peterbarker:
> how does this work?

Very well?

This module is used by `mavproxy_param.py` when you (e.g.) issue `param load mav.parm`.  This PR just includes `MIS_TOTAL`

I'm guessing since `CMD_` doesn't exist as a parameter in the modern ArduPilot that it may have been renamed to `MIS_` at some stage - and this file wasn't updated...

PR #326, opened 2019-06-18
Improve License description

amilcarlucas:
Add missing Copyright information

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -1,6 +1,19 @@
 This repository contains the generator for the MAVLink protocol. The generator itself is
-(L)GPL v3 licensed, while the generated code is subject to different licenses:
+(L)GPL v3 licensed, while the generated code is subject to the MIT license.
 
+Copyright 2010-2019
+Andrew Tridgell
+Lorenz Meier
+Peter Barker
+Thomas Gubler
+James Goppert
+Andreas Daniel Antener
+Amilcar do Carmo Lucas
+Pierre Kancir
+Vladimir Ermakov
+Max Odnovolyk
+Michael Day
+and other contributors
 
 ========================================================================================
 


peterbarker:
Doesn't actually clarify enough here.  Parts are strictly GPLv3.

https://github.com/ArduPilot/pymavlink/blob/master/generator/mavparse.py#L6

amilcarlucas:
I guess the MIT part is the generated files PLUS mavutil.py

@peterbarker Looks like some parts are LGPLv3 and other GPLv3

But what I would realy like to clarify here is the generated files plus the mavutil.py because those are AFAIK the ones used by dronekit python and MavROS

COPYING
```diff
@@ -1,6 +1,19 @@
 This repository contains the generator for the MAVLink protocol. The generator itself is
-(L)GPL v3 licensed, while the generated code is subject to different licenses:
-
+(L)GPL v3 licensed, while the generated code is subject to the MIT license.
```
OXINARF: I don't understand the need for this change, what is written below is much more clear - especially since what it says below is that the generated code is only under MIT license if the product where it is included is licensed as MIT itself.

COPYING
```diff
@@ -1,6 +1,19 @@
...
+
+Copyright 2010-2019
```
OXINARF: There really is no reason to put some names holding copyright here. If copyright names are wanted to be tracked then a CONTRIBUTORS.md file should be created with **everyone** inside.

COPYING
amilcarlucas: The legal department of Volkswagen requires that the copyright information be available in a single file on the same repo as the source code. This was my attempt at achieving that, having it all in the same file as the license.

The MIT license requires to have copyright information in the license, we do not have that now.

COPYING
amilcarlucas: For Ardupilot we have 412 Contributors, many of them with a username, but not a proper name.
Is there a way of automaticaly collect all of them and place them in a generated file?

I simply added the names of the 20 to contributors here, but I did it by hand, tracking down the real names behind the usernames.

COPYING
OXINARF: > The legal department of Volkswagen requires that the copyright information be available in a single file on the same repo as the source code.

A single file doesn't mean in the same file as the license.

> The MIT license requires to have copyright information in the license, we do not have that now.

No, it doesn't. What it says is that the Copyright notice needs to be included in any copy of the code. If there is no Copyright notice then there isn't anything to copy and anyone would still be complying with the license.

COPYING
OXINARF: > For Ardupilot we have 412 Contributors, many of them with a username, but not a proper name.

That's life when you don't track down the contributors, which we don't.

> Is there a way of automaticaly collect all of them and place them in a generated file?

Likely there are tools available for that. However, at least in ArduPilot, that is unlikely to collect all contributors as code was imported into Git from somewhere else.

> I simply added the names of the 20 to contributors here, but I did it by hand, tracking down the real names behind the usernames.

Besides the top 20 changing, it isn't fair to the other contributors. At a minimum, we should allow other contributors to be added to the file - and with the number of them, we clearly need a separate file.

Also, you seem to be concerned with the Volkswagen legal department: I don't see how listing a very small number of copyright holders is acceptable.

amilcarlucas: > Also, you seem to be concerned with the Volkswagen legal department: I don't see how listing a very small number of copyright holders is acceptable.

Well most FOSS projects only list a single copyright owner, so by listing 20 we are already on the safe side.

amilcarlucas:
After talking with tridge, I did a couple of PRs that should encounter less resistance :)

PR #333, opened 2019-06-25
Add flight mode 24 ZIGZAG to mavutil

hfujikawa77:
added flight mode 24 ZIGZAG to mavutil.py to resolve the following issue. 
#332

PR #334, opened 2019-06-25
mavutil: add ZIGZAG mode (number 24)

peterbarker:


peterbarker:
Closed in favour of #333

PR #336, opened 2019-06-26
Add a port in udp dissector in mavgen_wlua.py

Marxlp:
As shown in the [simulation section of devguide](https://dev.px4.io/en/simulation/), the PX4 also transports data on 14580 UDP port to communicate with external developer APIs by default.


amilcarlucas:
Thanks

PR #338, opened 2019-07-11
mavgen.py: correct handling of includes

olliw42:
corrects handling of mulit-level includes.
Enables to e.g. separate out minimal.xml by including it in common.xml. Corresponding xmls are included for testing in folder xmls_for_testing.
The code is highly verbose due to many prints, I left them in as their outputs may help to better understand what is going on. Will be removed of course. Also, improvements are possible, like the all_files or the args handling. Easy to do later.
Note: the generated files for ardupilot are essentially identical (the comments of some MAV_CMD entries seem to be more correctly handled now), but the generated files for the sub folders may not be identical, e.g. differ in their xcrcs or payload length. However the differences are as they are supposed to be according to their position in the include tree!
Have fun.

### DIFF
--- a/generator/mavgen.py
+++ b/generator/mavgen.py
@@ -21,8 +21,8 @@ from . import mavparse
 schemaFile = os.path.join(os.path.dirname(os.path.realpath(__file__)), "mavschema.xsd")
 
 # Set defaults for generating MAVLink code
-DEFAULT_WIRE_PROTOCOL = mavparse.PROTOCOL_1_0
-DEFAULT_LANGUAGE = 'Python'
+DEFAULT_WIRE_PROTOCOL = mavparse.PROTOCOL_2_0 ##mavparse.PROTOCOL_1_0
+DEFAULT_LANGUAGE = 'C' ##'Python'
 DEFAULT_ERROR_LIMIT = 200
 DEFAULT_VALIDATE = True
 DEFAULT_STRICT_UNITS = False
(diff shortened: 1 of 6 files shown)


magicrub:
you've changed every line of every file. please check your whitespace line endings.

olliw42:
???
clicking on mavgen.py comes out great for me, and the other files didn't had been there so all diff
I fail to see your issue
btw, EOL could be done by .gitattributes, so I rather would suggest you get this right ;)

magicrub:
Ahh I see, you've added new redundant xmls. That's not a good idea.

olliw42:
doing a PR is an even worse idea

magicrub:
It's too bad your people skills aren't as good as your coding skills

olliw42:
???
in this case it's definitely the other way around, you certainly could have been more appreciative and constructive, but sure, it's me, how convenient ...

auturgy:
That was quick - thanks for looking at this @olliw42
@magicrub there’s a little background on this.
@hamishwillee has discovered that whilst we build included xml’s down to 5 levels it appears we only generate the crc correctly for the first  level.
This is related to discussion on how to use minimal and standard xml’s a bit better, with the possible result that a bunch of messages may be moved out of common into them, pulled back via includes where needed (ie common will include standard, standard will include minimal).

Regards,

James

lvale:
@peterbarker

hamishwillee:
Er, I was working on this same problem, and not reading my messages quickly enough.

So something that does much the same job is https://github.com/ArduPilot/pymavlink/pull/339

generator/mavgen.py
```diff
@@ -21,8 +21,8 @@
 schemaFile = os.path.join(os.path.dirname(os.path.realpath(__file__)), "mavschema.xsd")
 
 # Set defaults for generating MAVLink code
-DEFAULT_WIRE_PROTOCOL = mavparse.PROTOCOL_1_0
-DEFAULT_LANGUAGE = 'Python'
+DEFAULT_WIRE_PROTOCOL = mavparse.PROTOCOL_2_0 ##mavparse.PROTOCOL_1_0
```
hamishwillee: These bits should not be changed without general agreement. Not relevant to fix, and not something all users would be happy with

generator/mavgen.py
```diff
@@ -65,38 +65,114 @@ def mavgen(opts, args):
...
 
     def expand_includes():
-        """Expand includes in current list of all files, ignoring those already parsed."""
-        for x in xml[:]:
-            for i in x.include:
-                fname = os.path.join(os.path.dirname(x.filename), i)
-
-                # Only parse new include files
-                if fname in all_files:
+        """Expand includes. We do have worked out all required root files before in main, and have their parsed objects in the xml list."""
```
hamishwillee: This is nicer than what was present. I copied this/retrofitted it to my fix.

generator/mavgen.py
```diff
@@ -65,38 +65,114 @@ def mavgen(opts, args):
...
+        for i in range(MAXIMUM_INCLUDE_FILE_NESTING):
+            if expand_oneiteration(): 
+                break
+            
+        if mavparse.check_duplicates(xml):
+            sys.exit(1)
+            
+    def update_includes():
+        """Update includes. We do have worked out all required files before in expand_includes() and have their parsed objects in the xml list."""
+        print("\n\n-------------------------------------------------------")
```
hamishwillee: Nice to keep output simple

hamishwillee:
Firstly, thanks for posting this @olliw42. Yes, we do need test files for a change like this, but perhaps we would agree where they should live first. 
I also worked on this problem in https://github.com/ArduPilot/pymavlink/pull/339 - I didn't see that you had posted a fix here until too late.

Obviously I prefer my own code :-) It is more plodding, but I think easier to understand. 

Anyway, upshot is that @peterbarker is main maintainer of this toolchain. He can decide which, if either, he likes. I'm going to run a few more tests on mine now.

peterbarker:
> Anyway, upshot is that @peterbarker is main maintainer of this toolchain. 

No, that would definitely still be @tridge.

peterbarker:
@olliw42 Hamish's PR is simpler than this.

What advantages do you see in the two-phase approach you've used here?

olliw42:
> @olliw42 Hamish's PR is simpler than this.
What advantages do you see in the two-phase approach you've used here?

I believe it gives a correct and complete algorithm. Let me explain by example of a compiler. A compiler you expect to exactly accept all valid inputs and reject all invalid inputs, with nothing in between. The algorithm's ability to throw corresponding errors is not just convenience but a signature of its correctness and completness. The algorithm suggested here is by construction doing exactly this, as indicated also by its ability to throw corresponding errors. The simple one-phase approach obviously doesn't only not throw similar errors, but - I believe - isn't correct and complete in the said sense. E.g., I believe you will find it hard to add the error detection provided in the 2nd phase of the two-phase algorithm, which would be a simple argument confirming my believe. In other words, the one-phase algorithm might work for the tested cases, but the test cases were obviously limited in scope and I believe that it fails in general because it isn't correct and complete by construction. In even simpler words: The next issue is just waiting around. I might be wrong, and you are of course free to prove me wrong.

Btw, the initial code was even simpler than either of the newer ones. Simplicity and provided advantages are maybe not the best yardsticks here, but producing correct results in even the general case, and to throw errors whenever a correct result can't be produced.

That's at least what I think.
Have fun

hamishwillee:
@olliw42 Do you plan on looking at the open issues on this that various people marked up? At the moment I don't think Peter can merge it even if he wanted. 

Or I can try tidy up the bits that are going to prevent it being merged, correct or otherwise?

peterbarker:
This was merged as #436 - thanks!

PR #339, opened 2019-07-12
CRCS and lengths for mulitple inclusion levels

hamishwillee:
The current code to include multiple dialect levels is a bit broken (which didn't matter too much because no one was using this). 
The problem is that dialect CRC lists have to include the CRCS of all included files (because the various bits are only included/#defined in the C code once). The previous version only added the previous parent CRCs.

What this code does is add all the CRCs for all the parents. It does this by calculating all the parents of each file in the first iteration. Then it works runs check to find all the children from all the parents. At that point it iterates the xml files a final time to add CRCs and other files.

I'm still running tests, but I think this is good for review.

@peterbarker ?

### DIFF
--- a/generator/mavgen.py
+++ b/generator/mavgen.py
@@ -136,7 +136,7 @@ def mavgen(opts, args):
                 x.message_target_system_ofs.update(xml_dict[child].message_target_system_ofs)
                 x.message_target_component_ofs.update(xml_dict[child].message_target_component_ofs)
                 x.message_names.update(xml_dict[child].message_names)
-                x.largest_payload = max(x.largest_payload, xml[-1].largest_payload)
+                # x.largest_payload = max(x.largest_payload, xml[-1].largest_payload)
 
 
 
--- a/generator/mavgen.py
+++ b/generator/mavgen.py
@@ -128,6 +128,7 @@ def mavgen(opts, args):
 
         # Update parents with child CRCs, message lengths, etc. 
         for x in xml:
+            
             for child in x.child:
                 x.message_crcs.update(xml_dict[child].message_crcs)
                 x.message_lengths.update(xml_dict[child].message_lengths)
@@ -136,7 +137,7 @@ def mavgen(opts, args):
                 x.message_target_system_ofs.update(xml_dict[child].message_target_system_ofs)
                 x.message_target_component_ofs.update(xml_dict[child].message_target_component_ofs)
                 x.message_names.update(xml_dict[child].message_names)
-                # x.largest_payload = max(x.largest_payload, xml[-1].largest_payload)
+                x.largest_payload = max(x.largest_payload, xml_dict[child].largest_payload)
 
 
 
@@ -187,10 +188,6 @@ def mavgen(opts, args):
     # expand includes
     expand_includes()
 
-    # work out max payload size across all includes
-    largest_payload = max(x.largest_payload for x in xml) if xml else 0
-    for x in xml:
-        x.largest_payload = largest_payload
 
     if mavparse.check_duplicates(xml):
         sys.exit(1)
--- a/generator/mavgen.py
+++ b/generator/mavgen.py
@@ -64,38 +64,81 @@ def mavgen(opts, args):
             print("WARNING: Unable to load XML validator libraries. XML validation will not be performed", file=sys.stderr)
             opts.validate = False
 
+
     def expand_includes():
-        """Expand includes in current list of all files, ignoring those already parsed."""
-        for x in xml[:]:
-            for i in x.include:
-                fname = os.path.join(os.path.dirname(x.filename), i)
+        """Expand includes. XML files from arguments already parsed inot xml list."""
+        
+        def expand_oneiteration():
+            noincludeadded = True
 
-                # Only parse new include files
-                if fname in all_files:
-                    continue
-                all_files.add(fname)
+            for x in xml[:]:
+                try:
+                    if not x.parent:
+                        pass
+                except:
+                    x.parent=set()
 
-                # Validate XML file with XSD file if possible.
-                if opts.validate:
-                    print("Validating %s" % fname)
-                    if not mavgen_validate(fname):
-                        return False
-                else:
-                    print("Validation skipped for %s." % fname)
+                for i in x.include:
+                    fname = os.path.join(os.path.dirname(x.filename), i)
+                    # Only parse new include files
+                    if fname in all_files:
+                        continue
+                    all_files.add(fname)
+                    # Validate XML file with XSD file if possible.
+                    if opts.validate:
+                        print("Validating %s" % fname)
+                        if not mavgen_validate(fname):
+                            print("ERROR Validation of %s failed" % fname)
+                            exit(1)
+                    else:
+                        print("Validation skipped for %s." % fname)
 
-                # Parsing
-                print("Parsing %s" % fname)
-                xml.append(mavparse.MAVXML(fname, opts.wire_protocol))
+                    # Parsing
+                    print("Parsing %s" % fname)
 
-                # include message lengths and CRCs too
-                x.message_crcs.update(xml[-1].message_crcs)
-                x.message_lengths.update(xml[-1].message_lengths)
-                x.message_min_lengths.update(xml[-1].message_min_lengths)
-                x.message_flags.update(xml[-1].message_flags)
-                x.message_target_system_ofs.update(xml[-1].message_target_system_ofs)
-                x.message_target_component_ofs.update(xml[-1].message_target_component_ofs)
-                x.message_names.update(xml[-1].message_names)
+                    parsed_xml = mavparse.MAVXML(fname, opts.wire_protocol)
+                    #add all parent(s) into included file
+                    parsed_xml.parent=set()
+                    parsed_xml.parent.add(x.basename)
+                    parsed_xml.parent.update(x.parent)
+                    xml.append(parsed_xml)
+
+                    noincludeadded = False
+            return noincludeadded        
+                    
+        for i in range(MAXIMUM_INCLUDE_FILE_NESTING):
+            if expand_oneiteration(): 
+                break
+
+        xml_dict=dict()
+        # Add children to parent.child lists.
+        for x in xml:
+            xml_dict[x.basename] = x
+            try:
+                if not x.child:
+                    pass
+            except:
+                x.child=set()
+            for a_parent in x.parent:
+               for x_inner in xml:
+                   if x_inner.basename == a_parent:
+                       #x is a child of x_inner
+                       x_inner.child.add(x.basename)
+
+
+        # Update parents with child CRCs, message lengths, etc. 
+        for x in xml:
+            for child in x.child:
+                x.message_crcs.update(xml_dict[child].message_crcs)
+                x.message_lengths.update(xml_dict[child].message_lengths)
+                x.message_min_lengths.update(xml_dict[child].message_min_lengths)
+                x.message_flags.update(xml_dict[child].message_flags)
+                x.message_target_system_ofs.update(xml_dict[child].message_target_system_ofs)
+                x.message_target_component_ofs.update(xml_dict[child].message_target_component_ofs)
+                x.message_names.update(xml_dict[child].message_names)
                 x.largest_payload = max(x.largest_payload, xml[-1].largest_payload)
+
+
 
     def mavgen_validate(xmlfile):
         """Uses lxml to validate an XML file. We define mavgen_validate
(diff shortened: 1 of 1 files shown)


hamishwillee:
From Ollie - I have not yet analysed.

> @hamishwillee
as regards the generator, I think your version is NOT correct. Your main subfolder may come out correctly, but I think you miss to see that your include subfolders are going to be incorrect! E.g., for ardupilot.xml, the files in the ardupilot subfolder may be correct, but I don't think the files in the other subfolders will be identical to then you would start your generator with them as root!!! But this is how it has to be, anything else is a bug. Please carefully compare the result of my version to the result of your version as regards ALL generated files, compared to when the an included file is used in the generator. You when will also find that in my version the external correction for payload length is not needed, another indication of that it does something more correct. You generally can assume that there are reasons for what I do. :)

generator/mavgen.py
```diff
@@ -142,17 +186,8 @@ def mavgen_validate(xmlfile):
...
 
     # expand includes
-    for i in range(MAXIMUM_INCLUDE_FILE_NESTING):
-        len_allfiles = len(all_files)
-        expand_includes()
-        if len(all_files) == len_allfiles:
-            # stop when loop doesn't add any more included files
-            break
-
-    # work out max payload size across all includes
```
hamishwillee: This did max payload of all dialects. I think incorrect so removed. Should be max of payloads that current dialect includes. Note the assumption (which appears to be true by inspection) is that the the value is set by the top level dialect.

hamishwillee:
@olliw42 

>as regards the generator, I think your version is NOT correct. Your main subfolder may come out correctly, but I think you miss to see that your include subfolders are going to be incorrect! E.g., for ardupilot.xml, the files in the ardupilot subfolder may be correct, but I don't think the files in the other subfolders will be identical to then you would start your generator with them as root!!! But this is how it has to be, anything else is a bug. Please carefully compare the result of my version to the result of your version as regards ALL generated files, compared to when the an included file is used in the generator. 

I am fairly sure that my code is correct in this respect. 

I ran both my version and your version of the generator separately against each of the dialect files and compared between dialects and generator outputs. The output of both generators are the same when any of the files has the same root. 

> You when will also find that in my version the external correction for payload length is not needed, another indication of that it does something more correct. You generally can assume that there are reasons for what I do. :)

It may well be that Peter will like your approach more. That's fine by me - he will almost certainly require you to correct some of those issues highlighted first.

hamishwillee:
@peterbarker I've run a bunch more tests on this and I am convinced that:
1. Existing generation of current files will not be broken by this.
2. This correctly generates the CRCs and other values such that parents get all the values of their children.
3. I tested on Python, Java, C. 

So this is ready for your review.

generator/mavgen.py
```diff
@@ -64,38 +64,82 @@ def mavgen(opts, args):
             print("WARNING: Unable to load XML validator libraries. XML validation will not be performed", file=sys.stderr)
             opts.validate = False
 
+
```
peterbarker: Extra whitespace.

generator/mavgen.py
```diff
@@ -64,38 +64,82 @@ def mavgen(opts, args):
...
-                x.message_min_lengths.update(xml[-1].message_min_lengths)
-                x.message_flags.update(xml[-1].message_flags)
-                x.message_target_system_ofs.update(xml[-1].message_target_system_ofs)
-                x.message_target_component_ofs.update(xml[-1].message_target_component_ofs)
-                x.message_names.update(xml[-1].message_names)
-                x.largest_payload = max(x.largest_payload, xml[-1].largest_payload)
+        """Expand includes. XML files from arguments already parsed inot xml list."""
+        
+        def expand_oneiteration():
+            noincludeadded = True
```
peterbarker:
```suggestion
            includeadded = False
```

generator/mavgen.py
```diff
@@ -64,38 +64,82 @@ def mavgen(opts, args):
...
+                    print("Parsing %s" % fname)
+
+                    parsed_xml = mavparse.MAVXML(fname, opts.wire_protocol)
+                    #add all parent(s) into included file
+                    parsed_xml.parent=set()
+                    parsed_xml.parent.add(x.basename)
+                    parsed_xml.parent.update(x.parent)
+                    xml.append(parsed_xml)
+
+                    noincludeadded = False
```
peterbarker:
```suggestion
                    includeadded = True
```

generator/mavgen.py
```diff
@@ -64,38 +64,82 @@ def mavgen(opts, args):
...
-                # include message lengths and CRCs too
-                x.message_crcs.update(xml[-1].message_crcs)
-                x.message_lengths.update(xml[-1].message_lengths)
-                x.message_min_lengths.update(xml[-1].message_min_lengths)
-                x.message_flags.update(xml[-1].message_flags)
-                x.message_target_system_ofs.update(xml[-1].message_target_system_ofs)
-                x.message_target_component_ofs.update(xml[-1].message_target_component_ofs)
-                x.message_names.update(xml[-1].message_names)
-                x.largest_payload = max(x.largest_payload, xml[-1].largest_payload)
+        """Expand includes. XML files from arguments already parsed inot xml list."""
```
peterbarker: Please translate into English.

generator/mavgen.py
```diff
@@ -64,38 +64,82 @@ def mavgen(opts, args):
...
+                    parsed_xml.parent=set()
+                    parsed_xml.parent.add(x.basename)
+                    parsed_xml.parent.update(x.parent)
+                    xml.append(parsed_xml)
+
+                    noincludeadded = False
+            return noincludeadded        
+                    
+        for i in range(MAXIMUM_INCLUDE_FILE_NESTING):
+            if expand_oneiteration():
```
peterbarker:
```suggestion
            if not expand_oneiteration(): 
```

generator/mavgen.py
```diff
@@ -64,38 +64,82 @@ def mavgen(opts, args):
...
+                    
+        for i in range(MAXIMUM_INCLUDE_FILE_NESTING):
+            if expand_oneiteration(): 
+                break
+
+        xml_dict=dict()
+        # Add children to parent.child lists.
+        for x in xml:
+            xml_dict[x.basename] = x
+            try:
```
peterbarker: This would be better written as:

```
if getattr(x, 'child', None) is None:
    x.child = set()
```

generator/mavgen.py
```diff
@@ -64,38 +64,82 @@ def mavgen(opts, args):
...
-                x.message_target_component_ofs.update(xml[-1].message_target_component_ofs)
-                x.message_names.update(xml[-1].message_names)
-                x.largest_payload = max(x.largest_payload, xml[-1].largest_payload)
+        """Expand includes. XML files from arguments already parsed inot xml list."""
+        
+        def expand_oneiteration():
+            noincludeadded = True
+
+            for x in xml[:]:
+                try:
```
peterbarker: This can be rewritten using `getattr`; see example in other comment.

generator/mavgen.py
```diff
@@ -64,38 +64,82 @@ def mavgen(opts, args):
...
+                    x.parent=set()
+
+                for i in x.include:
+                    fname = os.path.join(os.path.dirname(x.filename), i)
+                    # Only parse new include files
+                    if fname in all_files:
+                        continue
+                    all_files.add(fname)
+                    # Validate XML file with XSD file if possible.
+                    if opts.validate:
```
peterbarker: We're also validating at the top level.  What's up with that?

generator/mavgen.py
```diff
@@ -64,38 +64,82 @@ def mavgen(opts, args):
...
+                    else:
+                        print("Validation skipped for %s." % fname)
+
+                    # Parsing
+                    print("Parsing %s" % fname)
+
+                    parsed_xml = mavparse.MAVXML(fname, opts.wire_protocol)
+                    #add all parent(s) into included file
+                    parsed_xml.parent=set()
+                    parsed_xml.parent.add(x.basename)
```
peterbarker: You appear to be adding apples and oranges to the same set here.  Probably just the basename is supposed to go in?

generator/mavgen.py
```diff
@@ -64,38 +64,82 @@ def mavgen(opts, args):
...
+                    parsed_xml.parent.update(x.parent)
+                    xml.append(parsed_xml)
+
+                    noincludeadded = False
+            return noincludeadded        
+                    
+        for i in range(MAXIMUM_INCLUDE_FILE_NESTING):
```
peterbarker: Throwing an error if we go too far would seem to be a good idea.

generator/mavgen.py
```diff
@@ -64,38 +64,82 @@ def mavgen(opts, args):
...
-                x.message_lengths.update(xml[-1].message_lengths)
-                x.message_min_lengths.update(xml[-1].message_min_lengths)
-                x.message_flags.update(xml[-1].message_flags)
-                x.message_target_system_ofs.update(xml[-1].message_target_system_ofs)
-                x.message_target_component_ofs.update(xml[-1].message_target_component_ofs)
-                x.message_names.update(xml[-1].message_names)
-                x.largest_payload = max(x.largest_payload, xml[-1].largest_payload)
+        """Expand includes. XML files from arguments already parsed inot xml list."""
+        
+        def expand_oneiteration():
```
peterbarker: is there some reason not to go fully recursive here?

generator/mavgen.py
hamishwillee: We've validated the top level files already. This is validating (just) the included files as well.

generator/mavgen.py
hamishwillee: No, this is actually intentional - though it might not be the best way to do it.

So what I am "ultimately" trying to achieve is that every file knows it's child includes all the way to the bottom, because it's CRCs etc are the CRSs of the files and its children. But when you're parsing from the top down, you don't know what the children are as you're going down through the iterations. 
However you can build up a picture of the parents, and from that you can re-parse the files to work out the children.

So what this does is, for the current file, create a parent attribute, and add the name of its parent (`x.basename`) and add the set of parents (`parsed_xml.parent.update(x.parent)`) (some duplication, but this catches the top level case where you have only one parent - the basename).

generator/mavgen.py
hamishwillee: We can't go too far - `MAXIMUM_INCLUDE_FILE_NESTING` is the maximum expansion.  

Not sure if you really mean throw, so have added this code on next line:
```
        if expand_oneiteration():
            print("ERROR Nesting level greater than maximum: %s" % MAXIMUM_INCLUDE_FILE_NESTING)
            exit(1)
```
So basically if there is more expansion that can be done this will do it and exit. Otherwise this will run but do nothing (no new files).

generator/mavgen.py
hamishwillee: Yes, effort and clarity. The code that gets expanded here is discrete and separate from the rest.

hamishwillee:
Thanks for your review @peterbarker. I've fixed the issues I could fixed and answered the open questions. I also ran flake8 and fixed all the whitespace errors. I left in the lines too long and module in wrong place issues, as they pre-exist anything I did.

I re-ran the tests for C library (only) and the output matches what I had before. I also tested with an extra level of include to verify that this catches the case where things are too deeply nested.

Thoughts?

olliw42:
I am sorry to be the pain in the ass, but I really don't understand what you guys drives you here. I would think that one wants to listen to arguments.

You don't want your "compiler" to only work for a particular limited structure of the include tree, you want your compiler to (1) correctly resolve ANY include tree which is resolvable and (2) throw an error for ANY include tree which is not resolvable. With really nothing inbetween like throwing an error for an include tree which is resolvable or not throwing an error for an include tree which is not resolvable (and producing code for this). 

This to me sounds so simple and obvious that I struggle to see why you guys even would bother to argue, or not see.

The above algorithm/code fails in this simple regard.

Given that, and given that there is a suggestion for an algorithm/code which exactly already does it, it's simply a mystery why one would want to spend time on an incomplete algorithm/code. There are certainly many algorithms to do the resolving of the include tree, and the suggested one is just one of them, but so far no one has provided a functional argument against it, which even more raises the question why one would spend time on this here. It seems the only argument for clinging to the above is "we don't like it, or the messenger". But not doing a correct code for that reason really doesn't point to me as the kid in the block.

Frankly, this effort here comes across as driven by stubbornness. Doing a correct code obviously isn't the driving force.

With having made this effort to drive you guys into a direction which results into proper solutions I will TO of the topic. Good luck to all.

hamishwillee:
Hi @olliw42,

I'm my fixing review feedback. That is what a good community member does to make it easier to integrate.

As to the rest, really not up to me. I don't really understand your code, and hence reasoning why it is better. That's why I deferred to Peter.

hamishwillee:
Superseded by #436

PR #342, opened 2019-07-31
mavlink_helpers.h: count dropped_packets & parse errors

kobytg17:
Currently those counters are not updated correctly.
Changes include counting dropped packets based on the expected seq number, and also
populate parse_error count on r_malink_conn.

Tested against the dropped packets counter in QGroundControl, which produced the same number.

### DIFF
--- a/generator/C/include_v2.0/mavlink_helpers.h
+++ b/generator/C/include_v2.0/mavlink_helpers.h
@@ -809,14 +809,15 @@ MAVLINK_HELPER uint8_t mavlink_frame_char_buffer(mavlink_message_t* rxmsg,
 	// If a message has been sucessfully decoded, check index
 	if (status->msg_received == MAVLINK_FRAMING_OK)
 	{
-		//while(status->current_seq != rxmsg->seq)
-		//{
-		//	status->packet_rx_drop_count++;
-		//               status->current_seq++;
-		//}
+		if (status->packet_rx_success_count == 0) { // Initial condition: If no packet has been received so far, drop count is undefined
+			status->packet_rx_drop_count = 0;
+		} else {
+			uint8_t expected_seq = status->current_rx_seq + 1;
+			uint8_t lost_packets = rxmsg->seq - expected_seq;
+			status->packet_rx_drop_count += lost_packets;
+		}
+
 		status->current_rx_seq = rxmsg->seq;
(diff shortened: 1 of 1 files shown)


generator/C/include_v2.0/mavlink_helpers.h
```diff
@@ -809,14 +809,15 @@ MAVLINK_HELPER uint8_t mavlink_frame_char_buffer(mavlink_message_t* rxmsg,
...
-		//while(status->current_seq != rxmsg->seq)
-		//{
-		//	status->packet_rx_drop_count++;
-		//               status->current_seq++;
-		//}
+		if (status->packet_rx_success_count == 0) { // Initial condition: If no packet has been received so far, drop count is undefined
+			status->packet_rx_drop_count = 0;
+		} else {
+			uint8_t expected_seq = status->current_rx_seq + 1;
+			uint8_t lost_packets = rxmsg->seq - expected_seq;
```
OXINARF: What if you get a late message? The count of lost packets is going to be completely wrong.

kobytg17: That's true actually, was fixing this in the context of a serial connection where misordering is probably less likely, but you're right.

Any better suggestion? I still think that this kind of counting is still better than no counting at all? The same logic is applied in QGroundControl

OXINARF: I don't have a better suggestion, I don't think there is a reliable way to detect packet loss in MAVLink - given that I'm not sure if having an unreliable counter is a good idea (certainly there is a reason why it was disabled), but having the field being useless isn't good either.

kobytg17: I personally think that some kind of indication is still useful, even if not totally robust. I think that in most use-cases it should work well (serial interfaces, TCP, unrouted UDP). The only use case where this theoretically fails is UDP over WAN (routed), and even then misordering is not common.

On the other hand reporting something, can be a good indication of link issues or bad network configuration.

WDYT?

kobytg17: @OXINARF what do you think?

OXINARF: @kobytg17 It's not up to me. I think you have a point but it might mislead people thinking that it is an accurate packet loss detection; as I said, there is a reason why it was disabled.

PR #344, opened 2019-08-16
generator: fix message length table overflow

### LINKED ISSUE #145: MAVLINK_CHECK_MESSAGE_LENGTH have a bug
@julianoes in mavlink_helpers.h, if we define MAVLINK_CHECK_MESSAGE_LENGTH and when migid is greater than 256,mavlink_message_lengths[256] array will overflow.
```
#ifdef MAVLINK_CHECK_MESSAGE_LENGTH
#ifndef MAVLINK_MESSAGE_LENGTH
	static const uint8_t mavlink_message_lengths[256] = MAVLINK_MESSAGE_LENGTHS;
#define MAVLINK_MESSAGE_LENGTH(msgid) mavlink_message_lengths[msgid]
#endif
#endif
``` 

```
	case MAVLINK_PARSE_STATE_GOT_MSGID2:
		rxmsg->msgid |= c<<16;
		mavlink_update_checksum(rxmsg, c);
		status->parse_state = MAVLINK_PARSE_STATE_GOT_MSGID3;
#ifdef MAVLINK_CHECK_MESSAGE_LENGTH
	        if (rxmsg->len != MAVLINK_MESSAGE_LENGTH(rxmsg->msgid))
		{
			_mav_parse_error(status);
			break;
                }
#endif
		break;
```

julianoes:
Since MAVLink 2 we can have message IDs > 256 and we should probably look up the message length in the mavlink_msg_entry dict instead.

Fixes #145.

### DIFF
--- a/generator/C/include_v2.0/mavlink_helpers.h
+++ b/generator/C/include_v2.0/mavlink_helpers.h
@@ -49,6 +49,14 @@ MAVLINK_HELPER mavlink_message_t* mavlink_get_channel_buffer(uint8_t chan)
 }
 #endif // MAVLINK_GET_CHANNEL_BUFFER
 
+/* Enable this option to check the length of each message.
+    This allows invalid messages to be caught much sooner. Use if the transmission
+    medium is prone to missing (or extra) characters (e.g. a radio that fades in
+    and out). Only use if the channel will only contain messages types listed in
+    the headers.
+*/
+//#define MAVLINK_CHECK_MESSAGE_LENGTH
+
 /**
  * @brief Reset the status of a channel.
  */
@@ -566,19 +574,6 @@ MAVLINK_HELPER uint8_t mavlink_frame_char_buffer(mavlink_message_t* rxmsg,
                                                  mavlink_message_t* r_message, 
                                                  mavlink_status_t* r_mavlink_status)
 {
-	/* Enable this option to check the length of each message.
-	   This allows invalid messages to be caught much sooner. Use if the transmission
-	   medium is prone to missing (or extra) characters (e.g. a radio that fades in
-	   and out). Only use if the channel will only contain messages types listed in
-	   the headers.
-	*/
-#ifdef MAVLINK_CHECK_MESSAGE_LENGTH
-#ifndef MAVLINK_MESSAGE_LENGTH
-	static const uint8_t mavlink_message_lengths[256] = MAVLINK_MESSAGE_LENGTHS;
-#define MAVLINK_MESSAGE_LENGTH(msgid) mavlink_message_lengths[msgid]
-#endif
-#endif
-
 	int bufferIndex = 0;
 
 	status->msg_received = MAVLINK_FRAMING_INCOMPLETE;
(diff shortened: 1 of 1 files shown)


generator/C/include_v2.0/mavlink_helpers.h
```diff
@@ -681,11 +676,12 @@ MAVLINK_HELPER uint8_t mavlink_frame_char_buffer(mavlink_message_t* rxmsg,
                         status->parse_state = MAVLINK_PARSE_STATE_GOT_PAYLOAD;
                     }
 #ifdef MAVLINK_CHECK_MESSAGE_LENGTH
-                    if (rxmsg->len != MAVLINK_MESSAGE_LENGTH(rxmsg->msgid))
+                    if (rxmsg->len < mavlink_min_message_length(rxmsg) ||
+                        rxmsg->len > mavlink_max_message_length(rxmsg))
```
magicrub: I have a feeling one of these bounds checks is missing a >= or <=

generator/C/include_v2.0/mavlink_helpers.h
```diff
@@ -708,12 +704,13 @@ MAVLINK_HELPER uint8_t mavlink_frame_char_buffer(mavlink_message_t* rxmsg,
 			status->parse_state = MAVLINK_PARSE_STATE_GOT_PAYLOAD;
 		}
 #ifdef MAVLINK_CHECK_MESSAGE_LENGTH
-	        if (rxmsg->len != MAVLINK_MESSAGE_LENGTH(rxmsg->msgid))
-		{
+        if (rxmsg->len < mavlink_min_message_length(rxmsg) ||
+            rxmsg->len > mavlink_max_message_length(rxmsg))
+        {
```
OXINARF: Not that I like it, but the code around here has tabs as the indenting, you've put spaces here (and below), which makes it look bad.

generator/mavgen_c.py
```diff
@@ -103,10 +103,6 @@ def generate_main_h(directory, xml):
 
 // MESSAGE LENGTHS AND CRCS
 
-#ifndef MAVLINK_MESSAGE_LENGTHS
-#define MAVLINK_MESSAGE_LENGTHS {${message_lengths_array}}
-#endif
```
OXINARF: This needs more work:

- mavlink_helpers from v0.9 and v1.0 weren't changed so they still need this; there are also tests using it
- the comment above saying `// MESSAGE LENGTHS AND CRCS` is now a bit strange when the only definition is then `MAVLINK_MESSAGE_CRCS` (renaming this to `MAVLINK_MESSAGE_ENTRIES` is probably the best idea)
- the `message_lengths_array` variable is unused now, we should remove its generation too

generator/C/include_v2.0/mavlink_helpers.h
julianoes: I agree. There are spaces and tabs in this file, it would be nice to decide on one and fix it.

generator/mavgen_c.py
julianoes: > - mavlink_helpers from v0.9 and v1.0 weren't changed so they still need this; there are also tests using it

I think the implementation is correct for v0.9 and v1.0 because only msgids up to 256 were available.

> - the comment above saying // MESSAGE LENGTHS AND CRCS is now a bit strange when the only definition is then MAVLINK_MESSAGE_CRCS (renaming this to MAVLINK_MESSAGE_ENTRIES is probably the best idea)

Right, thanks.

> - the message_lengths_array variable is unused now, we should remove its generation too

I think I checked that and found that it was still used in other places (such as for v1.0).

generator/C/include_v2.0/mavlink_helpers.h
OXINARF: In ArduPilot we use spaces so I'm sure that's the preference on our side. If PX4 is the same, then I would say that a commit uniforming it to spaces would be well accepted.

generator/mavgen_c.py
OXINARF: > I think the implementation is correct for v0.9 and v1.0 because only msgids up to 256 were available.

Sorry, I might have explained poorly. You are removing the macro definition, which is used in the code - in v0.9, v1.0 and tests (tests seem to actually be covered by a local definition).

> I think I checked that and found that it was still used in other places (such as for v1.0).

I think that's the confusion you are making. The macro is used but you are removing it, which makes the Python variable useless - I can't find it being used anywhere else but here.

> Right, thanks.

I take my advice back. The `MAVLINK_MESSAGE_CRCS` only contains CRCs for MAVLink1 so renaming it isn't the best idea - but you are either going to revert removing `MAVLINK_MESSAGE_LENGTHS ` or change the MAVLink1 version to also include lengths, so the comment is likely to be fixed one way or the other.

julianoes: Thanks, I'll follow up on that.

generator/C/include_v2.0/mavlink_helpers.h
julianoes: I don't know :thinking:.

generator/C/include_v2.0/mavlink_helpers.h
julianoes: So here are some stats for the status quo for all files in generator/C/include_v2.0/:

4 spaces: 1305 matches
tab: 1295 matches

It's a close race! :smile:

generator/mavgen_c.py
julianoes: Restored it in f1ee3622ea59137471b3755cb94449c249cb14c1.

generator/C/include_v2.0/mavlink_helpers.h
julianoes: I'm fine to make a PR to make it consistent but it should probably be a separate PR.

julianoes:
@OXINARF do you need any other changes?

OXINARF:
@julianoes It should be squashed. And my comment regarding the formatting still applies (given we aren't fixing it here, it's better to not make it worse).

julianoes:
Ok, I squashed it to one commit and made the `case` I touched all tabs, and fixed the indenting.

magicrub:
so... spaces wins?

julianoes:
@magicrub I'm very much in favor of spaces but I'm not a maintainer here, so I don't think it should be up to me. :smile:

magicrub:
I'm on team SPACES

julianoes:
Anything else needed here? As said the tab vs. spaces discussion should not interfere with this bugfix in my opinion.

julianoes:
Bump 🥺.

tridge:
I'm not too fussed by the whitespace, merging, thanks!

PR #345, opened 2019-08-16
support instance variable expressions in DF logs

tridge:
this allows for graphs like this:

 graph RFND[0].Dist RFND[1].Dist

The instance field names come from FMTU msgs in DF logs. For mavlink we'll need to add some XML markup for instance vars


tridge:
ping @WickedShell

PR #347, opened 2019-08-19
mavgen_python: correct trailing-zero-byte trimming under Python3

peterbarker:
The struct.pack emits str under Python 2 and bytes under Python3.  bytes
are compared directly against zero but the str form needs the chr().

```
pbarker@bluebottle:~$ cat test-struct.py 
#!/usr/bin/choose-your-own-Python

import struct

x = struct.pack("<BB", 0, 0)
if x[0] == chr(0):
    print("First one is a zero using chr")
else:
    print("First one is NOT a zero using chr")

if x[0] == 0:
    print("First one is a zero using literal zero")
else:
    print("First one is NOT a zero using literal zero")
pbarker@bluebottle:~$ python2 test-struct.py 
First one is a zero using chr
First one is NOT a zero using literal zero
pbarker@bluebottle:~$ python3 test-struct.py 
First one is NOT a zero using chr
First one is a zero using literal zero
pbarker@bluebottle:~$ 
```

### DIFF
--- a/generator/mavgen_python.py
+++ b/generator/mavgen_python.py
@@ -212,7 +212,11 @@ class MAVLink_message(object):
         if WIRE_PROTOCOL_VERSION != '1.0' and not force_mavlink1:
             # in MAVLink2 we can strip trailing zeros off payloads. This allows for simple
             # variable length arrays and smaller packets
-            while plen > 1 and payload[plen-1] == chr(0):
+            nullbyte = chr(0)
+            # in Python2, type("fred') is str but also type("fred")==bytes
+            if str(type(payload)) == "<class 'bytes'>":
+                nullbyte = 0
+            while plen > 1 and payload[plen-1] == nullbyte:
                 plen -= 1
         self._payload = payload[:plen]
         incompat_flags = 0
--- a/tests/test_trim.py
+++ b/tests/test_trim.py
@@ -1,0 +1,34 @@
+#!/usr/bin/python
+
+
+"""
+test for trimming under Python 3
+"""
+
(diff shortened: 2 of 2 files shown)


khancyr:
Could you add this test to the test suite?

peterbarker:
@khancyr Yes, but I'm not going to.  PRs welcome :-P

peterbarker:
@khancyr well, I did go back and add the test and did find a problem.

mavproxy using Python2 worked just fine.  test run under Python2 was not.  I think something is playing with the equivalence of the two types byte and str.

generator/mavgen_python.py
```diff
@@ -212,7 +212,11 @@ def pack(self, mav, crc_extra, payload, force_mavlink1=False):
         if WIRE_PROTOCOL_VERSION != '1.0' and not force_mavlink1:
             # in MAVLink2 we can strip trailing zeros off payloads. This allows for simple
             # variable length arrays and smaller packets
-            while plen > 1 and payload[plen-1] == chr(0):
+            nullbyte = chr(0)
+            # in Python2, type("fred') is str but also type("fred")==bytes
```
OXINARF: Nitpicking: type("fred') -> type("fred")

generator/mavgen_python.py
```diff
@@ -212,7 +212,11 @@ def pack(self, mav, crc_extra, payload, force_mavlink1=False):
...
+            if str(type(payload)) == "<class 'bytes'>":
```
OXINARF: Wouldn't `if type(payload) != str` work?

PR #348, opened 2019-08-20
Fix signature checking under Python3

peterbarker:
Built on top of https://github.com/ArduPilot/pymavlink/pull/347 as it can't be easily tested without that.

Depending on Python version (and, I think, codepath), we may be passed in `bytes` as our payload - and using `chr()` to look into that doesn't work.

Similarly for signature checking - casting to string and taking a subset doesn't work well under Python3, so if we are passed `bytes` in then compare directly.

### DIFF
--- a/generator/mavgen_python.py
+++ b/generator/mavgen_python.py
@@ -681,8 +681,13 @@ class MAVLink(object):
             h = hashlib.new('sha256')
             h.update(self.signing.secret_key)
             h.update(msgbuf[:-6])
-            sig1 = str(h.digest())[:6]
-            sig2 = str(msgbuf)[-6:]
+            if str(type(msgbuf)) == "<class 'bytes'>":
+                # Python 3
+                sig1 = h.digest()[:6]
+                sig2 = msgbuf[-6:]
+            else:
+                sig1 = str(h.digest())[:6]
+                sig2 = str(msgbuf)[-6:]
             if sig1 != sig2:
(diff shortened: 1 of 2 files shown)


generator/mavgen_python.py
```diff
@@ -681,8 +681,13 @@ def check_signature(self, msgbuf, srcSystem, srcComponent):
...
             h.update(msgbuf[:-6])
-            sig1 = str(h.digest())[:6]
-            sig2 = str(msgbuf)[-6:]
+            if str(type(msgbuf)) == "<class 'bytes'>":
+                # Python 3
+                sig1 = h.digest()[:6]
+                sig2 = msgbuf[-6:]
+            else:
+                sig1 = str(h.digest())[:6]
+                sig2 = str(msgbuf)[-6:]
```
OXINARF: I know it's pre-existing, but why the cast to a str? It is already one...

peterbarker:
On Thu, 22 Aug 2019, Francisco Ferreira wrote:

> +                sig1 = str(h.digest())[:6]
> +                sig2 = str(msgbuf)[-6:]
> 
> I know it's pre-existing, but why the cast to a str? It is already one...

I'm trying not to bust existing - presumably working code with these 
changes.  That means erring on the side of minimal changes.

A lot of this code is oozing frustration.

tridge:
tested and confirmed, thanks!

PR #349, opened 2019-08-21
DFReader: add robust parsing

peterbarker:
Throws exceptions in the parsing code, catches them and throws them away if required higher-up.

Will make the tool a little more fault tolerant if your log is slightly corrupt.

### DIFF
--- a/DFReader.py
+++ b/DFReader.py
@@ -53,6 +53,23 @@ FORMAT_TO_STRUCT = {
 def u_ord(c):
 	return ord(c) if sys.version_info.major < 3 else c
 
+class BadFormatCharException(Exception):
+    def __init__(self, char, msg_type):
+        self.char = char
+        self.msg_type = msg_type
+    def __str__(self):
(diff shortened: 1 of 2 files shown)


DFReader.py
```diff
@@ -64,6 +81,9 @@ def __init__(self, type, name, flen, format, columns, oldfmt=None):
         self.unit_ids = None
         self.mult_ids = None
 
+        if len(self.columns) != len(format):
+            raise(FormatColumnCountMismatch(format, columns))
```
OXINARF: This makes me question the two lines below. Can we have a FMT message without format?

DFReader.py
```diff
@@ -497,7 +525,12 @@ def init_clock(self):
 
         have_good_clock = False
         while True:
-            m = self.recv_msg()
+            try:
+                m = self.recv_msg()
+            except BadFormatCharException as e:
+                if not self._robust_parsing:
+                    raise e
+                m = None
```
OXINARF: Why not continue?

PR #354, opened 2019-09-14
mavfft_isb: update to reduce artificial noise in output and make graphs comparable

andyp1per:
- use hanning windows on the input
 - overlap input frames by 50% and use welches method to cover hanning losses
 - convert output to power spectral density normalized by NEBW
 - change scaling and labels for accel and gyros to m^2/s/s/Hz and d^2/s/Hz linerized PSD

See https://holometer.fnal.gov/GH_FFT.pdf for a description of the techniques used

Before:
image

After:
image

### DIFF
--- a/tools/mavfft_isb.py
+++ b/tools/mavfft_isb.py
@@ -88,8 +88,8 @@ def mavfft_fttd(logfile):
         def __str__(self):
             return "%s[%u]" % (self.prefix(), self.instance)
 
-    print("Processing log %s" % filename)
-    mlog = mavutil.mavlink_connection(filename)
+    print("Processing log %s" % logfile)
+    mlog = mavutil.mavlink_connection(logfile)
 
     # see https://holometer.fnal.gov/GH_FFT.pdf for a description of the techniques used here
     things_to_plot = []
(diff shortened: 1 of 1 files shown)


andyp1per:
Actually if we use a log (dB) scale then things stand out a lot better:

image

andyp1per:
Interested in your thoughts @lthall

andyp1per:
After some discussion with @lthall I realized some things were not right. There were some bugs in the way the averages were calculated and I have now fixed those. However, there is still some funkiness with overlapping windows. For now I have stopped this so that we can generate the right output, but my suspicion is that there are actually some bugs in the batch sampler.
Here's a chart with the bugs fixed:
image
Looks much more reasonable. However there is still noise on the output because of the lack of overlap. I will look at that next,

andyp1per:
Ok, having digressed into https://github.com/ArduPilot/ardupilot/pull/12349 I have found out why window overlapping generates strange results and fixed a few bugs as a result. I have now made all of the various options optional and will build a picture below so that you can see the effect.

First linear spectral density on a linear scale with no windowing. This gives almost exactly the same output as before but corrected for some of the bugs that were present:
image

Because the detail is hard to see, I will now switch to a dB scale:
image

The detail is much clearer now and we can subsequently switch to the more normal Power Spectral Density view rather than Linear Spectral Density:
image

Now we can clearly see spectral leakage due to the lack of windowing, most noticeable in the X "oscillation", this is not a feature of the actual signal but due to the way it has been sampled and processed. Windowing cures this:
image

and now we can begin to see the effects of the harmonic notch (the axe-shaped dip)  that were completely invisible before

andyp1per:
@peterbarker any thoughts here?

PR #357, opened 2019-09-27
update gps leap seconds to the correct value

davidbuzz:


davidbuzz:
i'd like to also mention that there are a number of other, better implementation/s that could replace this one ( ie going from gps time to utc time and back again ), and i've run a number of these as tests in this gist....   https://gist.github.com/davidbuzz/df4113f22dd0df5f3787b644ed6c5db1

davidbuzz:
good reference website: http://leapsecond.com/java/gpsclock.htm

tridge:
a more future proof soln would be nice, but until then this is good, thanks!

davidbuzz:
i'm just just leaving a note for future me, or anyone else that comes to this issue...  check the code in my mentioned gist for samples that could be massaged into something better if/when it's needed.

PR #360, opened 2019-10-07
Travis CI: Add more flake8 tests

cclauss:


peterbarker:
A bit of work to do here yet.

Note we do intentionally pollute our namespace in some places so that evaluated expressions don't need to specify a namespace.  e.g. `graph sqrt(RANGEFINDER.time_us)`

peterbarker:
I've rebased this on master, squashed it and removed the upgrade to Python 3.8

peterbarker:
Merged, thanks!

cclauss:
Why no Py38 on focal instead of xenial?  Py39 is final candidate 1 can be tested with "3.9-dev"

peterbarker:
On Fri, 21 Aug 2020, Christian Clauss wrote:

> Why no Py38 on focal instead of xenial? Py39 is final candidate 1 can be tested with "3.9-dev"

I took on @khancyr's comment, and noted that moving Python forward wasn't 
related to the title of the PR....

It's also kind of important to ensure we don't inadvertently break old 
installations.  Perhaps for pymavlink we could afford to use a travis 
matrix?

PR #361, opened 2019-10-17
Fix CSV file missing last line bug.

tecnic08:
Fixed a bug that causes the last line of the csv file output to be dropped.


peterbarker:
@tecnic08 Did you actively hit this bug?

tecnic08:
Yes, I was using mavlogdump.py to convert bin log to csv. It drops out the last line. For most continuous data like ATT, CTUN, it's fine. It is critical with MODE and ERR. It drops out the last mode change and error log.

peterbarker:
Only improvement that could be made I think is to break a method out for printing the message.

But this is much better, thanks!

PR #366, opened 2019-11-07
Partial fix for python3 examples/mavtest.py 

langrind:
Parameter names must be be byte strings, not strings, in calls to param_set_send() and param_set_encode().

'python3 mavtest.py' still fails for an unrelated reason, but I think this PR is a useful step.


peterbarker:
An equivalent patch already went in for this - sorry for missing the PR - thanks anyway.

PR #367, opened 2019-11-11
Fix decoding of signed message on python3.

langrind:
mavgen_python.py has a template for MAVLink.check_signature().
Modified that method's python3 detection to check for
class 'bytearray' as well as class 'bytes'. This fixes python3
decoding of signed messages.

Also changed examples/mavtest.py to pass a bytearray for the
secret key instead of a string. This is required to send
a signed message using python3.

### DIFF
--- a/examples/mavtest.py
+++ b/examples/mavtest.py
@@ -28,7 +28,7 @@ def test_protocol(mavlink, signing=False):
     mav = mavlink.MAVLink(f)
 
     if signing:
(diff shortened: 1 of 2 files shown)


peterbarker:
I tweaked the commit message a bit.

This doesn't actually close #423 - but it does fix the next error that was encountered :-)

Thanks @langrind , merged.

PR #369, opened 2019-11-15
CPP11 message serialisation aliasing fixed

hdino:
This patch fixes the aliasing bug mentioned in #365.

@tridge Please review.

### DIFF
--- a/generator/CPP11/include_v2.0/msgmap.hpp
+++ b/generator/CPP11/include_v2.0/msgmap.hpp
@@ -122,17 +122,18 @@ inline uint64_t to_little_endian_internal<uint64_t>(uint64_t data)
 }
 
 template<typename _T>
-typename UintBuffer<_T>::Type to_little_endian(_T data)
+typename std::enable_if<std::is_floating_point<_T>::value, typename UintBuffer<_T>::Type>::type to_little_endian(_T data)
 {
     typedef typename UintBuffer<_T>::Type ReturnType;
+    ReturnType buf;
+    memcpy(&buf, &data, sizeof(ReturnType));
+    return to_little_endian_internal<ReturnType>(buf);
+}
 
-    if constexpr (std::is_floating_point<_T>::value) {
-        ReturnType buf;
-        memcpy(&buf, &data, sizeof(ReturnType));
-        return to_little_endian_internal<ReturnType>(buf);
-    } else {
-        return to_little_endian_internal<ReturnType>(data);
-    }
+template<typename _T>
+typename std::enable_if<std::is_integral<_T>::value, typename UintBuffer<_T>::Type>::type to_little_endian(_T data)
+{
+    return to_little_endian_internal<typename UintBuffer<_T>::Type>(data);
 }
 
 template<typename _T>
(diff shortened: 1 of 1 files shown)


generator/CPP11/include_v2.0/msgmap.hpp
```diff
@@ -39,7 +39,7 @@ class MsgMap {
 	}
 
 	template<typename _T>
-	void operator<< (const _T data);
+	void operator<< (const _T &data);
```
vooon: What is the reason to pass by ref?

hdino: It avoids a copy (though there won't be any benefit for data types <= pointer size).

vooon: It even can give speed penalty because you have to access pointer.

generator/CPP11/include_v2.0/msgmap.hpp
```diff
@@ -66,70 +63,47 @@ class MsgMap {
...
 // implementation
 
-template<typename _T, typename _Tin>
-void mavlink::MsgMap::msg_swap_memcpy(_T &buf, _Tin data)
-{
-	if (std::is_floating_point<_Tin>::value) {
-		buf = *static_cast<const _T *>(static_cast<const void *>(&data));
-	} else {
-		buf = data;
-	}
```
vooon: Actually you can just put `memcpy()` instead of that deref.

hdino: Indeed, that's true. I changed the design, because previously both `MsgMap::msg_swap_memcpy` and `MsgMap::operator<<` modified `_MAV_PAYLOAD_NON_CONST` (see [here](https://github.com/ArduPilot/pymavlink/blob/master/generator/CPP11/include_v2.0/msgmap.hpp#L97) and [here](https://github.com/ArduPilot/pymavlink/blob/master/generator/CPP11/include_v2.0/msgmap.hpp#L108)).

vooon: `msg_swap_memcpy()` is just a way to avoid copy-paste in `<<`.

hdino:
I created a new version that avoids the switch statement and is easier to read.

vooon:
@hdino yes, that looks better.
Need to do same for `operator>>` (from le to host). But please note that in mavlink v2 message with trailing zero bytes may be trimmed.

hdino:
@vooon I refactored the code for `operator>>`, please have a look.

To keep things consistent, I also adjusted some of the previous changes.

Should we move the newly introduced functions into the `mavlink` namespace?

generator/CPP11/include_v2.0/msgmap.hpp
```diff
@@ -140,94 +155,78 @@ void mavlink::MsgMap::operator<< (const std::array<_T, _Size> &data)
...
-		return;
-	}
+    assert(cmsg);
+    assert(pos + sizeof(_T) <= MAVLINK_MAX_PAYLOAD_LEN);
 
-	switch (sizeof(_T)) {
-	case 1:
-		data = _MAV_PAYLOAD(cmsg)[pos];
-		break;
+    int remaining_non_zero_data = cmsg->len - pos;
```
vooon: Probably better to use ssize_t.

vooon:
Please fix: https://travis-ci.org/ArduPilot/pymavlink/jobs/615053828#L8701

https://en.cppreference.com/w/cpp/language/static_assert - add message.

hdino:
@vooon Thanks for your suggestions! Before merging, please check if the padding with zeroes is still correct, because I implemented it according to the documentation.

generator/CPP11/include_v2.0/msgmap.hpp
```diff
@@ -54,180 +55,185 @@ class MsgMap {
...
-	//std::cout << "M> s: " << sizeof(_T) << " p: " << pos << " d: " << data << std::endl;
-
-	pos += sizeof(_T);
+    assert(cmsg);
+    assert(pos + sizeof(_T) <= MAVLINK_MAX_PAYLOAD_LEN);
+
+    ssize_t remaining_non_zero_data = cmsg->len - pos;
+    typename mavlink::impl::UintBuffer<_T>::Type buf;
+
+    if (static_cast<ssize_t>(sizeof(_T)) <= remaining_non_zero_data) { // field is not truncated
```
hdino: I changed this cast to `ssize_t`

generator/CPP11/include_v2.0/msgmap.hpp
```diff
@@ -54,180 +55,185 @@ class MsgMap {
...
+        memcpy(&buf, &_MAV_PAYLOAD(cmsg)[pos], sizeof(_T));
+    } else if (remaining_non_zero_data <= 0) {
+        // there is no non-zero data left, so just fill with 0
+        buf = 0;
```
hdino: Added this optimisation for the case when there's no data left.

PR #373, opened 2019-12-03
Setup: make mavnative optional by default and update readme

khancyr:
Mavnative don't get support for long years and didn't get safety improvement. I doubt it is really used and only support mavlink1.
I propose to make it install optional by default and allow people that need it to be able to manually install it.
This allow use to keep pymavlink as a pure python package and reduce greatly it installation dependencies : no more need of a compiler.
For example this allow now to use simple docker image for testing that is only 276mb instead of the default python one's that is 1.2Gb ! That save some trees and bees !
On other addition, it allow to embeded sim_vehicle.py on simple docker images simply and keep the size low !


README.md
```diff
@@ -20,19 +20,30 @@ The following instructions assume you are using Python 2 and a Debian-based (lik
 
 ## Dependencies
 
-Pymavlink requires 3 dependencies :
+Pymavlink requires 4 dependencies :
 
     - future : for python 2 and python 3 interoperability (http://python-future.org/)
     - lxml : for checking and parsing xml file (http://lxml.de/installation.html)
     - python-dev : for mavnative
+    - gcc : for mavnative
```
peterbarker: Will any C compiler work?

README.md
```diff
@@ -20,19 +20,30 @@ The following instructions assume you are using Python 2 and a Debian-based (lik
 
 ## Dependencies
 
-Pymavlink requires 3 dependencies :
+Pymavlink requires 4 dependencies :
```
peterbarker:
```suggestion
Pymavlink has several dependencies :
```

README.md
```diff
@@ -54,23 +65,36 @@ Lxml can be installed with a windows installer from here : https://pypi.python.o
...
 It is recommended to install pymavlink from PyPi with pip, that way dependencies should be auto install by pip.
+
 ```bash
 sudo pip2 install -U pymavlink
 ```
+
 The -U parameter allow to update pymavlink version if it is already installed.
 
+#### Mavnative
+By default, pymavlink try to compile and install mavnative that is a C extension for mavlink python parser. Mavnative is quite old already and only support mavlink1.
```
peterbarker:
```suggestion
By default, pymavlink will try to compile and install mavnative which is a C extension for parsing mavlink. Mavnative only supports mavlink1.
```

README.md
```diff
@@ -54,23 +65,36 @@ Lxml can be installed with a windows installer from here : https://pypi.python.o
...
+To skip mavnative installation and reduce dependencies like `gcc` and `python-dev`, you can pass `DISABLE_MAVNATIVE=True` environement variable to installation command :
```
peterbarker:
```suggestion
To skip mavnative installation and reduce dependencies like `gcc` and `python-dev`, you can pass `DISABLE_MAVNATIVE=True` environment variable to the installation command:
```

README.md
```diff
@@ -54,23 +65,36 @@ Lxml can be installed with a windows installer from here : https://pypi.python.o
...
 sudo MDEF=PATH_TO_message_definitions pip2 install . -v
 ```
+
 The -v parameter will output the installation commands on the terminal.
 The MDEF usage is require as pip install is done from /tmp directory, so it is necessary to use MDEF variable to 
 point on the message_definitions directory.
 Use pip should auto install dependencies and allow to keep them up-to-date with pip. 
 
-Or 
+Or :
```
peterbarker:
```suggestion
Or:
```

setup.py
```diff
@@ -75,7 +76,14 @@ def generate_content():
                 sys.exit(1)
 
 extensions = []  # Assume we might be unable to build native code
-if platform.system() != 'Windows':
+# check if we neet to compile mavnative
```
peterbarker:
```suggestion
# check if we need to compile mavnative
```

setup.py
```diff
@@ -85,7 +93,9 @@ def generate_content():
                        ]
                    ) ]
 else:
-    print("Skipping mavnative due to Windows possibly missing a compiler...")
+    print("###################################")
+    print("Skipping mavnative as requested ...")
```
peterbarker:
```suggestion
    print("Skipping mavnative")
```

README.md
khancyr: no, both pip and setup.py will try to call gcc by default

khancyr: at least on linux

khancyr:
Tested on both python2 and 3 with :
```` docker
FROM python:2.7-alpine

RUN apk update && apk add --no-cache --update \
    libxslt \
	libxslt-dev \
    libxml2 \
    libxml2-utils \
    libxml2-dev \
    libc-dev \
    py-lxml \
    git \
    && rm -rf /var/cache/apk/*
# need if apk install lxml
ENV PYTHONPATH=/usr/lib/python3.8/site-packages
RUN pip install -U future lxml

RUN git clone https://github.com/khancyr/pymavlink.git
RUN git clone git://github.com/ArduPilot/mavlink.git
RUN cd pymavlink && git fetch && git checkout -b mavnative origin/mavnative \
    && MDEF=/mavlink/message_definitions DISABLE_MAVNATIVE=True pip install .
````
Change FROM python:2.7-alpine to FROM python:3.8-alpine or 3.7, pass DISABLE_MAVNATIVE=True to False, or change pip install to python setup.py install ! 
That is very fast

peterbarker:
On Wed, 22 Jan 2020, Pierre Kancir wrote:

> no, both pip and setup.py will try to call gcc by default

A bit of reading shows that some people have the opposite problem - clang 
being called instead.  I think we can just say, "a C compiler"?

khancyr:
done ! 
That is interesting as I tried it with clang and it was also failing with explicit call to gcc !

peterbarker:
Merged, thanks!

Issue #383, opened 2020-02-11
Typescript message definitions do not handle arrays


nthorn552:
We have been working to get a NodeJS/Typescript server to communicate with PX4 through MAVLink, but have run into an issue with some messages.  We connected the issue to messages that have field types that are arrays.
Currently, the typescript generator creates the `_message_fields` property in the format `[string, string, boolean][]`, where the values are name, type, and isMavlinkV2.  For messages like LOG_DATA, data is a field with type `uint8_t[90]` (https://mavlink.io/en/messages/common.html#LOG_DATA), but the TypeScript message definition that is created shows `["data", "uint8_t", false]`, leaving a parser with no information it should be expecting the next 90 values to be part of that array. 

What would you think about including the number of values in the array in this `_message_fields` property?  My thought was something like `[string, string, boolean, number`], where the new number value would be number of items.  The default, for most fields, could be "1", or for cases like LOG_DATA, it could be `["data", "uint8_t", false, 90].  I was going to try making this change locally to try it out, but thought I would post it here as well in case anyone has tips or guidance for best ways to go about this, or if it's even worth addressing.

If there is already a way of handling these types of messages with TypeScript, please let me know!  I am currently working with this parser (https://github.com/ifrunistuttgart/node-mavlink).

PR #384, opened 2020-02-12
Add array field consideration to TypeScript message definition script

nthorn552:
This change was in response to an issue [(#383)](https://github.com/ArduPilot/pymavlink/issues/383) I was having trying to build a MAVLink service in TypeScript, where parsers are unable to get information about a field that uses an array.  The change I'm proposing would be to include the array length as an additional value in the `_message_field` array.  The number would be `0` when the field is not an array, and would be an integer representing the array length when the field is an array.

### DIFF
--- a/generator/mavgen_typescript.py
+++ b/generator/mavgen_typescript.py
@@ -81,6 +81,8 @@ def generate_classes(dir, registry, msgs, xml):
                 for field in m.fields:
                     if field.enum:
                         f.write("\tpublic {}!: {};\n".format(field.name, camelcase(field.enum)))
+                    elif field.array_length > 0:
+                        f.write("\tpublic {}!: {}[]; //size: {}\n".format(field.name, ts_types[field.type], field.array_length))
                     else:
                         f.write("\tpublic {}!: {};\n".format(field.name, ts_types[field.type]))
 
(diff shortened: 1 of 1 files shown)


rummanwaqar:
Hey @nthorn552 thanks for doing this. Any idea what needs to be done to push this in? 
Did you also by any chance push the chances from node-mavlink side? A fork possibly?

generator/mavgen_typescript.py
```diff
@@ -81,6 +81,8 @@ def generate_classes(dir, registry, msgs, xml):
                 for field in m.fields:
                     if field.enum:
                         f.write("\tpublic {}!: {};\n".format(field.name, camelcase(field.enum)))
+                    elif field.array_length > 0:
+                        f.write("\tpublic {}!: {}[]; //size: {}\n".format(field.name, ts_types[field.type], field.array_length))
```
rummanwaqar, line 85: I think we might have a bug here for strings:

This line will convert 
```<field type="char[249]" name="message">message payload</field>``` 
to 
```public message!: string[]; //size: 249```

We need this to be just string and not array of string

davidbuzz:
I suspect this PR can either be closed, or needs a rework after https://github.com/ArduPilot/pymavlink/pull/431 ?

PR #385, opened 2020-02-20
C/Cpp Generator support relative include paths in xml definitions

len0rd:
Per the [MAVLink XML schema guide](https://mavlink.io/en/guide/xml_schema.html) relative include paths in message dialect definitions should be supported. ie:

```xml
<include>../mavlink/message_definitions/v1.0/common.xml</include>
```
is a valid include in a message definition file.

However, the C/Cpp generators currently take the whole path from an include tag and place it in an include as shown:
```c
#include "../../mavlink/message_definitions/v1.0/common/../mavlink/message_definitions/v1.0/common.h"
```

This PR modifies mavgen_c and mavgen_cpp11 behavior to agree with the schema guide and allow relative includes. It does this by stripping any path components off the base include string before adding it to the include list. This in turn generates the expected include of :

```c
#include "../common/common.h"
```

### DIFF
--- a/generator/mavgen_cpp11.py
+++ b/generator/mavgen_cpp11.py
@@ -330,7 +330,7 @@ def generate_one(basename, xml):
     # work out the included headers
     xml.include_list = []
     for i in xml.include:
-        base = i[:-4]
+        base = os.path.basename(i)[:-4]
         xml.include_list.append(mav_include(base))
 
     # and message metadata array


tridge:
I've merged the other PR from @olliw42 that used the correct os.path.basename(). We should now do the same for cpp11

tridge:
I've rebased and fixed to use os.path.basename(), can merge when passes CI

PR #391, opened 2020-03-08
Fix compile warnings in mavlink_conversions.h

LorenzMeier:
Modern compilers might throw a warning due to the implicit increase in precision. The explicit cast (in C syntax to not force C++ for MAVLink library adopters) tells the compiler that this is intentional. We store as an intermediate double to ensure that compilers use full double precision for the operation.


peterbarker:
Merged, thanks!

PR #392, opened 2020-03-12
Adding bitfield option to enum, to indicate multi bits enum

BazookaJoe1900:
as been discussed on [adding bitmask="true" for enums #1316
](https://github.com/mavlink/mavlink/pull/1316)

### DIFF
--- a/generator/mavschema.xsd
+++ b/generator/mavschema.xsd
@@ -238,6 +238,7 @@
             <xs:element ref="entry" maxOccurs="unbounded"/>
         </xs:sequence>
         <xs:attribute ref="name" use="required"/>
+        <xs:attribute ref="bitmask" type="xs:boolean" default="false"/>
     </xs:complexType>
 </xs:element>
 


peterbarker:
How does this interact with the `display="bitmask"` we have on fields?

BazookaJoe1900:
@peterbarker 
as @hamishwillee sayed [here](https://github.com/mavlink/mavlink/pull/1316#issuecomment-598431390), we can remove the `display="bitmask"` on later stage

hamishwillee:
@peterbarker As linked by @BazookaJoe1900 , the intention is that this will eventually _replace_ the `display="bitmap"` on fields. I wanted to do it in two parts, because while I am sure this is better, I am not sure mission planner does not depend on `display="bitmap"`  (@meee1 - could you comment?).

I know this is not yet used in QGC [see comment from Don on the linked post]

The logic for this is that if something is a bitmap/set of flags then it should be that in all fields it is used in. So this is more efficient and descriptive.

peterbarker:
On Sun, 15 Mar 2020, Hamish Willee wrote:

> @peterbarker As linked by @BazookaJoe1900 , the intention is that this will eventually replace the display="bitmap" on fields. I wanted to do it in two
> parts, because while I am sure this is better, I am not sure mission planner does not depend on display="bitmap" (@meee1 - could you comment?).

There are tools around that are reliant, so I don't think it can go away 
for quite some time.  The firmware on the SkyViper, for example.

There's also ./mavlink_bitmask_decoder.py within the pymavlink directory.

> I know this is not yet used in QGC [see comment from Don on the linked post]
> 
> The logic for this is that if something is a bitmap/set of flags then it should be that in all fields it is used in. So this is more efficient and
> descriptive.

I agree, which kind of raises the question as to why it wasn't done that 
way in the first place.  Perhaps @amilcarlucas knows?

I do have a recollection that there are some *weird* things going on with 
kinda-sorta enumerations.  I can't immediately think of what that was, 
however :-/

peterbarker:
Should this new attribute imply display="bitmask" where otherwise unspecified on a field?

hamishwillee:
> Should this new attribute imply display="bitmask" where otherwise unspecified on a field?

Yes, or at least I can't think of a case where you would apply it and it would be incorrect for the field.

PR #393, opened 2020-03-17
DFReader.py: continue parsing if a bad header is received

peterbarker:
Allows us to continue in the face of mildly corrupt logs


peterbarker:
Note that the progress callback stuff won't be done.

PR #395, opened 2020-03-20
mavmission: trim mission length to number of items in last item found

peterbarker:


peterbarker:
Thanks to @WickedShell for fixing the bugs.

WickedShell:
This helps you not get odd mission files when someone has trimmed a mission down to be shorter then it was. I was looking at a mission dumped by this and was very confused because the last half of it was extremely different/remote from the first half, and it picked up in the middle of a mapping mission apparently (among other odd things). What had happened is that the mission file had picked up the first (larger) mission, but a smaller mission was later upload which only had about half the number of items in it. The later ones were no longer available on the vehicle, so dumping them out as part of mavmission.py gave an incorrect view of what was happening.

PR #396, opened 2020-03-20
DFReader: clear message dictionary rather than replacing it

peterbarker:
Some places take a reference to the old array, holding onto old data
when they shouldn't have it.  They can copy.copy it if they really want
it.

Also correct parsing of instance-number-containing fields


peterbarker:
tab completion is broken in MAVExplorer after this patch; presumably it wants things out of that messages hash, even after the log those messages were derived from has been rewound.  tab-completion may need to take a copy of the messages hash.

PR #406, opened 2020-04-16
Add debug option keyword to evaluate_expression

peterbarker:
Will allow callers like MAVProxy to show failures in evaluating
expressions.


peterbarker:
Possible bug where v is not defined in this function?

peterbarker:
Push example PR into mavproxy for this.

PR #408, opened 2020-04-28
mavgen.py: print stringification of Exception causing us to not validate

peterbarker:


BazookaJoe1900:
similar change done here: https://github.com/ArduPilot/pymavlink/commit/f1ac04b004de00178ea8cef1640df1c2cac92d8b

you can rephrase the text, of course

PR #411, opened 2020-05-02
mavutil: use command_long for change modes for ardupilot

peterbarker:
This may allow us to sunset some really nasty code in ArduPilot which
acks a message which is not a command_long or command_int.

### DIFF
--- a/mavutil.py
+++ b/mavutil.py
@@ -643,9 +643,17 @@ class mavfile(object):
                 return
             mode = mode_map[mode]
         # set mode by integer mode number for ArduPilot
-        self.mav.set_mode_send(self.target_system,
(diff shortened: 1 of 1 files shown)


peterbarker:
From https://github.com/ardupilot/ardupilot/blob/master/libraries/GCS_MAVLink/GCS_Common.cpp#L2073

```
    // send ACK or NAK.  Note that this is extraodinarily improper -
    // we are sending a command-ack for a message which is not a
    // command.  The command we are acking (ID=11) doesn't actually
    // exist, but if it did we'd probably be acking something
    // completely unrelated to setting modes.
    if (HAVE_PAYLOAD_SPACE(chan, COMMAND_ACK)) {
        mavlink_msg_command_ack_send(chan, MAVLINK_MSG_ID_SET_MODE, result);
    }
```

PR #413, opened 2020-05-06
allow force disarming

dwiel:


peterbarker:
Personally I'm hesitant to merge this.  Firstly, because these methods themselves should really be deprecated - by abstracting them we're hiding stuff which shouldn't be hidden.  For example, after this method returns the vehicle isn't disarmed - if the message is lost it may never attempt to disarm.  The caller *could* be monitoring for acks (and must monitor HEARTBEAT) - the abstraction hides that fact.

Now my real problem.... building in the magic.  We put that magic value in there to make it slightly harder for a programmer to accidentally crash a vehicle - to think about whether forcing the disarm is appropriate.  By hiding it behind a "force" flag we are going to lose part of it.

... again, I'm surprised people are using these methods :-)

dwiel:
Fair enough. I have some cases where the disarm logic is incorrectly deciding not to disarm even when I definitely know I want to, and I use this call to override that. I understand if abstracting over the magic defeats the purpose of forcing a user into the documentation before actually calling it though.

peterbarker:
On Wed, 6 May 2020, Zach Dwiel wrote:

> Fair enough. I have some cases where the disarm logic is incorrectly deciding not to disarm even when I definitely know I want to, and I use this call to
> override that. I understand if abstracting over the magic defeats the purpose of forcing a user into the documentation before actually calling it though.

Are you content to essentially to the command_long call yourself?

I'm not blocking this - this is just an ancient, trivial routine which I'm 
not sure really belongs in mavutil at all :-)

dwiel:
> Firstly, because these methods themselves should really be deprecated - by abstracting them we're hiding stuff which shouldn't be hidden.

I'm curious which methods do you think should be deprecated and what is hiding that shouldn't be? Just the magic number or something else?

dwiel:
Ahh that makes sense. I just followed the call stack from dronekit to here so made the patch here. I have no preference on where the transition to command_long call happens.

peterbarker:
On Wed, 6 May 2020, Zach Dwiel wrote:

> Ahh that makes sense. I just followed the call stack from dronekit to here so made the patch here. I have no preference on where the transition to
> command_long call happens.

dronekit calls this?!  Darn, so it does.  I'd merge a patch in dronekit to 
add the force option, I think.  So long as the force option also took a 
magic number rather than a boolean ;-)

peterbarker:
Closing in preference to dronekit-python getting a patch to stop using this routine.  Thanks anyway!

PR #414, opened 2020-05-11
generator: removed old debug output

tridge:
this verbosity just made it harder to spot real issues


magicrub:
oh good, ditch that spam!

PR #417, opened 2020-05-24
tools: add mavfft_pid.py to allow FFT analysis for pids in log files

andyp1per:
Example output

image

FYI @lthall

### DIFF
--- a/tools/mavfft_pid.py
+++ b/tools/mavfft_pid.py
@@ -1,0 +1,92 @@
+#!/usr/bin/env python
+
+'''
+fit estimate of PID oscillations
+'''
+from __future__ import print_function
+
+import numpy
+import pylab
+
+from argparse import ArgumentParser
+parser = ArgumentParser(description=__doc__)
+parser.add_argument("--condition", default=None, help="select packets by condition")
+parser.add_argument("--sample-rate", dest='sample_rate', type=int, default=400, help="sample rate of PID values")
+parser.add_argument("logs", metavar="LOG", nargs="+")
+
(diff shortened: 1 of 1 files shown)


tools/mavfft_pid.py
```diff
@@ -0,0 +1,92 @@
...
+        for ax in ['P', 'I', 'D']:
+            data[gyr+'.'+ax] = []
+
+    # now gather all the data
+    while True:
+        m = mlog.recv_match(condition=args.condition)
+        if m is None:
+            break
+        type = m.get_type()
+        if type.startswith("PIDR"):
```
peterbarker, line 42: Why not direct equivalence?  Could be rather faster if there's a checksum associated with the strings, for example.

tools/mavfft_pid.py
```diff
@@ -0,0 +1,92 @@
...
+            data[type+'.P'].append(m.P)
+            data[type+'.I'].append(m.I)
+            data[type+'.D'].append(m.D)
+        if type.startswith("PIDP"):
```
peterbarker, line 46: `elif` here will make things faster.

tools/mavfft_pid.py
andyp1per: My python is like my french - limited and broken. I tend to stick with things I've copied and pasted. Anyway fixed now.

tools/mavfft_pid.py
andyp1per: Fixed

peterbarker:
Merged, thanks!

PR #418, opened 2020-05-27
Update README.md

spigelli:
Cleaned up some of the links and added a bit of detail on MDEF explanation for developer installations. Added a reference to the pymavlink section of the ArduSub book.


peterbarker:
Merged, thanks!

Those instructions are not good IMO; most users should not be using `sudo` to install pymavlink IMO.

PR #427, opened 2020-06-26
CI: Update build environment

stephendade:
The version of Ubuntu used in TravisCI needs to be updated in order to use a modern version of nodejs.

Currently, the nodejs tests are failing with the same error over as https://github.com/mavlink/mavlink/issues/1404.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ python:
   - "3.6"
   - "3.7"
 before_install:
+  - nvm install 12.18.1 # need recent version of nodejs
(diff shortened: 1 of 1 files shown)


stephendade:
So, this fixes the nodejs issues but breaks the c tests in ``test_generator.sh``. Will need to look further.

stephendade:
Ok, fixed up by specifying the current nodejs LTS

peterbarker:
Thanks @stephendade !

PR #428, opened 2020-06-27
mavgen: Javascript add parser array support

jmachuca77:
The javascript generator was not able to handle messages that contain arrays. This fixes the parsing of the incoming messages. The sending side is still not working.


peterbarker:
I've rebased this to ensure it passes CI.

peterbarker:
Merged, thanks!

PR #430, opened 2020-07-06
Make sure init_arrays() also observes 528 garbage skip

andyp1per:
Give file position when a bad header or message is encountered
Check for zero length file

### DIFF
--- a/DFReader.py
+++ b/DFReader.py
@@ -758,7 +758,11 @@ class DFReader_binary(DFReader):
         while ofs+3 < self.data_len:
             hdr = self.data_map[ofs:ofs+3]
             if hdr[0] != HEAD1 or hdr[1] != HEAD2:
-                print("bad header 0x%02x 0x%02x" % (u_ord(hdr[0]), u_ord(hdr[1])), file=sys.stderr)
+                # avoid end of file garbage, 528 bytes has been use consistently throughout this implementation
+                # but it needs to be at least 249 bytes which is the block based logging page size (256) less a 6 byte header and
+                # one byte of data. Block based logs are sized in pages which means they can have up to 249 bytes of trailing space.
+                if self.data_len - ofs >= 528 or self.data_len < 528:
+                    print("bad header 0x%02x 0x%02x at %d" % (u_ord(hdr[0]), u_ord(hdr[1]), ofs), file=sys.stderr)
                 ofs += 1
                 continue
             mtype = u_ord(hdr[2])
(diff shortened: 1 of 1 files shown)


DFReader.py
```diff
@@ -755,19 +755,28 @@ def init_arrays(self, progress_callback=None):
...
         while ofs+3 < self.data_len:
             hdr = self.data_map[ofs:ofs+3]
             if hdr[0] != HEAD1 or hdr[1] != HEAD2:
-                print("bad header 0x%02x 0x%02x" % (u_ord(hdr[0]), u_ord(hdr[1])), file=sys.stderr)
-                ofs += 1
-                continue
+                # avoid end of file garbage, block based logs can have up to 256 bytes of zeros
+                if self.data_len - ofs >= 528 or self.data_len < 528:
+                    print("bad header 0x%02x 0x%02x at %d" % (u_ord(hdr[0]), u_ord(hdr[1]), ofs), file=sys.stderr)
+                    ofs += 1
```
peterbarker, line 767: This seems to be changing logic, not just squelching warnings.

andyp1per: Fixed

DFReader.py
```diff
@@ -755,19 +755,26 @@ def init_arrays(self, progress_callback=None):
         HEAD2 = self.HEAD2
         lengths = [-1] * 256
 
+        if self.data_len < 3:
```
peterbarker, line 758: This doesn't look right.  Why would we do this?

mavlogdump of an empty file should produce zero output, I think.  Of course, it currently falls over with an mmap error, but that's a separate issue.

andyp1per: Hmmn, I was trying to fix another problem I see. But you are right I will take it out

DFReader.py
```diff
@@ -758,16 +758,20 @@ def init_arrays(self, progress_callback=None):
         while ofs+3 < self.data_len:
             hdr = self.data_map[ofs:ofs+3]
             if hdr[0] != HEAD1 or hdr[1] != HEAD2:
-                print("bad header 0x%02x 0x%02x" % (u_ord(hdr[0]), u_ord(hdr[1])), file=sys.stderr)
+                # avoid end of file garbage, block based logs can have up to 256 bytes of zeros
```
peterbarker, line 761: The 256 and the 528 are at odds with one-another here.  There might be a good reason for, but if it's making me curious it probably needs a comment at least :-)

andyp1per: Comment expanded

peterbarker:
Merged, thanks!

PR #431, opened 2020-07-08
js mavlink parser fixes and test suite 

davidbuzz:
node/javascript generator now handles int64 , broken / in-complete jspack (our own included version now), arrays (without pre-flattening), optional verbosity increases available, packet variables like 'id' and 'name' , 'header' no longer being blindly overwritten with user ones, lots of additional tests in a new and heavily expanded node/js test suite.

 - Test suite now also uses the ardupilot dialect not just the common one to be able to handle a larger set of packet types.
 - fix javascript generator to support  array-based fields and allow the use of 'id' as a user-supplied column name by moving the internal use of that to _id
 - also includes some minimal examples for python and node using different parsers
 - adding a pretty thorough test suite for js bindings in three stages. 
 - we can now auto-build a mocha test suite from the output of the C bindings and some python/node code with make_tests.py, resulting in test/made_tests.js that 'mocha' can run.
 - added local copy of 'long' js library as its needed to properly parse all 64bits, as node numbers are not long enough, and fixes to long library
 - added 'mocha' and 'should' as proper dependancies in package.json
 - added local 'jspack' and 'long' as proper dependancies in package.json, installed from 'local_modules'
 - jspack changes includes more debug everywhere and de-obfusticated a lot of jspack.js. DEBUG flag is disabled by default.
 - we now make a mavlink.tests.js output from mavgen_javascript.py, setup so they can be called individually when require-ed as a module, or together as a 'main' script .  this script uses the same 'test data' values as the other bindings, and is inspired by the C/python bindings.
 - make_tests.py builds on top of the new mavlink.tests.js module properly, and includes comments about how to expand tests where different data is needed.  It builds end-to-end tests to the byte level, comparing a raw 'buffer' from the C inpl to the node-assembled packet/s.
 - make_tests.py test maker is also able to identify packets with signing, and skip over those tests, which aren't properly handled yet.
Notes/fixes included:
 - binary encoding when using the node .toString('binary') is essential to getting clean handling of bits from 128-255,  the js generator to uses this binary encoding on 'strings' correctly.
 - local jspack adds the ability to pack and unpack from a source-array , thx to two commits by AndreasAntener, which essentially gets rid of the need to pre-flatten the input and fixes a bunch of side-effects that flattening has cascaded onto Long handling.
 - jspack has a few manually curated tests added into the test/int64.js file for a few corner cases buzz was chasing, good to leave here.
 - long.js respects the sign of zero value/s during a multiply, as per https://github.com/dcodeIO/long.js/pull/76/files - not 100% sure this is needed, but its a good change anyway.
 -this ensures the generator output uses Long.fromString() instead of Long.fromNumber() as the latter seems to have issues with some specific big numbers and fails tests-  known bug in jspack issues list, but worked-around.
 - we explicitly use a  IntXXXArray to 'cast' when turning test data into SIGNED content, as test data may be outside the range limits of jspack, and causes tests to fail without this fix. (C test data frequently uses the unsigned equivalent of a signed value, and so we now support this with casting) 
 - added test_gen_js.sh testing shell script with everything we have now,, runs all tests in sequence.
 - _header, _msgbuf, _name, _id, _data, _reason variable renames as described in first line.
 - report stomping on attributes that are non-numeric, as its a common use-case for the parser breaking.
 - npm version 5 and above support using a local copy of node modules, so we now rely on that.


stephendade:
I've now a look at this:

./generator/C/test/posix:
making this gives the following error message:
``testmav.c:17:10: fatal error: version.h: No such file or directory``

./examples/min.js:
Doesn't appear to work: Bad prefix error

.test_gen_js.sh:
How is this different from "npm test"?
Has hardcoded paths from Buzz's laptop
Got the following error message on running:
```
test creating and packing:statustext
test creating and packing:debug
+ mavgen.py --lang C ../message_definitions/v1.0/ardupilotmega.xml -o generator/C/include_v1.0 --wire-protocol=1.0
  File "/home/stephen/.local/bin/mavgen.py", line 4, in <module>
    __import__('pkg_resources').run_script('pymavlink==2.4.9', 'mavgen.py')
  File "/home/stephen/.local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 666, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/home/stephen/.local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 1444, in run_script
    .format(**locals()),
pkg_resources.ResolutionError: Script 'scripts/mavgen.py' not found in metadata at '/home/stephen/Documents/UAV_code/mavlink/pymavlink/pymavlink.egg-info'
```
./generator/javascript/package.lock.js
I don't think this file is usually committed.

./generator/javascript/test/made_tests.js:
Is this the generated file to test all packet types?
Is it autogenerated on "npm test"?, if not, how is it generated?
Is this mavlink2 only?

./generator/javascript/test/make_tests.py
Unused import os and json

Need to add documentation about how to add the modified node modules. I assume the parser won't work without those?

davidbuzz:
@stephendade , thanks for the feedback, i'll massage it some more to to to answer all your concerns/questions and get back to you.

davidbuzz:
> I've now a look at this:
> 
> ./generator/C/test/posix:
> making this gives the following error message:
> `testmav.c:17:10: fatal error: version.h: No such file or directory`
i will look into this.
> 
> ./examples/min.js:
> Doesn't appear to work: Bad prefix error
i will look into this.
> 
> .test_gen_js.sh:
> How is this different from "npm test"?
There's three different stages of testing-related code now, and 'npm test' is only really the last/biggest and it relies on the earlier states to work...
stage 1: the same mavgen_javascript.py we know and love to make the runtime bindings (eg mavlink.js) also builds and supplies enough test data to make mavlink.tests.js along-side it, which contains fixed test data and fixed mav data and tests the static "build" of each packet type with static data as if for 'sending'.
stage2: a new test script examples/testparser.js demonstrates the other side of the test/s, doing a "recieve" of a large binary mavlink data from the C implementation test tools, wrapped with appropriate shell scripts, this recieves them and prints the successful parsing of the data onto the console.
stage 3: 'npm test' now also executes an auto-generated part , where the generator for the 'auto' part is "make_tests.py" and the output of this is called "made_tests.js", and this tool relies on the  .tests.js libraries generated in stage1  to run, and also wraps the same C bindings that are used in stage2 ( ie it brings both the 'send' and 'recieve' parts together, including the available static test data from stage 1, and the dynamic byte-parsing in stage 2 ). 

> Has hardcoded paths from Buzz's laptop
sorry, i'll fix them in next push, they slipped through here. 

> Got the following error message on running:
> 
> ```
> test creating and packing:statustext
> test creating and packing:debug
> + mavgen.py --lang C ../message_definitions/v1.0/ardupilotmega.xml -o generator/C/include_v1.0 --wire-protocol=1.0
> Traceback (most recent call last):
>   File "/home/stephen/.local/bin/mavgen.py", line 4, in <module>
>     __import__('pkg_resources').run_script('pymavlink==2.4.9', 'mavgen.py')
>   File "/home/stephen/.local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 666, in run_script
>     self.require(requires)[0].run_script(script_name, ns)
>   File "/home/stephen/.local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 1444, in run_script
>     .format(**locals()),
> pkg_resources.ResolutionError: Script 'scripts/mavgen.py' not found in metadata at '/home/stephen/Documents/UAV_code/mavlink/pymavlink/pymavlink.egg-info'
> ```
weird, i will look into this further.

> ./generator/javascript/package.lock.js
> I don't think this file is usually committed.
 Actually, that's innaccurate, this file is specifically designed to be committed.   https://github.com/npm/npm/blob/v5.0.0/doc/files/package-lock.json.md


> ./generator/javascript/test/made_tests.js:
> Is this the generated file to test all packet types?
yes, where 'all' is all the packets that are output by the C testsuite implementation, so if the C test are  incomplete, so is this, but its much much better than we had before.

> Is it autogenerated on "npm test"?, if not, how is it generated?
> Is this mavlink2 only?
The intention with this branch is that it be both mavlink1 and mavlink2, and at least one of the the script/s is already supporting both versions, but right now there are a few .js scripts that assume mavlink2.  , its a wip after all.

> ./generator/javascript/test/make_tests.py
> Unused import os and json
I will fix these when i push next, no problem.

> Need to add documentation about how to add the modified node modules. I assume the parser won't work without those?
The updated package.json auto installs them correctly as a symlink from local_modules when you run 'npm install' (or 'make') in the generator/javascript folder, so there is no special action needed that isn't already done.

davidbuzz:
> ./generator/C/test/posix:
> making this gives the following error message:
> `testmav.c:17:10: fatal error: version.h: No such file or directory`
that error appears when u type 'make' on its own and is a pre-existing bug, but when u type 'make clean testmav1.0_ardupilotmega testmav2.0_ardupilotmega testmav1.0_common testmav2.0_common' as a single command, it should build what we need just fine.    the test_gen_js.sh script does this correctly.

davidbuzz:
> ./examples/min.js:
> Doesn't appear to work: Bad prefix error
min.js is now fixed

davidbuzz:
i was unable to re-create the eror u reported.. "pkg_resources.ResolutionError: Script 'scripts/mavgen.py' not found in metadata at '/home/stephen/Documents/UAV_code/mavlink/pymavlink/pymavlink.egg-info'"       - any chance u could re-test the pr as i've pushed a fix for everything else..? - thx @stephendade  ( or anyone else )

jmachuca77:
That last one is an installation issue with pymavlink, he might need to clean up his installation

davidbuzz:
... bonus... it now tests against 2 versions and 2 base/s ( ie ardupilotmega/common and 1.0/2.0 
.... and 'npm test' now autobuilds the tests before executing them

davidbuzz:
just dropping a note here that @stephendade has done an initial review on it and was able to execute the entire test suite properly on his computer, so we are close.   :-)

generator/C/test/posix/Makefile
```diff
@@ -1,4 +1,4 @@
-CFLAGS = -g -Wall -Werror -O0
+CFLAGS = -g -Wall -Werror -O0 -Wno-address-of-packed-member
```
peterbarker, line 1: Please add a comment here to explain why it's OK to add this compiler directive in the case of mavlink bindings.

generator/C/test/posix/testmav.c
```diff
@@ -120,6 +120,9 @@ static void print_message(mavlink_message_t *msg)
 	}
 	const mavlink_field_info_t *f = m->fields;
 	unsigned i;
+	printf("sysid:%d ", msg->sysid);
```
peterbarker, line 123: Is this stray debug?

generator/mavgen_javascript.py
```diff
@@ -234,35 +245,44 @@ def field_descriptions(fields):
         outf.write("""
 
     this.format = '%s';
-    this.id = %s.MAVLINK_MSG_ID_%s;
```
peterbarker, line 237: This will break people's code pretty horribly.

Likewise for _name.

Probably better to leave this in place, but with a comment explaining why people should use `_id` and `_name`.

I can't see a technical way to deprecate use of these attributes with name collisions.

generator/C/test/posix/testmav.c
davidbuzz: Its a deliberate addition to the C tests/binding/s that causes them to emit enough human-readable info about each of the 'test' packets so that a human can cross-reference what the C bindings think the packet is with what the testparser.js thinks the bytes mean. future todo - at some point it would be nice to  extend testparser.js to also consume this data and cross-check the internal/s of the packets as well, not just 'did the hex parse into a known packet type' that it does now.

generator/mavgen_javascript.py
davidbuzz: Yes, i'm totally worried about the attribute collisions, as we have a bunch of them without my changes...  the list below are all places where the old javascript code collides between xml field names and internal parser goodness, which is very bad, and should absolutely not happen.   right now, i've had to break backward compatibility in this commit because
(example/s) 
 - with a message of type BATTERY_STATUS -> msg.id  refers to field 'Battery ID'  not MsgID which was also msg.id before this pr
 - with a messages of type PLAY_TUNE_V2  ->  msg.format refers to bitmask of supported tune formats enum="TUNE_FORMAT" , not the parser's FMT string, which was also msg.format before this pr
 and this happens all over the place.

The other alternative/s dont look any better, for example if we moved all the xml attributes to something like msg.field.id and msg.field.format and msg.field.name then that still breaks backward compatibility because there's gonna be a buncg of code that uses (eg) msg.name and msg.id  , and there's not way to magically fix this.   

The solution ive used here follows the python parser's convention/s as closely as reasonably possible... python tends to prefix internal attributes with an _  ( like _msgbuf ), so thats also one of the reasons i chose this way

current problematic field/s i've resolved in this pr:
name => _name
id => _id
format => _format
payload => _payload 
header => _header
data => _data

I have thought about making a dynamic gettter/setter for the attribute/s of a packet , so that it would maintain compatibility for as many packets as possible by looking for an attribute *first* in the xml attributes  and then in the header/s, essentially meaning that we've encoded a 'preference' that (example) "msg.id"  should mean msg._id most of the time but in packet/s that have an 'id' field in their xml, it would instead mean msg.fields.id  , but although that's a pretty smart backward/s compatability shim that could work for most things most of the time, it's utterly non-compliant. 

Sooo... the only answer i have is to make a one-off breaking change to the parser, that actually fixes it properly, aligns it more with the way python does it, and importantly, gets the parser passing all its tests, which it never had before.

generator/C/test/posix/Makefile
davidbuzz: without this, i get many-many (166) build failures from the mavlink library.... and with it the problem goes away....
example error:
../../include_v2.0/ardupilotmega/../common/testsuite.h: In function ‘mavlink_test_distance_sensor’:
../../include_v2.0/ardupilotmega/../common/testsuite.h:6417:292: error: taking address of packed member of ‘struct __mavlink_distance_sensor_t’ may result in an unaligned pointer value [-Werror=address-of-packed-member]

number of errors without the fix:
buzz@buzz:~/GCS/mavlink/pymavlink/generator/C/test/posix$ make clean testmav1.0_ardupilotmega testmav2.0_ardupilotmega testmav1.0_common testmav2.0_common 2>&1 | grep 'error:' | wc
    166    4770   60806

number of errors with the fix:
buzz@buzz:~/GCS/mavlink/pymavlink/generator/C/test/posix$ make clean testmav1.0_ardupilotmega testmav2.0_ardupilotmega testmav1.0_common testmav2.0_common 2>&1 | grep 'error:' | wc
      0       0       0

amilcarlucas:
113K new lines of code!!! This is a big one!

davidbuzz:
> 113K new lines of code!!! This is a big one!

u should ignore everything in the generator/javascript/implementations which is just there during this WIP so that people can get an idea of the generator output without having to actually run it for themselves....  that gets rid of more than 40k lines .

also generator/javascript/local_modules/   are a couple of customised node modules that are mostly the same as the original, but just far enough from stock that we have to copy them in, for now.

package.json and package-lock.json are autogenerated files and can also be ignored ( node generates these ) 

generator/javascript/test/made_tests.js   is part of the now comprehensive test suite for the node/javascript , and is about 70k, and its also autogenerated, so in theory doesn't really need to be comitted, but its important for people reviewing this to understand what its doing and how it works. so for this WIP and review i am including it .   

... but yes... this is a big PR which in my opinion actually makes the javascript implementation FINALLY compliant and i have auto-generated tests out the wazoo to proove it.  :-)

stephendade:
@davidbuzz - I've now got a chance to take a look at this. I'm getting the following error when running (via ``npm install``) with the latest ArduPilot MAVLink:
```
ERROR: /home/stephen/Documents/UAVCode/mavlink/message_definitions/v1.0/common.xml:3676:0:ERROR:SCHEMASV:SCHEMAV_CVC_COMPLEX_TYPE_3_2_1: Element 'field', attribute 'instance': The attribute 'instance' is not allowed.
/home/stephen/Documents/UAVCode/mavlink/message_definitions/v1.0/common.xml:4376:0:ERROR:SCHEMASV:SCHEMAV_CVC_COMPLEX_TYPE_3_2_1: Element 'field', attribute 'instance': The attribute 'instance' is not allowed.
/home/stephen/Documents/UAVCode/mavlink/message_definitions/v1.0/common.xml:4676:0:ERROR:SCHEMASV:SCHEMAV_CVC_COMPLEX_TYPE_3_2_1: Element 'field', attribute 'instance': The attribute 'instance' is not allowed.
/home/stephen/Documents/UAVCode/mavlink/message_definitions/v1.0/common.xml:4810:0:ERROR:SCHEMASV:SCHEMAV_CVC_COMPLEX_TYPE_3_2_1: Element 'field', attribute 'instance': The attribute 'instance' is not allowed.
/home/stephen/Documents/UAVCode/mavlink/message_definitions/v1.0/common.xml:5043:0:ERROR:SCHEMASV:SCHEMAV_CVC_COMPLEX_TYPE_3_2_1: Element 'field', attribute 'instance': The attribute 'instance' is not allowed.
/home/stephen/Documents/UAVCode/mavlink/message_definitions/v1.0/common.xml:5052:0:ERROR:SCHEMASV:SCHEMAV_CVC_COMPLEX_TYPE_3_2_1: Element 'field', attribute 'instance': The attribute 'instance' is not allowed.
/home/stephen/Documents/UAVCode/mavlink/message_definitions/v1.0/common.xml:5058:0:ERROR:SCHEMASV:SCHEMAV_CVC_COMPLEX_TYPE_3_2_1: Element 'field', attribute 'instance': The attribute 'instance' is not allowed.
```

amilcarlucas:
@stephendade that is caused by the recent #459 PR.

davidbuzz:
Now that this has undergone review, i've rebased it, dropped all the generated code, done whitespace cleanups, fixed tests so they pass again, and removed the WIP flag. 
Its still important to note that this IS a breaking commit for any previous users of the bindings, as they previously had attribute collissions that have been properly fixed by this.    Given the README.md for this binding says 'This implementation should be considered pre-beta', i think it's acceptable make this one-off breaking change in this case, as there really is no other choice.

stephendade:
I get a few errors running ``cd generator/javascript && npm test`` if I haven't run ``./test_generator.sh`` beforehand. Maybe some dependency between the two?

davidbuzz:
> I get a few errors running `cd generator/javascript && npm test` if I haven't run `./test_generator.sh` beforehand. Maybe some dependency between the two?    

Yes,  now that the tests are being cross-matched and tested against the C binding, its important that the C bindings are compiled first....   there is a test_gen_js.sh  script that compiles the needed bindings first and then runs all the js tests.  running  test_generator.sh first achieves basically the same result.    I also put some output into the npm 'pretest' stage in generator/javascript/package.json so that it emits a helpful message when u type 'npm test' ..it says: 
"consider running this if the test fails: cd ../.. && ./test_gen_js.sh"

I guess we could make this pre-test just always run test_generator.sh, but i don't think that needs to stop this PR going in.

stephendade:
> "consider running this if the test fails: cd ../.. && ./test_gen_js.sh"

Would be worth putting this in the README.md for the javascript generator, under the "Development" section.

test_gen_js.sh
```diff
@@ -0,0 +1,50 @@
...
+pushd generator/C/test/posix
+make clean testmav1.0_ardupilotmega testmav2.0_ardupilotmega testmav1.0_common testmav2.0_common
+./testmav1.0_ardupilotmega | grep '^fe' | node ../../../../examples/testparser.js ardupilotmega 1.0
+./testmav2.0_ardupilotmega | grep '^fd' | node ../../../../examples/testparser.js ardupilotmega 2.0
+./testmav1.0_common | grep '^fe' | node ../../../../examples/testparser.js common 1.0
+./testmav2.0_common | grep '^fd' | node ../../../../examples/testparser.js common 2.0
+popd
+
+
+# make big collection ~990 of mocha tests based on C output like the above but more thorough, this includes byte-level checking of the packed results
```
stephendade, line 42: This section can be simplified to:
```
cd generator/javascript/ 
npm test
cd ../..
```
Also avoids the need to have mocha globally installed.

davidbuzz:
> Would be worth putting this in the README.md for the javascript generator, under the "Development" section.

i have fixed that request in latest push of this PR by making test_gen_js.sh a little bit smarter, and so it can now be called either from the command-line or via the 'pretest' stage of 'npm test'.      it's also now less verbose by default, but u can easily change its verbosity by changing line 25.       testparser.js also understands the verbosity , and can be more or less shouty on the console as required.

davidbuzz:
In order to keep the pre-existing javascript implementation as-is with respect to its API and compatibility, i've cloned the original JavaScript generator adding the "_stable" extension, leaving it otherwise essentially alone and as-is.  Since Bruce was the original author credited with most of that work, i've committed on his behalf.  Right now both before and after this PR, anyone using the classic mavgen '--lang=JavaScript' will continue to use Bruce's API unchanged. I've then tweaked my PR so that it can be utilised with "--lang=JavaScript_NextGen" for user/s who choose to try it.  Keeps existing user/s backward compatible and also allows us to improve it moving forward.

davidbuzz:
CI checks now passing , i had permissions wrong on a couple of files, forgot to chmod them.

stephendade:
Awesome! Could we add a deprecation warning to the stable JavaScript? It would help encourage users to upgrade.

Otherwise happy with this.

davidbuzz:
> Awesome! Could we add a deprecation warning to the stable JavaScript? It would help encourage users to upgrade.

Depreciation notice  for '_stable' added to it's readme, as requested.

stephendade:
Great! Happy for this to go in.

davidbuzz:
discussed it with @peterbarker   and did a few cleanups as per his request...   moved C++ error/s to separate PR (merged) , split local_modules into their own commit to clarify their history better, and properly prefix commits with 'Generator:'.   I've also dropped min.py from this pr as it's not strictly java script related..  leaving in min.js thoug as it is. :-)

tridge (approved):

PR #434, opened 2020-07-11
fix python2 deprecation

rbeall:


peterbarker:
This doesn't look like it's present in Python2's libraries?  Has it been tested on Python2?

We may need to do a "running_python3" sort of thing here.

rbeall:
Yes concur Peter.  Not sure where you guys would want to take it from a python2 perspective.  As is it doesn't work in python3 and adding my fix will deprecate it from working in python2.  Not familiar with the cleanest way to use both.

peterbarker:
@rbeall something like https://github.com/ardupilot/ardupilot/blob/master/Tools/scripts/uploader.py#L89 would work.

peterbarker:
@rbeall are you still interested in chasing this one?

peterbarker:
@rbeall I merged a modified version of this into master.

I suggest you might want to investigate the use of git "topic" branches; pushing and PR'ing from your master branch isn't recommended :-)

PR #438, opened 2020-07-30
generator: fix non-ASCII character

chobitsfan:
Fix visual studio warning `C4819: The file contains a character that cannot be represented in the current code page`. It makes errors while building AirSim.


peterbarker:
Merged, thanks!

PR #439, opened 2020-07-30
Avoid 'msgid' byte override for messages with 'id' field

Vavooon:
Since lot of messages contain 'id' field and all fields are stored directly in message object, setting 'id' field on message construction automatically overrides message's own 'id' property used for storing 'msgid' byte. Here I use 'msgid' property name instead (no message uses such field name).

### DIFF
--- a/generator/mavgen_javascript.py
+++ b/generator/mavgen_javascript.py
@@ -148,7 +148,7 @@ ${MAVHEAD}.message.prototype.pack = function(mav, crc_extra, payload) {
 
     t.write(outf, """
     var incompat_flags = 0;
-    this.header = new ${MAVHEAD}.header(this.id, this.payload.length, mav.seq, mav.srcSystem, mav.srcComponent, incompat_flags, 0,);    
+    this.header = new ${MAVHEAD}.header(this.msgid, this.payload.length, mav.seq, mav.srcSystem, mav.srcComponent, incompat_flags, 0,);
     this.msgbuf = this.header.pack().concat(this.payload);
     var crc = ${MAVHEAD}.x25Crc(this.msgbuf.slice(1));
 
@@ -234,7 +234,7 @@ ${COMMENT}
         outf.write("""
 
     this.format = '%s';
-    this.id = %s.MAVLINK_MSG_ID_%s;
+    this.msgid = %s.MAVLINK_MSG_ID_%s;
     this.order_map = %s;
     this.crc_extra = %u;
     this.name = '%s';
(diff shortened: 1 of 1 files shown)


peterbarker:
On Thu, 30 Jul 2020, Vavooon wrote:

> Since lot of messages contain 'id' field and all fields are stored directly in message object, setting 'id' field on message construction automatically
> overrides message's own 'id' property used for storing 'msgid' byte. Here I use 'msgid' property name instead (no message uses such field name).

I think we should probably discuss APIs here.

Getting the source system from a message isn't done by accessing a field - 
there's an accessor for it.

Do you see any good reason not to add an accessor for the message ID? 
Something similar can be done for the Python bindings - and perhaps we add 
a decorator to warn anybody currently using the id field that the API is 
changing and leave it like that for a couple of years...

In terms of storage - assuming this has to go into the same object as the 
rest of the data (I don't know these bindings well) - perhaps using 
something *way* outside sensible scope for a mavlink message field name 
would be appropriate - __mavlink_msg_id__ for example...

Vavooon:
Not quite sure what you mean by 'accessor' (is it like a getter is JS?). Do you mean this line 
`this.header = new new mavlink20.header(this.id, this.payload.length, mav.seq, mav.srcSystem, mav.srcComponent, incompat_flags, 0,);`
where we read `seq`, `srcSystem` and `srcComponent` properties from `mav` object (which is instance of `MAVLink20Processor`)?
Not sure we could store `msgid` there since each message should have it's own value, unlike `seq` or `srcSystem`.

Do you have an idea how we could warn users about API changes? Should we print warning to console?

Renaming the field into `__mavlink_msg_id__` is a good idea, however it would require renaming of other similar fields too (like `name`).

Just checked common dialect for all fields called `id`: there are 10 messages which can't be properly packed (so they are never recognized by other systems) and reading `id` on JS side would point to field `id` instead of `msgid`.
There are also 5 messages with field `name` which also overrides message`s own field where string representation of message ID is stored.
I didn't check other common fields though.

davidbuzz:
This PR has been superceeded (and fixed) by the recent JS update/s , which now include a test suite and fixes for this.  https://github.com/ArduPilot/pymavlink/pull/431      it renames msg.id to msg._id as well as others like .name to ._name and .header to ._header etc.     it breaks backward compatability ,so we've forked the old JS bindings to _Stable for historical sake.

davidbuzz:
oh... and @Vavooon  - id like to draw your attention to https://github.com/ArduPilot/pymavlink/pull/482   which i hope will be merged soon, which adds Mavlink2 packet signing (and tests for it) to the JS/Node bindings too.    :-)

Vavooon:
Thanks a lot for the fixes. I'll try it when I get back to that project. Signing would be useful too.

PR #441, opened 2020-07-30
Use absolute import path in file generated by mavgen_python

ar13pit:
The current generator uses a relative path from the directory `dialects/v<X>` to look for `mavcrc` module in the `generator` module. As a result, any custom generated files need to be placed in the `pymavlink` sources.

With this PR, the generated file can be placed inside the module of the user's choice.

### DIFF
--- a/generator/mavgen_python.py
+++ b/generator/mavgen_python.py
@@ -31,7 +31,7 @@ from builtins import range
 from builtins import object
 import struct, array, time, json, os, sys, platform
 
-from ...generator.mavcrc import x25crc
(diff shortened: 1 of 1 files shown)


ar13pit:
Can anyone please look into this? Its a very small contribution but does save a lot of hassle of tooling for placing the generated code into a custom module.

peterbarker:
Tested and working, thanks.

I've adjusted the commit message a little and will merge once CI passes.

peterbarker:
Merged, thanks!

PR #435, opened 2020-07-16
Added verification of ascii inside unicode conditional

mday299:
Added verification of ascii inside unicode conditional

PR #447, opened 2020-08-18
This adds a conditional that prevents unicode values from being added as float values.

QuPsi:
Continued work from

https://github.com/ArduPilot/pymavlink/pull/435

### DIFF
--- a/tools/mavtomfile.py
+++ b/tools/mavtomfile.py
@@ -8,10 +8,13 @@ from builtins import range
 
 import os
 import re
+import sys
 from pymavlink import mavutil
 
 def process_tlog(filename):
     '''convert a tlog to a .m file'''
+
+    is_py2 = sys.version_info < (3,0)
 
     print("Processing %s" % filename)
 
@@ -62,6 +65,11 @@ def process_tlog(filename):
             f.write("%s.columns = {'timestamp'" % mtype)
             for field in fieldnames:
                 val = getattr(m, field)
+
+                if is_py2:
+                    if isinstance(val,unicode): # NOQA
+                        val = str(val)
+
                 if not isinstance(val, str):
                     if type(val) is not list:
                         f.write(",'%s'" % field)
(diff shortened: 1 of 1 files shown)


peterbarker:
Can't tell you how much I love Python2 vs Python3 shenanigans.

From: https://github.com/ArduPilot/pymavlink/blob/master/DFReader.py#L238

```
        is_py2 = sys.version_info < (3,0)
```

```
            if is_py2:
                if isinstance(v,unicode): # NOQA
                    v = str(v)
            else:
                if isinstance(v,str):
```

peterbarker:
Not sure what's up with that extra commit.  Seems out-of-place.

QuPsi:
That was my mistake. I had committed an nonupdated version of the file and @mday299 was walking me through it. 

Great find on the python differences. That definitely seems to be the root of the issue!

stephendade:
Yes, you'll need to remove that commit. Not sure how it got in there - possibly rebasing gone wrong.

peterbarker:
@QuPsi are you still chasing this one?

QuPsi:
Made an even simpler change.... just changed the values to 'str'.

I tried to implement your suggestion @peterbarker , however, simply having the work 'unicode' in the script upset versions >3.0.

The checks have passed and the script runs as intended on my side.

peterbarker:
I can see code which is run later under this which is wondering if the type is `list`.  Since you've just forced it to string, that can never be the case now, so I don't think this change can be correct.

Note `DFReader.py` - which I know works under both Python2 and Python3 - uses `unicode` bare: https://github.com/ArduPilot/pymavlink/blob/master/DFReader.py#L239

QuPsi:
@peterbarker Have a look at my latest commit when you can please. I think I got it now.

peterbarker:
@QuPsi looks solid.

Has it been tested on both Py2 and Py3?

peterbarker:
Please squash the commits down into a single commit.  Let me know if you'd like some direction on how to do that (or let me know and I'll do it before merging)

QuPsi:
Ok, I think I got the commits squashed now and yes, tested in both py2 and py3!

peterbarker:
I've touched up the commit message a bit and removed some introduced whitespace.  Will merge once CI is done.

peterbarker:
Merged, thanks!

QuPsi:
Thank you so much! Learned a lot! I really appreciate the assistance!

mday299:
Is this scheduled to be pulled upstream into mavlink and then into ardupilot soon?

peterbarker:
@mday299 tridge was muttering about doing another pymavlink release recently, but there's no schedule as such.

PR #451, opened 2020-08-26
add port

Yu-ZHOU-1994:
add udp port 14540,14570, and add tcp port 4560

### DIFF
--- a/generator/mavgen_wlua.py
+++ b/generator/mavgen_wlua.py
@@ -447,8 +447,13 @@ wtap_encap:add(wtap.USER0, mavlink_proto)
 -- bind protocol dissector to port 14550 and 14580
 
 local udp_dissector_table = DissectorTable.get("udp.port")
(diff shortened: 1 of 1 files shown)


peterbarker (changes requested):
There should only be one commit in here, and it needs a better commit message.

I suggest, `mavgen_lua: add udp port 14540,14570, and add tcp port 4560`

@WickedShell your thoughts on this, please....

peterbarker:
Problem with this sort of PR is that we'll end up trying to decode mavlink on *every* port eventually :-)  We really need to stick to the common ones or people will find their wireshark slowing down as it is always trying to decode mavlink.

Instead, I suggest you use the "Decode As..." dialogue box and add the ports to your local wireshark configuration.  You specify the input method (e.g. UDP), the port number and the protocol to attempt to decode from that source - and then it just works.

image

Closing this - thanks anyway!

PR #454, opened 2020-09-06
README.md: Encourage Windows users to install current lxml

cclauss:
[Current lxml](https://pypi.org/project/lxml) is 4.5.2, not 3.6.0.  This matches with https://github.com/ArduPilot/pymavlink/blob/master/requirements.txt#L1

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ Examples can be found [in the repository](examples/) or in the [ArduSub book](ht
 
 Pymavlink supports both Python 2 and Python 3.
 
-The following instructions assume you are using Python 3 and a Debian-based (like Ubuntu) installation.
+The following instructions assume you are using Python 2.7 and a Debian-based (like Ubuntu) installation.
 
 .. note::
 
@@ -29,7 +29,7 @@ Pymavlink has several dependencies :
 
     - [future](http://python-future.org/) : for Python 2 and Python 3 interoperability
     - [lxml](http://lxml.de/installation.html) : for checking and parsing xml file 
-    - python3-dev : for mavnative (or python-dev for Python 2)
+    - python-dev : for mavnative (or python3-dev for Python 3)
     - a C compiler : for mavnative
 
 Optional :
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,101 +1,96 @@
...
-The -U parameter allow to update pymavlink version if it is already installed.
-
 #### Mavnative
 
 By default, pymavlink will try to compile and install mavnative which is a C extension for parsing mavlink. Mavnative only supports mavlink1.
 To skip mavnative installation and reduce dependencies like `gcc` and `python-dev`, you can pass `DISABLE_MAVNATIVE=True` environment variable to the installation command:
 
 ```bash
-sudo DISABLE_MAVNATIVE=True pip2 install -U pymavlink
+sudo DISABLE_MAVNATIVE=True python3 -m pip install --upgrade pymavlink
```
peterbarker, line 77: Are there systems out there where just `python` shouldn't work?  Do we need to specify a Python here?

peterbarker:
@stephendade this obviously impact the Windows port for MAVProxy.

README.md
cclauss: [PEP020](https://www.python.org/dev/peps/pep-0020): _Explicit is better than implicit..._

Also, https://snarky.ca/why-you-should-use-python-m-pip

On my Mac:
% `python --version ` # Python 2.7.16
% `python2 --version ` # Python 2.7.16
% `python3 --version ` # Python 3.8.5

We don't really want to encourage the use of legacy Python if we don't have to.

peterbarker:
On Sun, 6 Sep 2020, Christian Clauss wrote:

> On my Mac:
> % python --version # Python 2.7.16
> % python2 --version # Python 2.7.16
> % python3 --version # Python 3.8.5
> 
> We don't really want to encourage the use of legacy Python if we don't have to.

So wait - after running those instructions, the software wouldn't actually 
work *for you*?!

I know we shouldn't encourage people to run Python2 - it's going to be a 
fantastic day when we lose the ability to do so (maybe that impending pip 
change!).  I look forward to removing the "sys.version" hacks.

But we don't want to increase support load by giving out instructions that 
won't lead to working systems.

cclauss:
Reverted to legacy Python

stephendade:
>@stephendade this obviously impact the Windows port for MAVProxy.

The Windows installer for MAVProxy always uses the latest stable lxml, so no issues there.

README.md
khancyr: on ubuntu 20.04 by default python doesn't exist ... only python3

cclauss: Those changes were reverted.

peterbarker:
On Sun, 6 Sep 2020, Pierre Kancir wrote:

> on ubuntu 20.04 by default python doesn't exist ... only python3

So we could suggest they install python-is-python3 or modify the 
instructions?

cclauss:
More heat than light... https://www.python.org/dev/peps/pep-0394/

peterbarker:
I've rebased this on changes already taken from here.  While our scripts continue to reference a "python" command, I'm not inclined to recommend users explicitly run `python3` as they'll end up with a non-functional installation.

@cclauss do you disagree?  Is there anything else in this PR that should be brought in?

cclauss:
Closing... These changes were made elsewhere.

PR #449, opened 2020-08-21
Travis CI: Upgrade Node.js and Python

cclauss:
@peterbarker
* https://nodejs.org/en/about/releases/
* https://www.python.org/download/pre-releases/

~Python 3.9 fc1 is failing because numpy is not installed.~

`bytes` is a Python built-in data types so let's use `bytes_array` as the variable name.

PR #456, opened 2020-09-08
array.array.fromstring() was removed in Python 3.9

peterbarker:
https://docs.python.org/3.9/library/array.html#array.array.frombytes

From [What’s New In Python 3.9](https://docs.python.org/3.9/whatsnew/3.9.html#removed))
> array.array: tostring() and fromstring() methods have been removed. They were aliases to tobytes() and frombytes(), deprecated since Python 3.2.

Swiped from https://github.com/ArduPilot/pymavlink/pull/449/files

### DIFF
--- a/generator/mavcrc.py
+++ b/generator/mavcrc.py
@@ -31,5 +31,8 @@ class x25crc(object):
         accum = self.crc
         import array
         bytes = array.array('B')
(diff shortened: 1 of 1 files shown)


peterbarker:
@cclauss I swiped this into a new PR as it looks like a good stand-alone fix.

PR #468, opened 2020-10-19
mavutil: Allow for custom mapping of vehicle modes

sypaq-nexton:
We're doing some experimentation on vehicle types and vehicle modes. However, the hard-coded map of vehicle type to mode name makes this difficult, as it prevents us from easily using MAVProxy for testing.

This PR allows the user to create a `ap_custom_mode_map.py` file that contains a custom mode mapping and have it loaded when `pymavlink` is initialised.

### DIFF
--- a/mavutil.py
+++ b/mavutil.py
@@ -2009,19 +2009,28 @@ AP_MAV_TYPE_MODE_MAP_DEFAULT = {
 
 
 try:
-    # Allow for using custom mode maps by importing a dict named
-    # AP_MAV_TYPE_MODE_MAP_CUSTOM from "~/.pymavlink/ap_custom_mode_map.py"
-    # and using it to extend the hard-coded AP_MAV_TYPE_MODE_MAP_DEFAULT dict.
-    import sys
+    # Allow for using custom mode maps by importing a JSON dict from
+    # "~/.pymavlink/custom_mode_map.json" and using it to extend the hard-coded
+    # AP_MAV_TYPE_MODE_MAP_DEFAULT dict.
     from os.path import expanduser
 
-    sys.path.append(expanduser('~') + '/.pymavlink')
-    from ap_custom_mode_map import AP_MAV_TYPE_MODE_MAP_CUSTOM
+    _custom_mode_map_path = expanduser('~/.pymavlink/custom_mode_map.json')
+    with open(_custom_mode_map_path) as f:
+        _json_mode_map = json.load(f)
+
+    try:
+        _custom_mode_map = {}
+        for mav_type, mode_map in _json_mode_map.items():
+            # make sure the custom map has the right datatypes
+            _custom_mode_map[int(mav_type)] = { int(mode_num): str(mode_name) for mode_num, mode_name in mode_map.items() }
+    except:
+        print("Error: invalid pymavlink custom mode map dict in " + _custom_mode_map_path)
+        raise
 
     AP_MAV_TYPE_MODE_MAP = AP_MAV_TYPE_MODE_MAP_DEFAULT.copy()
-    AP_MAV_TYPE_MODE_MAP.update(AP_MAV_TYPE_MODE_MAP_CUSTOM)
-    print("Using custom MAV_TYPE mode map:")
-    print(repr(AP_MAV_TYPE_MODE_MAP_CUSTOM))
+    AP_MAV_TYPE_MODE_MAP.update(_custom_mode_map)
+    print("Using pymavlink custom MAV_TYPE mode map:")
+    print(repr(_custom_mode_map))
 except:
     AP_MAV_TYPE_MODE_MAP = AP_MAV_TYPE_MODE_MAP_DEFAULT
 
--- a/mavutil.py
+++ b/mavutil.py
@@ -2030,8 +2030,6 @@ try:
 
     AP_MAV_TYPE_MODE_MAP = AP_MAV_TYPE_MODE_MAP_DEFAULT.copy()
     AP_MAV_TYPE_MODE_MAP.update(_custom_mode_map)
-    print("Using pymavlink custom MAV_TYPE mode map:")
-    print(repr(_custom_mode_map))
 except:
     AP_MAV_TYPE_MODE_MAP = AP_MAV_TYPE_MODE_MAP_DEFAULT
 
--- a/mavutil.py
+++ b/mavutil.py
@@ -2016,8 +2016,12 @@ try:
 
     _custom_mode_map_path = os.path.join("~", ".pymavlink", "custom_mode_map.json")
     _custom_mode_map_path = expanduser(_custom_mode_map_path)
-    with open(_custom_mode_map_path) as f:
-        _json_mode_map = json.load(f)
+    try:
+        with open(_custom_mode_map_path) as f:
+            _json_mode_map = json.load(f)
+    except:
+        print("Error: pymavlink custom mode file ('" + _custom_mode_map_path + "') is not valid JSON.")
+        raise
 
     try:
         _custom_mode_map = {}
--- a/mavutil.py
+++ b/mavutil.py
@@ -1985,7 +1985,7 @@ mode_mapping_sub = {
     19: 'MANUAL',
 }
 
-AP_MAV_TYPE_MODE_MAP = {
+AP_MAV_TYPE_MODE_MAP_DEFAULT = {
     # copter
     mavlink.MAV_TYPE_HELICOPTER:  mode_mapping_acm,
     mavlink.MAV_TYPE_TRICOPTER:   mode_mapping_acm,
@@ -2006,6 +2006,25 @@ AP_MAV_TYPE_MODE_MAP = {
     # sub
     mavlink.MAV_TYPE_SUBMARINE: mode_mapping_sub,
 }
+
+
+try:
+    # Allow for using custom mode maps by importing a dict named
+    # AP_MAV_TYPE_MODE_MAP_CUSTOM from "~/.pymavlink/ap_custom_mode_map.py"
+    # and using it to extend the hard-coded AP_MAV_TYPE_MODE_MAP_DEFAULT dict.
+    import sys
+    from os.path import expanduser
+
+    sys.path.append(expanduser('~') + '/.pymavlink')
+    from ap_custom_mode_map import AP_MAV_TYPE_MODE_MAP_CUSTOM
+
+    AP_MAV_TYPE_MODE_MAP = AP_MAV_TYPE_MODE_MAP_DEFAULT.copy()
+    AP_MAV_TYPE_MODE_MAP.update(AP_MAV_TYPE_MODE_MAP_CUSTOM)
+    print("Using custom MAV_TYPE mode map:")
+    print(repr(AP_MAV_TYPE_MODE_MAP_CUSTOM))
+except:
+    AP_MAV_TYPE_MODE_MAP = AP_MAV_TYPE_MODE_MAP_DEFAULT
+
 
 # map from a PX4 "main_state" to a string; see msg/commander_state.msg
 # This allows us to map sdlog STAT.MainState to a simple "mode"
--- a/mavutil.py
+++ b/mavutil.py
@@ -1919,7 +1919,8 @@ mode_mapping_apm = {
     22 : 'QAUTOTUNE',
     23 : 'QACRO',
     24 : 'THERMAL',
-    }
+}
+
 mode_mapping_acm = {
     0 : 'STABILIZE',
     1 : 'ACRO',
@@ -1946,6 +1947,7 @@ mode_mapping_acm = {
     23 : 'FOLLOW',
     24 : 'ZIGZAG',
 }
+
 mode_mapping_rover = {
     0 : 'MANUAL',
     1 : 'ACRO',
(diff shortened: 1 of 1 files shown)


sypaq-nexton:
Example `ap_custom_mode_map.py` file attached: [ap_custom_mode_map.py.txt](https://github.com/ArduPilot/pymavlink/files/5521436/ap_custom_mode_map.py.txt)

Note that it's a Python file -- the `.txt` extension was required for GitHub upload.

amilcarlucas:
Hello, I like the approach on ardupilot/mavlink#86  ardupilot/mavlink#87  ardupilot/mavlink#88  ardupilot/mavlink#89 better.

joshanne:
@amilcarlucas slightly confused here. All the linked PRs here are changes to the xml files.
Meaning if we wish to change a description, or add a new mode, we essentially require to reinstall `pymavlink` with the new message definitions.

The intent of this PR is to allow for adding/overriding mode maps without the need to reinstall `pymavlink`. Is there a better way of achieving this outcome?

amilcarlucas:
The problem I am trying to addess here is: Get it to fully work, not just a bit working.
With current solutions, you can easily add a flight mode, yes. But you do not add it to the groundstation. Lots of users then come along and ask how to add it to thie groundstation. To do so they need to either add it in mavlink and then in pymavlink, or hardcode it in the ground station.

What I am doing here is forcing people to having to do it in mavlink once, then reconpile and you get that mode as a define in ardupilot and in the GCS. You still need to tell the GCS to use your localy edited .xml mavlink defenition though.

I hope this is a bit clearer now.
Or we can talk in discord.

joshanne:
I think we're talking about two different use cases here.

The end user shouldn't be required to re-install `pymavlink` simply to attain one (or more) new mode(s), rename a mode or add modes for MAV_TYPES not supported by pymavlink.

eg: The following code block performs the `mode_mapping` in pymavlink. If the user wishes to add/rename modes or use MAV_TYPES not present in mavproxy, there's more work than just updating an xml. eg. What if I wish to use MAV_TYPE_ROCKET or MAV_TYPE_AIRSHIP? 

```
    def mode_mapping(self):
        # ... removed ...
        if mav_type in [mavlink.MAV_TYPE_QUADROTOR,
                        mavlink.MAV_TYPE_HELICOPTER,
                        mavlink.MAV_TYPE_HEXAROTOR,
                        mavlink.MAV_TYPE_OCTOROTOR,
                        mavlink.MAV_TYPE_DODECAROTOR,
                        mavlink.MAV_TYPE_COAXIAL,
                        mavlink.MAV_TYPE_TRICOPTER]:
            map = mode_mapping_acm
        if mav_type == mavlink.MAV_TYPE_FIXED_WING:
            map = mode_mapping_apm
        if mav_type == mavlink.MAV_TYPE_GROUND_ROVER:
            map = mode_mapping_rover
        if mav_type == mavlink.MAV_TYPE_SURFACE_BOAT:
            map = mode_mapping_rover # for the time being
        if mav_type == mavlink.MAV_TYPE_ANTENNA_TRACKER:
            map = mode_mapping_tracker
        if mav_type == mavlink.MAV_TYPE_SUBMARINE:
            map = mode_mapping_sub
        if map is None:
            return None
```

As I said, there should be no need for a user/developer to require installing pymavlink with modified xml's in order to retrieve the new modes, names, vehicle types, etc.

khancyr (approved):
I like the idea. It simplifies the current mapping code and allow to load custom mode map optionally, so the current behavior isn't changed.

sypaq-nexton:
Thanks @joshanne, @khancyr and @amilcarlucas. 

I've rebased this PR onto `master` to fix up the merge conflicts.

mavutil.py
```diff
@@ -2003,7 +1982,48 @@ def mode_string_v09(msg):
...
+
+
+try:
+    # Allow for using custom mode maps by importing a dict named
+    # AP_MAV_TYPE_MODE_MAP_CUSTOM from "~/.pymavlink/ap_custom_mode_map.py"
+    # and using it to extend the hard-coded AP_MAV_TYPE_MODE_MAP_DEFAULT dict.
+    import sys
+    from os.path import expanduser
+
+    sys.path.append(expanduser('~') + '/.pymavlink')
```
tridge, line 2017: if doing this then the sys.path.append needs to be conditional on the existance of ~/.pymavlink/ap_custom_mode_map as otherwise it polluting the system path for all modules
but I'd much this was a data file, such as json format, in ~/.pymavlink/modemap.json, and it parses that file to replace (or augment?) the dictionary. I think that would be more portable and less likely to break things

tridge (changes requested):
i'd be happy with this if it was a modemap.json file instead of a py file

nickexton:
> 
> i'd be happy with this if it was a modemap.json file instead of a py file

Ack. I originally used a `.py` file for two reasons:

1. Laziness; and
2. It enabled me to import and use the `pymavlink` typedefs in the custom map.

nickexton:
I've updated to import from a JSON file located at `~/.pymavlink/custom_mode_map.json`.

File contents look like:
```json
{
    "1": {
        "0":  "MANUAL",
        "1":  "CIRCLE",
        "2":  "STABILIZE",
        "3":  "TRAINING",
        "4":  "ACRO",
        "5":  "FBWA",
        "6":  "FBWB",
        "7":  "CRUISE",
        "8":  "AUTOTUNE",
        "10": "AUTO",
        "11": "RTL",
        "12": "LOITER",
        "13": "TAKEOFF",
        "14": "AVOID_ADSB",
        "15": "GUIDED",
        "16": "INITIALISING",
        "17": "QSTABILIZE",
        "18": "QHOVER",
        "19": "QLOITER",
        "20": "QLAND",
        "21": "QRTL",
        "22": "QAUTOTUNE",
        "23": "QACRO"
    }
}
```

mavutil.py
```diff
@@ -2004,7 +1983,57 @@ def mode_string_v09(msg):
...
+}
+
+
+try:
+    # Allow for using custom mode maps by importing a JSON dict from
+    # "~/.pymavlink/custom_mode_map.json" and using it to extend the hard-coded
+    # AP_MAV_TYPE_MODE_MAP_DEFAULT dict.
+    from os.path import expanduser
+
+    _custom_mode_map_path = expanduser('~/.pymavlink/custom_mode_map.json')
```
tridge, line 2017: this is not portable to windows. I think expanduser would work, but the / in the path won't. 
It needs to use os.path.join so the right path separator is used

mavutil.py
```diff
@@ -2004,7 +1983,57 @@ def mode_string_v09(msg):
...
+        for mav_type, mode_map in _json_mode_map.items():
+            # make sure the custom map has the right datatypes
+            _custom_mode_map[int(mav_type)] = { int(mode_num): str(mode_name) for mode_num, mode_name in mode_map.items() }
+    except:
+        print("Error: invalid pymavlink custom mode map dict in " + _custom_mode_map_path)
+        raise
+
+    AP_MAV_TYPE_MODE_MAP = AP_MAV_TYPE_MODE_MAP_DEFAULT.copy()
+    AP_MAV_TYPE_MODE_MAP.update(_custom_mode_map)
+    print("Using pymavlink custom MAV_TYPE mode map:")
```
tridge, line 2032: i think printing it out each time could get annoying, especially with tools like mavlogdump and MAVExplorer

tridge (changes requested):

mavutil.py
nickexton: Ack. Updated.

mavutil.py
nickexton: Ack, removed.

sypaq-nexton:
Added an error message for when `custom_mode_map.json` is not a valid JSON file.

Thanks @joshanne!

mavutil.py
```diff
@@ -2004,7 +1983,60 @@ def mode_string_v09(msg):
...
+    # "~/.pymavlink/custom_mode_map.json" and using it to extend the hard-coded
+    # AP_MAV_TYPE_MODE_MAP_DEFAULT dict.
+    from os.path import expanduser
+
+    _custom_mode_map_path = os.path.join("~", ".pymavlink", "custom_mode_map.json")
+    _custom_mode_map_path = expanduser(_custom_mode_map_path)
+    try:
+        with open(_custom_mode_map_path) as f:
+            _json_mode_map = json.load(f)
+    except:
```
tridge, line 2022: needs to be "except Exception:"

mavutil.py
```diff
@@ -2004,7 +1983,60 @@ def mode_string_v09(msg):
...
+        print("Error: pymavlink custom mode file ('" + _custom_mode_map_path + "') is not valid JSON.")
+        raise
+
+    try:
+        _custom_mode_map = {}
+        for mav_type, mode_map in _json_mode_map.items():
+            # make sure the custom map has the right datatypes
+            _custom_mode_map[int(mav_type)] = { int(mode_num): str(mode_name) for mode_num, mode_name in mode_map.items() }
+    except:
```
tridge, line 2031: needs to be "except Exception:"

mavutil.py
```diff
@@ -2004,7 +1983,60 @@ def mode_string_v09(msg):
...
+    # AP_MAV_TYPE_MODE_MAP_DEFAULT dict.
+    from os.path import expanduser
+
+    _custom_mode_map_path = os.path.join("~", ".pymavlink", "custom_mode_map.json")
+    _custom_mode_map_path = expanduser(_custom_mode_map_path)
+    try:
+        with open(_custom_mode_map_path) as f:
+            _json_mode_map = json.load(f)
+    except:
+        print("Error: pymavlink custom mode file ('" + _custom_mode_map_path + "') is not valid JSON.")
```
tridge, line 2023: need to not throw an error if the file doesn't exist

tridge (changes requested):

mavutil.py
joshanne: Now only catches `JSONDecodeError`. Any other failure to open or load the json reverts to default mode map.

joshanne: Error only thrown on failure to load json.

mavutil.py
joshanne: Ack. Added.

mavutil.py
joshanne: Ack. Added.

peterbarker:
@khancyr any idea what went wrong with CI here?

We just merged something to our master mavlink repo with storm.xml in it - a new file, which doesn't happen very often...

khancyr:
@peterbarker path to the storm32.xml is wrong. The external directory live into the mavlink directory like messages_definition, not into pymavlink

peterbarker:
On Wed, 17 Feb 2021, Pierre Kancir wrote:

> @peterbarker path to the storm32.xml is wrong. The external directory live into the mavlink directory like messages_definition, not into pymavlink

Where's the path stored?

sypaq-nexton:
> On Wed, 17 Feb 2021, Pierre Kancir wrote: @peterbarker path to the storm32.xml is wrong. The external directory live into the mavlink directory like messages_definition, not into pymavlink
> Where's the path stored?

It's in the `mavlink` repo here: https://github.com/mavlink/mavlink/blob/35adbf09da7f612393b9179a9ba02f5e4fb67c81/message_definitions/v1.0/all.xml#L25

Looks like it's using relative directories, which breaks when `pymavlink` isn't built from within the `mavlink` repo.

sypaq-nexton:
I've submitted a PR with a (hopefully temporary) work-around to address these build issues: https://github.com/ArduPilot/pymavlink/pull/506

peterbarker:
Alternative work-around for the storm32 problem: https://github.com/ArduPilot/mavlink/pull/197

@hamishwillee @olliw42 @auturgy we should move that `external` folder into `message_definitions`, or remove it entirely and just put (e.g.) `storm32.xml` alongside (e.g.) `icarous.xml`.

peterbarker:
Also @amilcarlucas

peterbarker:
Merging, thanks for your patience @sypaq-nexton

sypaq-nexton:
Great, thanks all!

amilcarlucas:
Is the json file format documented in pymavlink documentation ?

sypaq-nexton:
> 
> Is the json file format documented in the pymavlink documentation ?

@amilcarlucas: It's not documented anywhere. If you let me know where it needs to be documented, I'm happy to do it. Is the `pymavlink` README the best place?

olliw42:
> Looks like it's using relative directories, which breaks when pymavlink isn't built from within the mavlink repo.

I don't understand the issue

I can call mavgen from "everywhere", especially from outside the pymavlink and the mavlink repos. I have tested this on my system extensively and use it since a while all the time for 4 different projects ... 

so, frankly, I think you should investigate what pymavlink is actually doing "when pymavlink isn't built from within the mavlink repo"

sypaq-nexton:
> so, frankly, I think you should investigate what pymavlink is actually doing "when pymavlink isn't built from within the mavlink repo"

@olliw42: you're right, this is a `pymavlink` problem. The issue seems to be that the generator makes a copy of the `mavlink` `message_definitions` folder and builds from there. As the `storm32.xml` file is _outside_ of the `message_definitions` folder, the generator can't resolve the relative directories.

olliw42:
great to hear the issue seems to be understood, so should be easy to fix pymavlink :)

amilcarlucas:
@sypaq-nexton yes I think that the pymavlink README.md is the best place to do it.

PR #470, opened 2020-10-20
tools: mavkml: adjust shebang line to use Python from PATH

### LINKED ISSUE #469: Bad python interpreter header causes Yocto recipe failure
I'm pulling pymavlink into my Yocto image build using a custom recipe, and the recipe is failing during the QA check phase.  The QA check is verifying that all the scripts are executable, but it chokes when it comes to this one:
/tools/mavkml.py
The problem is caused by the python interpreter specification in the first line of the script:
"#!/usr/bin/python"
Which causes my Yocto build to look for a Python 2 interpreter.  Since I'm installing it for Python 3, the script should instead use:
"#!/usr/bin/env python"
This correct specifier is used in all the pymavlink scripts except for mavkml.py.
Once this is fixed, I won't have to install a patch, or disable my Yocto QA check.
Thanks.
   dmillard-TD

peterbarker:
Closes #469

### DIFF
--- a/tools/mavkml.py
+++ b/tools/mavkml.py
@@ -1,4 +1,4 @@
-#!/usr/bin/python
+#!/usr/bin/env python
 
 '''
 simple kml export for logfiles


khancyr (commented):
Lgtm

PR #473, opened 2020-11-02
generator: mavlink_channel_t with correct length

julianoes:
For cases where more than 4 mavlink channels are required, the mavlink_channel_t does not match. Therefore, I suggest that the enum needs to be provided separately if the default channel number is not used.

### DIFF
--- a/generator/C/include_v2.0/mavlink_types.h
+++ b/generator/C/include_v2.0/mavlink_types.h
@@ -161,43 +161,25 @@ typedef struct __mavlink_message_info {
 #define mavlink_ck_a(msg) *((msg)->len + (uint8_t *)_MAV_PAYLOAD_NON_CONST(msg))
 #define mavlink_ck_b(msg) *(((msg)->len+(uint16_t)1) + (uint8_t *)_MAV_PAYLOAD_NON_CONST(msg))
 
-/*
- * Applications can set MAVLINK_COMM_NUM_BUFFERS to the maximum number of
- * buffers they require. If more channels are used than allocated the result
- * will be a stack overrun.
- *
- * If MAVLINK_COMM_NUM_BUFFERS is defined outside, the typedef enum mavlink_channel_t
- * also needs to be provided with the appropriate numbers of channels.
- */
-#ifndef MAVLINK_COMM_NUM_BUFFERS
-#if (defined linux) | (defined __linux) | (defined  __MACH__) | (defined _WIN32)
-# define MAVLINK_COMM_NUM_BUFFERS 16
-typedef enum {
-    MAVLINK_COMM_0,
-    MAVLINK_COMM_1,
-    MAVLINK_COMM_2,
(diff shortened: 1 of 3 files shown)


len0rd:
Hey, nice work! This is similar to a [PR I opened](https://github.com/ArduPilot/pymavlink/pull/426) up a few months ago in response to an issue I made about this ( #404 ). Your change is more along the lines of the fix I was thinking when I made my PR originally.

Unfortunately, to preserve backwards compatibility, it may be necessary to maintain the default definition of `mavlink_channel_t` somehow. (Perhaps some type of ifndef wrapper similar to my PR)?

julianoes:
Thanks for the comment @len0rd. I would be happy with either PR.

hamishwillee:
@peterbarker Could you please review this?

hamishwillee:
@peterbarker ?

generator/C/include_v2.0/mavlink_types.h
```diff
@@ -161,23 +161,43 @@ typedef struct __mavlink_message_info {
 #define mavlink_ck_a(msg) *((msg)->len + (uint8_t *)_MAV_PAYLOAD_NON_CONST(msg))
 #define mavlink_ck_b(msg) *(((msg)->len+(uint16_t)1) + (uint8_t *)_MAV_PAYLOAD_NON_CONST(msg))
 
-typedef enum {
```
tridge, line 164: this will break pretty much all users of the library, as most users do define MAVLINK_COMM_NUM_BUFFERS and also use MAVLINK_COMM_0.
The common pattern is (MAVLINK_COMM_0+i)

tridge (changes requested):
I don't really understand what issue this is solving. Are you really trying to use MAVLINK_COMM_7? 
Maybe we should add 
#define MAVLINK_COMM_N(n) mavlink_channel_t(MAVLINK_COMM_0+(n))
or we could add:
#ifndef HAVE_MAVLINK_CHANNEL_T
around the enum define, to allow it to be overridden

julianoes:
@tridge thanks for your answer. You're right it would break existing users that define MAVLINK_COMM_NUM_BUFFERS.

> Maybe we should add
> #define MAVLINK_COMM_N(n) mavlink_channel_t(MAVLINK_COMM_0+(n))
> or we could add:
> #ifndef HAVE_MAVLINK_CHANNEL_T

What about doing both?

julianoes:
@tridge ok, I've added `ifndef HAVE_MAVLINK_CHANNEL_T` as you suggested. That should work, and is basically what @len0rd suggested in https://github.com/ArduPilot/pymavlink/pull/426.

julianoes:
@tridge could you check this again please? Thanks!

hamishwillee:
Hi @tridge 
Can you please re-review?

tridge (approved):

julianoes:
Thanks!

PR #475, opened 2020-11-11
Generator: allow tests to complete despite gcc 9 warnings ...

davidbuzz:
Until the generator can handle these cases better, this allows the test/s to complete with GCC9

error: taking address of packed member of ... may result in an unaligned pointer value.


tridge (approved):

davidbuzz:
is there something i should to to get this merged.?.. its approved.

PR #480, opened 2020-11-26
Add pickle support to DFReader

srmainwaring:
This adds custom pickle support to DFReader_binary and DFReader_text so they can be used with multiprocessing on macOS.

There are three types of object in DFReader that need custom treatment:
 
- struct.Struct - compiled formats used in the unpackers dict
- io.TextIOWrapper - the filehandle
- mmap.mmap - the data_map   

The Struct is provided with a custom pickler that persist its internal format attribute

The filehandle and mmap are treated similarly. Their seek offset is saved just prior to persistence and the attributes are then excluded from the object dictionary. On loading the file is reopened and the filehandle and mmap restored to their previous state using the persisted offsets.

This has had limited testing on macOS using the MAVExplorer MagFitUI module.

### DIFF
--- a/DFReader.py
+++ b/DFReader.py
@@ -12,6 +12,7 @@ from builtins import range
 from builtins import object
 
 import array
+import copyreg
 import math
 import sys
 import os
@@ -49,6 +50,14 @@ FORMAT_TO_STRUCT = {
     "q": ("q", None, long),  # Backward compat
     "Q": ("Q", None, long),  # Backward compat
     }
+
+# make struct.Struct pickleable - needed to pickle DFReader 
+# See: https://bugs.python.org/issue29628
+
+def pickle_struct(s):
+    return struct.Struct, (s.format, )
+
+copyreg.pickle(struct.Struct, pickle_struct)
 
 def u_ord(c):
 	return ord(c) if sys.version_info.major < 3 else c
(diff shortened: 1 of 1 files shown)


DFReader.py
```diff
@@ -1207,6 +1263,42 @@ def last_timestamp(self):
         m = self.recv_msg()
         return m._timestamp
 
+    def __getstate__(self):
```
tridge, line 1266: if would be nice if we could avoid the duplication either by inheriting from a common class, or create a FileHandle class to hold the filehandle that has the pickle hooks on it

tridge (changes requested):

DFReader.py
tridge: a FileHandle class would be nicer, as we could use that for tlogs as well, in mavlogfile() class

srmainwaring: Yes that would work and be better encapsulated as we wouldn't need the additional attributes I added to the DRReader class. We'll need a similar class to provide pickle support for mmap.

Where is the best place to add these classes? I can add them to an existing utils type module or create a new one if preferred.

tridge: hmm, it's tempting to put it in MAVProxy/modules/lib/, and do this:

 - before we fork:  mlog.filehandle = WrapFileHandle(mlog.filehandle)
 - in the child: mlog.filehandle = mlog.filehandle.unwrap()

that keeps all of the pickling and OS dependent code in MAVProxy

srmainwaring: I've made the changes here: https://github.com/ArduPilot/MAVProxy/pull/840. I think the approach is in line with what you have in mind. There are a couple of things at least that will need changing:

- the custom pickle code / classes will need a different home, I've kept them in the MAVExplorer file for now
- I think a lock of some sort is required in the code block where child process is spawned. For a couple of lines mestate.mlog has different types for the filehandle and data_map that will break other threads that attempt to read at that point.

srmainwaring:
Cancelling this PR as superseded by https://github.com/ArduPilot/MAVProxy/pull/840.

PR #485, opened 2020-12-05
Tools: Add mav to .mat conversion

jgooday:
I've noticed a bunch of newbies (including myself) wanting to be able to convert an AP log to a .mat file that can be directly imported into MATLAB. A bunch of us figured out that we could do this by using the "Create MATLAB file" tool in Mission Planner. This is a bad way of doing things.

I've written a basic script, based on mavlogdump.py, that just takes a stream and puts it in a .mat file (in the same format that Mission Planner outputs). Compared to the hacky method above, this has the following advantages:
 - You can **tee up a whole folder** of log files to convert using a simple bash script, rather than having to do each one separately through MP
 - You can **compress** the outputted .mat files using the native MATLAB internal compression. Instead of having 200MB .BIN files, you get 40MB .mat files
 - You can use the **condition statements** you'd use with mavlogdump.py to filter the file based on a log variable - not possible with MP

I noticed there's already a tool for converting a log to a .m file (looks like it just spits out MATLAB code). I don't think this is nearly as efficient (at least in my case).

I haven't done much testing on this script, and there's probably a whole bunch of functionality in there that it doesn't need (from when I copied it over from mavlogdump.py). It may even be worth adding a `matlab` format option in mavlogdump instead - I'd be happy to give that a crack.

Let me know what you think - I definitely don't have the whole picture of pymavlik in my grasp just yet. This script definitely would've helped a bunch of us at the start of the year though!


amilcarlucas:
It would be better if you add a `matlab` format option in mavlogdump instead.

jgooday:
Moved this feature into the `mavlogdump.py` script. Thoughts @amilcarlucas?

tridge (approved):

PR #493, opened 2020-12-24
Tools/ : Added automatic analysis script of telemetry log

konishi-junichi:
Added a script to analyze the telemetry log.
Execution example -> `python mavtlog_analyzer.py flight.tlog`
Currently, only vibration and EKF states are analyzed.
image

### DIFF
--- a/tools/mavtlog_analyzer.py
+++ b/tools/mavtlog_analyzer.py
@@ -1,0 +1,91 @@
+#!/usr/bin/env python
+
+'''
+Automatically analyze flight from telemetry log
+Execution example -> python mavtlog_analyzer.py flight.tlog
+'''
+from __future__ import print_function
+
+from pymavlink import mavutil
+import time
+
+#args
+from argparse import ArgumentParser
+parser = ArgumentParser(description=__doc__)
+parser.add_argument("logs", metavar="LOG", nargs="+")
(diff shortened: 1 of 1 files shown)


peterbarker:
@konishi-junichi have you seen `LogAnalyzer.py` in the ardupilot repository?  It analyzes dataflash logs rather than tlogs, of course, but the concept is the same.  There's a huge amount more data in dataflash logs for doing flight analysis.

There's also `dronekit-la`, a semi-maintained analyzer which takes both dataflash and tlog files.

konishi-junichi:
@peterbarker 
Thank you for your comment.
Certainly, it is best to use the dataflash log for crash analysis.
However, telemetry logs also have the advantage of being easy to get.

Thank you for introducing dronekit-la. The analysis program is substantial and very wonderful!

peterbarker:
On Fri, 25 Dec 2020, konishi junichi wrote:

> Thank you for introducing dronekit-la. The analysis program is substantial and very wonderful!

Contributions over there welcome :-)

Most significant issue with it at the moment is that it doesn't understand 
ArduPilot's "instance" variables - it is still looking for IMU, IMU2 etc 
etc, rather than just "IMU" with an instance field in it.

Peter

PR #494, opened 2020-12-29
use os.path.abspath() for inlcude files

olliw42:
this allows to include dialects from external folders
needed for https://github.com/mavlink/mavlink/pull/1537

### DIFF
--- a/generator/mavgen.py
+++ b/generator/mavgen.py
@@ -93,7 +93,7 @@ def mavgen(opts, args):
             includeadded = False
             for x in xml[:]:
                 for i in x.include:
-                    fname = os.path.join(os.path.dirname(x.filename), i)
+                    fname = os.path.abspath(os.path.join(os.path.dirname(x.filename), i))
                     # Only parse new include files
                     if fname in all_files:
                         continue
@@ -150,7 +150,7 @@ def mavgen(opts, args):
                 #check if all its includes were already done
                 all_includes_done = True
                 for i in x.include:
-                    fname = os.path.join(os.path.dirname(x.filename), i)
+                    fname = os.path.abspath(os.path.join(os.path.dirname(x.filename), i))
                     if fname not in [d.filename for d in done]:
                         all_includes_done = False
                         break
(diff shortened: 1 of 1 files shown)


khancyr (commented):
LGTM
I didn't test on windows thus

olliw42:
I'm on windows ;)

khancyr:
WSL, cygwin or Windows ?

khancyr:
Anyways, I have put it on the list for next devcall, it should be merged monday or before

olliw42:
windows = Windows ;)

tridge:
does this end up embedding the path in any generated files? If it does it would be a problem for projects that pre-generate the headers and then have them on a download site (I think px4 do that??)

olliw42:
> does this end up embedding the path in any generated files? If it does it would be a problem for projects that pre-generate the headers and then have them on a download site (I think px4 do that??)

files are continued to be generated into the specified output folder; one gets in fact exactly the same as if one would have placed the dialect into mavlink/message_definitions/v1.0/. 

E.g., if you have something like this

mavlink/message_definitions/v1.0/
- minimal.xml
- common.xml

mavlink/external/dialects/
- blablabla.xml

and if you have specified 'bbbout' as output folder, then you will get 

bbbout/
- minimal/
- common/
- blablabla/

with the files in them as usual, and in these files you will only see #includes as before. E.g. if blablabla.xml includes common.xml, then in blablabla.h you'll find a #include "../common/common.h", very much like you'll find a #include "../minimal/minimal.h" in common.h.

All stuff goes into one folder, and I thus think it should not affect pre-generation, like what MAVLink does with https://github.com/mavlink/c_library_v2. Btw, their script https://github.com/mavlink/mavlink/blob/master/scripts/update_c_library.sh does not generate external dialects anyway, so there is zero change here. But I think it would also just work fine if it would generate external dialects.

olliw42:
> put it on the list for next devcall

to have said it: frankly, such additions (additions to MAVLink code generators) should be discussed in the MAVLink dev call, and not anywhere else.

auturgy:
> > put it on the list for next devcall
> 
> to have said it: frankly, such additions (additions to MAVLink code generators) should be discussed in the MAVLink dev call, and not anywhere else, right?

Ideally, yes, but at the moment pymavlink and mavlink are managed separately.  The only thing really managed on the mavlink dev call is the xml.  The generators/schema is all done here, and generally discussed during an ArduPilot dev call.  The ArduPilot "Euro" dev call is on at the same time as the mavlink devcall, so I float between both.

tridge (approved):

PR #495, opened 2020-12-31
mavgen: protect for too deep inlcude trees

olliw42:
protect against too deep include trees
useful for https://github.com/mavlink/mavlink/pull/1537 (the include trees tend to become deeper and deeper ...)

### DIFF
--- a/generator/mavgen.py
+++ b/generator/mavgen.py
@@ -115,6 +115,9 @@ def mavgen(opts, args):
         for i in range(MAXIMUM_INCLUDE_FILE_NESTING):
             if not expand_oneiteration():
                 break
+        if i >= MAXIMUM_INCLUDE_FILE_NESTING:
+            print("\nERROR include tree is too deeply nested!")
+            exit(1)
 
         if mavparse.check_duplicates(xml):
             sys.exit(1)


hamishwillee:
@peterbarker This looks like a very good idea to me (eventually we'll hit the limit of inclusion, and this will save us a lot of time explaining to people why their dialect doesn't work)

peterbarker:
Perhaps this would be neater as:
```
count = 0
while True:
    if not expand_one_iteration():
       break;
    count += 1
    if count >= MAXIMUM_INCLUDE_FILE_NESTING:
        print("ERROR include tree is too deeply nested!")
        sys.exit(1)
```

Give-or-take an off-by-one error or two.

My reasoning is just that `for` loops are usually used for bounded iterations, so checking the number of times you iterated looks a little odd.

lightmare:
The original PR won't abort, because `i` will never be set to `MAXIMUM_INCLUDE_FILE_NESTING`. `range(N)` does not contain `N`.
(In other words, `for i in range(N)` is not the same thing as `for(i = 0; i < N; ++i)` ;))

The variant suggested by @peterbarker will work. I'd only note that `while True: ... if something: break` seems awkward, and the comparison should probably be strict, so as to not abort after doing exactly the allowed amount.

```py
count = 0
while expand_one_iteration():
    count += 1
    if count > MAXIMUM_INCLUDE_FILE_NESTING:
        print("ERROR include tree is too deeply nested!")
        sys.exit(1)
```

olliw42:
strange
I certainly had tested it for that occasion (by simply reducing the limit to a small value like 1 or 2 which triggers) and it worked for me (otherwise I wouldn't have dared to offer it). I just double checked again and still found it to work for me.
But I agree it's not so nice for python

the simplest probably would be 

        for n in range(MAXIMUM_INCLUDE_FILE_NESTING+1):
            if n >= MAXIMUM_INCLUDE_FILE_NESTING:
                print("ERROR include tree is too deeply nested!")
                sys.exit(1)
            if not expand_oneiteration():
                break

works for me

lightmare:
Yeah if the range is extended by +1, it will get to the max. If you really want to use a for loop, put the error in its else branch, you don't need to check "are we past max rounds" each time.
```py
for _ in range(MAXIMUM_INCLUDE_FILE_NESTING+1):
    if not expand_oneiteration():
        break
else:
    print("ERROR include tree is too deeply nested!")
    sys.exit(1)
```

But I still think the while loop is more easily understood, especially to people not familiar with nuances of how ranges and for-break-else work in Python.

PR #497, opened 2021-01-09
rotmat: added rotation by enum

tridge:


rotmat.py
```diff
@@ -401,3 +416,59 @@ def plane_intersection(self, plane, forward_only=False):
...
+        self.roll = roll
+        self.pitch = pitch
+        self.yaw = yaw
+        self.r = Matrix3()
+        self.r.from_euler(self.roll, self.pitch, self.yaw)
+        self.rt = self.r.transposed()
+
+# the rotations used in APM
+rotations = [
+    Rotation("ROTATION_NONE",                      0,   0,   0),
```
khancyr, line 431: why are we maintaining our Rotation instead of using MAVLINK one's ? the diff is quite small

tridge: we could switch to that, good future enhancement, though it would need enum description parsing

PR #498, opened 2021-01-15
mavgen_wlua: overall UX improvements for Wireshark dissector

PurkkaKoodari:
This PR contains a bunch of improvements to the UX of the Wireshark dissector (`mavgen_wlua.py`). Details are in commit messages, but here's a summary:

- Command names are now shown in the packet list info column
- Command parameter names and parameters specified as enums are shown in the payload instead of just `paramX`
- Enum types, enum values and flags/bitfield enum values are rendered in a user-friendly way
- Fields are now shown in the order they are specified in the XML, not in wire order
- Message ID field now shows message name and decimal ID (hexadecimals are not used in any docs I could find)
- Magic number field now shows protocol version
- Header and payload subtrees are linked to the respective byte ranges
- Bugfix for an edge case with truncated `float`/`char[]` fields
- Refactoring to keep the code somewhat readable even after these features

There are some pretty significant changes (for this one file), which is why I didn't squash the commits yet. I'm open to changing any of the stuff here, or splitting the PR to multiple ones.

**Before:**
image
**After:**
image
**Before:**
image
image
**After:**
image
image
_(note that I edited my dialect files to use ArduPilot enums for common messages)_

I tested this on Python 3.7 and 2.7 and they produce identical results from `ardupilotmega.xml`. I also tested the dissection with data between Mission Planner and an ArduCopter SITL instance, and didn't find any errors.

The new features don't make many assumptions about the dialect; only the enum `MAV_CMD`, commands `COMMAND_LONG`, `COMMAND_INT`, `COMMAND_ACK` and `COMMAND_CANCEL`, and their parameters `command`, `param1..7`, `x`, `y` and `z` are mentioned by name (not ID), and the generator works just fine even if the dialect does not include them. I think this is a worthwhile tradeoff for getting usable data on commands.

The updated build does produce larger output by about +25% lines and +60% bytes, but I didn't find it significantly slower.


amilcarlucas (approved):
The results look great.

peterbarker:
I've tested locally and squashed the commits together.

Thanks for this!

If you're looking for a "next thing" - currently the disector doesn't understand mavlink extensions.  These fields are simply(?) tacked on after the crc.  You can see this in this decode:
image

Will merge on CI pass.

peterbarker:
Merged, thanks!

PR #502, opened 2021-02-04
.gitignore

khancyr:
Don't track generated file


peterbarker:
I don't have a `generator/generator` - how did you get it?

khancyr:
humm good question ... it looks like it called test_generator.sh from pymavlink directory. Closing

PR #506, opened 2021-02-18
Workaround: Ignore broken MAVLink dialects

sypaq-nexton:
Currently, we can't build the `all.xml` MAVLink dialect because it references files outside of the `message_definitions` folder. As a work-around, ignore this dialect and build everything else.

This is intended to be a temporary work-around until the issue can be addressed properly.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -50,11 +50,6 @@ def generate_content():
             print("No XML message definitions found")
             sys.exit(1)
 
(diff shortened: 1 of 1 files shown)


sypaq-nexton:
Better solution in: https://github.com/ArduPilot/mavlink/pull/197

PR #510, opened 2021-02-25
Mavgen cs rewrite

Hylke-Atmos:
I did a complete rewrite of the CS generator and generated files, because at the time of use (oct.-nov. last year) the CS version was very outdated and seemed unsupported.
However I just saw that the issues we had, have been addressed in december!

Nevertheless I would like to share my solution as an alternative, since I think the current version still has some flaws.
I see the following issues in the current implementation/version:
* **Can get out of sync:** The parser is trying to parse large byte sections at once (based on start marker and length information) but if the frame is out of sync it can (theoretically) take forever before the it gets in-sync again (e.g. data corruption or mid transmission initialization can cause incorrect start marker identification/location, resulting in reading/removing incorrect bytes sections from the stream which as a result of removing too much, can cause the parser to stay in an out of synced state).
* Non MAVLink compliant 8 byte frame in front of the message called "timestamp" or "packettime". It is probably a custom frame, but should (in my opinion) not be part of the community code. If something like this is desired it makes more sense (from Object Oriented point of view) to rewrite it in such a way that you can extend the parser class and override certain sections for customization.

The Implementation I created has a more Object Oriented approach, using typical C# methods (such as delegates and events) also I separated everything into sperate classes for readability.

Unfortunately I've not yet been able to add the signing section yet (I currently don't have a good setup to test this), but I'm planning on adding this in coming weeks/months.

An overview of the changes I've made this far:
* Fires events on each successfully parsed packet. **(brakes interfacing compatibility)**
* Parser based on byte array input instead of Stream (easier serial/ethernet interfacing). **(brakes interfacing compatibility)** but could relatively easily be made compatible with wrapper function!
* Keep track of "incorrectly parsed" bytes incase of a failed or corrupted parse action (see issue mentioned above).
* Add MAVdeprecated object to mavparse.py so that it can be used in the parsers
* Add Obsolete Attribute to field and messages marked as such.
* Only add Units attribute if the units are provided in the XML
* Tried to set it up more Object oriented based.
* Parse enum typed fields as enum instead of primitives. **(brakes type compatibility)**
* Split all Messages to separate files (consistent with other generators!).
* Added significant amount of documentation.

Please let me know if you consider this as a valid approach and enhancement of the current generator/parser.


Hylke-Atmos:
As mentioned in the PR, this is not compatible, but does fix some bugs in the current implementation, also it is (in my opinion) more C# oriented.
Although it might be considered "not done" (signing is missing), I like to know if there is any interest for this rewrite and if so what is required for it to be merged.

@meee1 @peterbarker @amilcarlucas Can you please review, comment on my changes. 

P.S. I'm not sure why it is not building. I don't have these copy conflicts locally (everything is checked-in).

PR #512, opened 2021-03-02
Fixing mavlink message ID schema type.

jblee123:
The XML schema that validates mavlink dialect XML files specified that message
IDs were unsignedShorts, thereby restricting them to 16-bit values. However,
mavlink 2 message IDs are 24-bit values. This commit adds a custom 24-bit
message ID type to the schema to allow for validation of the full 24-bit range
of message IDs.

### DIFF
--- a/generator/mavschema.xsd
+++ b/generator/mavschema.xsd
@@ -11,7 +11,7 @@
 <!-- definition of attributes -->
 <xs:attribute name="name"> <!-- used in enum,entry,message,field elements -->
 </xs:attribute>
-<xs:attribute name="id" type="xs:unsignedShort"/> <!-- used in message elements -->
+<xs:attribute name="id" type="mavlinkMsgId"/> <!-- used in message elements -->
 <xs:attribute name="print_format" type="xs:string"/> <!-- used in field elements -->
 <xs:attribute name="enum" type="xs:string"/> <!-- used in field,param elements -->
 <xs:attribute name="display" type="xs:string"/> <!-- used in field elements -->
@@ -63,6 +63,13 @@
   </xs:simpleType>
 </xs:attribute>
 <xs:attribute name="replaced_by" type="xs:string"/> <!-- used in deprecated elements -->
+
+<!-- mavlink message IDs are unsigned 24-bit values. -->
+<xs:simpleType name="mavlinkMsgId" id="mavlinkMsgId">
+  <xs:restriction base="xs:unsignedInt">
+    <xs:maxInclusive value="16777215" id="mavlinkMsgId.maxInclusive"/>
+  </xs:restriction>
+</xs:simpleType>
 
 <xs:simpleType name="SI_Unit">
   <xs:restriction base="xs:string">


mday299:
@peterbarker Can you take a look at this PR sometime soon?

tridge:
This looks good to me, it just needs some testing. SITL testing with ardupilot after generating with this change would be sufficient

jblee123:
Testing on our end or independent testing on your end? And if on our end, do you need any specific supporting data?

The testing I did involved a custom mavlink dialect we use for over-the-air comms and verified that: 1) the OTA comms still worked, and 2) the validation rejected ID values greater than (2^24-1) but allowed lesser values (i.e. accepts the range it's supposed to). Note that this does not affect code generation in any way; it only fixes an XML-validation bug.

peterbarker:
I'm curious to know "why now?".

The existing address space is rather large.

Yes, its a wart that we're using a 16-bit space rather than allowing for the full 24 bits - but (at the risk of going down in history), isn't 65 thousand message types enough for anybody?

One drawback here is that now you can specify a 32-bit quantity and have it stored, and we're relying on a secondary check to ensure the 24-bit compliance.

I'd previously mused that since nobody could've used these extra 8 bits that perhaps they could be repurposed.  Making this change would preclude that.

So (at least one) one drawback, no gain?

jblee123:
In our case, we started out with a custom dialect that extended the common mavlink messages by including common.xml back in the mavlink 1 days, when IDs were at a premium, and even had to stay on earlier versions of common.xml since new messages were added that encroached on the IDs we used. After the move to mavlink2, we renumbered everything to stay away from the common IDs. So how far up to go? I saw new messages in common.xml that were in the 400's, and then jumped to 9000, and then jumped to the 12900 range. With jumps like that I was concerned about how fast the greatest "common" message ID would increase, so I picked 1,000,000, because that should be legal, and I figured it'd be a while before common.xml messages got up that high. It's not always about the number of values, but the ranges they cover. They aren't "defragged", so to speak.

That's when I saw the schema bug. pymavgen complained when trying to generate our mavlink messages. As a workaround, I ended up, during the build, programmatically "fixing" the schema (and then "unfixing" it after generation so it didn't show up as a change in source control). It was either that or turn off schema verification, which I didn't want to do. I finally decided the best thing to do would be to simply get the bug fixed.

You mentioned that a drawback is that now you can specify a 32-bit value and have it stored, but that's not correct. A generated mavlink2 message only allocates 3 bytes of memory in the message for the ID, which will be true no matter what the schema says.

As far as keeping people from using those 8 bits so they can be repurposed, I don't think that's feasible, either. People *can* use those 8 bits now -- they just have to turn off testing against the schema, which is a supported command option, IIRC.

Hopefully that addresses your concern of any drawbacks. As far as gains, I'd argue that the fix makes the schema correct, whereas it's currently incorrect. It lets people use IDs that mavlink says they can use. That sounds like a gain to me.

peterbarker:
Thanks @jblee123 - we'll discuss at a DevCall.

peterbarker (approved):
Tested generation of headers with ardupilot build; identical to master, both p2 and p3

peterbarker:
Merged, thanks!

PR #519, opened 2021-03-22
Tools: mavlogdump: do not rearrange .mat before exporting

IamPete1:
This is a change to directly output .mat files rather than rearranging first. 

Currently you get lots of cells and labels:
Capture

Capture1

With this change you get friendly structures with correctly named values:
Capture2

Capture3

This results in a format that is different to what you get with MP, but it is also much nicer. 

I guess we might want to make it optional if anyone is particularly attached to the old format.

### DIFF
--- a/tools/mavlogdump.py
+++ b/tools/mavlogdump.py
@@ -347,22 +347,7 @@ while True:
 
 # Export the .mat file
 if args.format == 'mat':
-    # Rearrange the dictionary so it exports correctly
-    MAT2 = {}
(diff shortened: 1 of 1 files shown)


IamPete1:
@jgooday any thoughts?

jgooday:
That looks much better to me @IamPete1. No performance difference between structs and tables? (I expect they're both pretty slow)

tridge (approved):

PR #526, opened 2021-04-05
Add invalid attribute to mavschema

olliw42:
title says it

required for https://github.com/mavlink/mavlink/pull/1614


auturgy:
Thanks Olli

olliw42:
my pleasure
hope it's ok (never touched xsd before LOL)

tridge (approved):

hamishwillee:
@olliw42 Thanks, this is excellent. 

There does need to be some follow on work though because once we add the invalid attributes we'll pull this information from the description. At that point it needs to be added back into the generated documentation. That should be done minimally for Python and C, and I would hope we could add an issue to get some help to do it for the other generators.

FYI @tridge

PR #530, opened 2021-04-09
generator: fix bug in Node/javascript exposed by a new 'events interface' pr

davidbuzz:
https://github.com/mavlink/mavlink/pull/1531
The 'arguments' object , has special meaning in javascript, as it contains the values of the arguments passed to that function.... re-worked the generator to not use it, and use instead the ES6 compatible Rest parameters method.


peterbarker:
Merged, thanks.

PR #536, opened 2021-04-29
Add script in description

brunoolivieri:
Add @khancyr end-2-end script.

Thanks for the example, it is very good.


peterbarker:
@brunoolivieri that example doesn't seem to exist.  I don't think the PR has been merged yet (doesn't pass CI last time I looked).

@khancyr

PR #544, opened 2021-05-20
Fix up enum generation

### LINKED ISSUE #538: Enum parsing is a little broken
So the way this works is that when you do [merge_enums()](https://github.com/ArduPilot/pymavlink/blob/master/generator/mavparse.py#L465-L495) the `x.enums` for each xml file gets modified. If the file is the first to have an enum it gets to keep it and all the duplicate enum values are merged in. If the enum is "new" then it gets added to a list of new enums that get assigned back to the xml file.

The generator then spits out the enums in each file at the end. The parent has all the merged enums and #includes the dialect files to get their unique enums. 

What this means is that the top level file works fine - it has access to all enums. However if you were to include mavlink.h from a dialect you would find if missing any enums that were declared in the parent, including any values that were in the XML. This is different from other things like messages, where we use #defines to make sure that at each level you get what is underneath. What it also means is that if you generate multiple dialect files the order matters.

Does this make sense - do we all agree it is a bug?

If so, any thoughts on fixing?

hamishwillee:
This is very crappy "first draft" for #538. Mostly for me to verify my assumptions that enum generation is broken, and verify that it can be fixed (i.e. the concept is not fundamentally broken).

The old version of the code worked by working through each enum creating an "enum map". After parsing a file the new enums (with respect to the map) were assigned back to the file. What this meant is that at the top level you have a definition that is correctly reflects all included files. However following files e.g. common are left with only the enum entries that are _unique_ to them. So if you build ardupilot, the generated common.xml does not get the MAV_CMDS - they are stripped out. 
This is contrary to the intended design - you should be able to build ardupilot.xml and use the library for common.xml. It also means that the order you build files in matters. 

This version does a far more predictable merge. Essentially it merges bottom-up from included files first - and it only merges if it needs to (i.e. if both of the files to be considered actually have enums in them, and then, only if for the enums that match). It does not remove any enums defined at a particular level, it just ensures that any new entries from the lower level are appended.

Why is it crap? I included everything in mavgenerate.py. I think that the merging and sorting code should be part of the enum object and called from mavgenerate. There are some hacks in there because it can't be. Further I do not check (at least deliberately) for duplicates. I don't think there are any but we should actually replace/overwrite them or fail and report this (which?).
Further I don't do any reordering as a result of autogeneration of ids. Why? Because if we do that then the order of inclusion must matter - it does not now but that is just dumb luck because there is actually only a small about of merging going on -mostly in MAV_CMD. We should disallow that.

What has it shown? I have attached a zip of the libraries built with the current mavlink/mavlink generator and with my version. This allows you to do a comparison between the libraries as we have now  - which demonstrate how broken and unpredictable things are now - and that the order matters. For example, compare ardupilotmega.xml with the one built using all.xml.

It also allows you to compare the new libraries to each other and to the old libraries. This pretty much shows (at least on first scan) that the new tools generate consistent libraries for the dialects irrespective of the order they are called. They are also the same libraries for the top level as previously - so old and new tools build the same ardupilotmega.xml. The difference is that the new tools build a correct common.xml. 

Does anyone know why we have to have that end marker. It technically does break consistency - a system that understands ardupilot will have a different understanding of that enum value than one that understands common. I'm not sure that this really matters, but it is an inconsistency.

[comparison.zip](https://github.com/ArduPilot/pymavlink/files/6514393/comparison.zip)

@peterbarker @julianoes Before I tidy this code up, can you please look at the output zip and confirm that my approach is sound (ie going bottom up) - ie that the result of this does "the right thing" and I'm not fooling myself. No huge urgency, but 

@auturgy FYI. 

Fixes #538.


olliw42:
@hamishwillee you may want to look at fastmavlink how i solved it there, i think conceptually more correctly, should not be so difficult to translate to pymavlink

hamishwillee:
@olliw42 I basically just hooked into your [expand_includes()](https://github.com/ArduPilot/pymavlink/pull/544/files#diff-771a8f0d763a6e2675773e6358708f7791964f8ccd4fe869056361d674b78ba2R255) then call merge on every include. At the end clean up all the enums to add the end value. I'm fine with the algorithm - just need to move the location of the merge code to the enum itself as that makes more sense. 

I'll have a look at the code thanks. In all likelihood though it is likely to take me more time to work out what you're doing and  pulling it into mavgen than I have free for this. Happy to take a PR from you though.

Mostly I just want @peterbarker to look at the PR and confirm that the end result is correct.

peterbarker:
I'm a bit confused here.

```
pbarker@bluebottle:~$ grep --after-context=10 '# MAV_CMD$' /tmp/master.py 
# MAV_CMD
enums['MAV_CMD'] = {}
MAV_CMD_NAV_WAYPOINT = 16 # Navigate to waypoint.
enums['MAV_CMD'][16] = EnumEntry('MAV_CMD_NAV_WAYPOINT', '''Navigate to waypoint.''')
enums['MAV_CMD'][16].param[1] = '''Hold time. (ignored by fixed wing, time to stay at waypoint for rotary wing)'''
enums['MAV_CMD'][16].param[2] = '''Acceptance radius (if the sphere with this radius is hit, the waypoint counts as reached)'''
enums['MAV_CMD'][16].param[3] = '''0 to pass through the WP, if > 0 radius to pass by WP. Positive value for clockwise orbit, negative value for counter-clockwise orbit. Allows trajectory control.'''
enums['MAV_CMD'][16].param[4] = '''Desired yaw angle at waypoint (rotary wing). NaN to use the current system yaw heading mode (e.g. yaw towards next waypoint, yaw to home, etc.).'''
enums['MAV_CMD'][16].param[5] = '''Latitude'''
enums['MAV_CMD'][16].param[6] = '''Longitude'''
enums['MAV_CMD'][16].param[7] = '''Altitude'''
```

```
pbarker@bluebottle:~$ grep --after-context=10 '# MAV_CMD$' /tmp/hamish.py 
# MAV_CMD
enums['MAV_CMD'] = {}
MAV_CMD_NAV_WAYPOINT = 16 # Navigate to waypoint.
enums['MAV_CMD'][16] = EnumEntry('MAV_CMD_NAV_WAYPOINT', '''Navigate to waypoint.''')
enums['MAV_CMD'][16].param[1] = '''Hold time. (ignored by fixed wing, time to stay at waypoint for rotary wing)'''
enums['MAV_CMD'][16].param[2] = '''Acceptance radius (if the sphere with this radius is hit, the waypoint counts as reached)'''
enums['MAV_CMD'][16].param[3] = '''0 to pass through the WP, if > 0 radius to pass by WP. Positive value for clockwise orbit, negative value for counter-clockwise orbit. Allows trajectory control.'''
enums['MAV_CMD'][16].param[4] = '''Desired yaw angle at waypoint (rotary wing). NaN to use the current system yaw heading mode (e.g. yaw towards next waypoint, yaw to home, etc.).'''
enums['MAV_CMD'][16].param[5] = '''Latitude'''
enums['MAV_CMD'][16].param[6] = '''Longitude'''
enums['MAV_CMD'][16].param[7] = '''Altitude'''
--
# MAV_CMD
enums['MAV_CMD'] = {}
MAV_CMD_NAV_WAYPOINT = 16 # Navigate to waypoint.
enums['MAV_CMD'][16] = EnumEntry('MAV_CMD_NAV_WAYPOINT', '''Navigate to waypoint.''')
enums['MAV_CMD'][16].param[1] = '''Hold time. (ignored by fixed wing, time to stay at waypoint for rotary wing)'''
enums['MAV_CMD'][16].param[2] = '''Acceptance radius (if the sphere with this radius is hit, the waypoint counts as reached)'''
enums['MAV_CMD'][16].param[3] = '''0 to pass through the WP, if > 0 radius to pass by WP. Positive value for clockwise orbit, negative value for counter-clockwise orbit. Allows trajectory control.'''
enums['MAV_CMD'][16].param[4] = '''Desired yaw angle at waypoint (rotary wing). NaN to use the current system yaw heading mode (e.g. yaw towards next waypoint, yaw to home, etc.).'''
enums['MAV_CMD'][16].param[5] = '''Latitude'''
enums['MAV_CMD'][16].param[6] = '''Longitude'''
enums['MAV_CMD'][16].param[7] = '''Altitude'''
pbarker@bluebottle:~$ 
```

We seem to have the same definitions twice?

hamishwillee:
You'll have the definitions in every file "above" where the MAV_CMD was defined in the dialect chain. But they should be guarded so that they are imported only once, in the top level file. Is that not what you're seeing?

I.e. is that grep of one file, or showing the definitions across multiple files?

peterbarker:
On Thu, 17 Jun 2021, Hamish Willee wrote:

> I.e. is that grep of one file, or showing the definitions across multiple files?

One file

hamishwillee:
This is Python right? I ask because I tested on C. 

Of course Python has to work too and I will look at this.

hamishwillee:
@peterbarker Fixed for python. Identical-ish output now before and after: [py_compared.zip](https://github.com/ArduPilot/pymavlink/files/6699140/py_compared.zip)

In summary, the original parser was flawed because it stripped out merged enums from the nested dialect objects. My fix for C added the appropriate merged enums for each level - which means that you get the correct set for whatever dialect you include.

Unfortunately the python parser simply appended all the enums from all the files it included - that worked before when the duplicates had already been stripped, but resulted in duplicates after the fix for C. In other words, pymavlink worked because the parser was broken.

What this does now is re-run merging the enums in the Python parser. All the duplicates get chucked out.

I haven't checked how the other parsers handle things yet. Anything that does the same thing as Python will be broken too though.

hamishwillee:
So quick scan indicates that the same problem probably affects JavaScript at least. I wonder if we might be best off doing a "total" enums calculation in the top level and passing it to all the generators so they don't have to duplicate this code. Thoughts?

[JavaScript_NextGen.zip](https://github.com/ArduPilot/pymavlink/files/6699450/JavaScript_NextGen.zip)

hamishwillee:
@peterbarker Discussed this in the dev call with @auturgy. He raised the point, (which we are already aware of) that the fix to make C work cascades to all generators. 

I made the point that I did this PR to verify the defective behaviour and test our appetite for "doing the right thing". I'm not the greatest coder and I would much prefer that someone else (you ideally) took this over. If not, then I am happy to progress this but would prefer to do so with your review of the architecture or some roll out plan.

We could for example create a pre-merged enum in the top level of mavgen and pass that to the generators like pymavlink to use instead of merging the individual files themselves. We'd have to update them all, but the code would be simpler.

Or we could make the changes I did for C-only. The other libraries would remain the same and also be strictly  compatible "at the top level". The costs would be that a) anything constructed using a nested structure like C would still be wrong for the inner nesting levels, b) inconsistent implementation and a little confusing for anyone as to how they implement a new parser.

Anyway, have a chat to James and let me know how you want to progress this.

davidbuzz:
As the author of large parts of JavaScript_NextGen, i'm here and paying attention, if i'm needed.

hamishwillee:
That's great @davidbuzz . 

I think we could modify this slightly to only impact C and Python now, allowing other runtimes to modify themselves at will.

peterbarker:
So it does look like this solution will cause issues in the javascript bindings and in the c-sharp bindings.

Is this just a C/C++-11 generation issue?  

I'm wondering if we should actually stop generating any of this as multiple include files - so if you generate from ardupilotmega.xml you only get an includefile for `ardupilotmega.xml` - to get a `common.h` you'd need to generate from `common.xml`.

People probably won't bite on that one....

While I do see the problem here, and do agree that it needs fixing, we can't break random language bindings that people are happily using (or we get bug reports that have to be chased in short order)).

I'll keep looking at this.

davidbuzz:
I have reviewed the output from this pr with respect to how it affects the current JavaScript  bindings, and it looks good to me.  ( by example... it leaves the output derived from common.xml as-is, but it increases the size of the output of 'ardupilotmega.xml and/or 'all.xml' by the appropriate number of formerly missing enums.   looks good to me.

hamishwillee:
This PR should work properly (now) for anything that builds per-dialect files, and break anything that generates a single flat library (other than Python which I specifically fixed). This will show up in the flat libraries as duplicate entries.

I think you're right that it is very risky to accept this, without knowing which libraries are affected.

We could however make this change restricted to C and Python and then build it out to other things following more precise testing.

> People probably won't bite on that one....

IMO both the parser and the generators are much more complex than they need to be because the assumption is that you build all the dialects. I am not sure whether there are any particular benefits to resulting code either.

Upshot you're probably right, but it would not be me biting. I am awestruck by the cleverness of the code which I don't understand the justification for.

hamishwillee:
Pretty please review @peterbarker

tstastny:
@hamishwillee we just ran into the effects of not having this. @peterbarker can we move this forward?

hamishwillee:
@tstastny FMI "where"/"how" did you run into this? Provided you build the top level dialect and use the top level dialect there are no issues. The only problem is when you build a dialect but then try to access the commands through a lower level generated library will you run into issues.

Right now https://github.com/mavlink/c_library_v2/ _should_ be ordered such that if you take that library the you can use any level. 

I have been nagging about this for years. @peterbarker What's the current story?

tstastny:
@hamishwillee the issue comes when a dialect in the chain sets an enum. that means that only in the dialect header the enums are defined, so if one of the other generated headers (e.g. development) is included in some library (e.g. as in PX4 SITL..), but mavlink is built with a different dialect on the top level project (PX4), then it wont find the enums.

So in that case the only solution is to set the dialect as the include path for also the potential submodule lib (is the case in PX4 SITL). But that means a bunch of forks.

What would be better is if you can include any of the generated headers and still have access to the enums. Which is what I understood this PR to do.. but I could be misunderstanding.

I'm probably generally very poorly explaining this. @bkueng could you help add context?

hamishwillee:
@tstastny So what is supposed to happen is that if you build a top level dialect - say ardupilotmega.xml, you can include ardupilotmega/mavlink.h and get the enums and entities for common.xml, standard.xml, minimal.xml. If you wanted to, from that same build, you _should_ also be able to include common/mavlink.h and get everything that was defined in standard.xml, minimal.xml (but not ardupilotmega.xml) and so on.

However what happens is that if you build ardupilotmega.xml the _merged_ enums, which include MAV_CMDs only appear in ardupilotmega/mavlink.h. If you import common/mavlink.h and ardupilotmega.xml has any MAV_CMDs then those are gone - they appear in the higher mavlink. You still get any enums that aren't merged.

If you want to use the libraries now, you should just take the top level one that has all the definitions you want and include the mavlink.h for that library. If you want to use a lower level library, rebuild for that xml and use that version. 

The prebuilt library https://github.com/mavlink/c_library_v2/  SHOULD work correctly because it builds the high level libraries first - so they get all the enums - and then builds the lower level libraries (overlaying) so that each level has the correct enums, and so on. Testing indicates this works, but it may be that my testing wasn't good enough. Integrating this bug fix would mean that it would all work, but as I say, good to understand why it is needed (I don't from your explanation).

tstastny:
@hamishwillee correct on all that you say. so the "why" in this case is particularly:
- PX4 built with a dialect will handle using that dialect on all includes within PX4 repo itself
- PX4_SITL is a submodule which specifically includes development xml
- Since currently the enums will only appear in the higher dialect header, SITL wont work due to not having it's development enums, even though they do in principle still exist (just not in development header)

My contention is that it should not be necessary to go now down into that submodule and change to the high level dialect to make everything play nice together. But that as you say - the low-up-to-defined header merged enums should be defined in all headers, such that if SITL includes development.. it still gets all enums development and down, despite the fact that PX4 itself was built with a higher level mavlink dialect which may or may not add additional enums.

I believe this is also what you are saying.. but correct me if I'm wrong

hamishwillee:
> @hamishwillee correct on all that you say. so the "why" in this case is particularly:
> 
> * PX4 built with a dialect will handle using that dialect on all includes within PX4 repo itself

If it includes `dialect/mavlink.h` 

> * PX4_SITL is a submodule which specifically includes development/mavlink.h

You're saying that you build MAVLink once for PX4 using some dialect. If that happens to be development.xml then everything works. However if you build it with all.xml then PX4_SITL will miss the headers. Right?

Shouldn't PX4 SITL use the version of MAVlink you're building with?

> * Since currently the enums will only appear in the higher dialect header, SITL wont work due to not having it's development enums, even though they do in principle still exist (just not in development header)

Correct. Though to be very precise, it will be missing any enums merged with the higher layer, such as MAV_CMDs. It will get to keep non-merged enums.

> My contention is that it should not be necessary to go now down into that submodule and change to the high level dialect to make everything play nice together. 

I agree. Now we have to convince @peterbarker. This fixes pymavlink and c/c++ libraries but the other libraries might still not work. I think it is worth making them break so that they get fixed, or we might make the changes conditional on particular language. Or we might throw this all away, build the library into a single monolith, and stop pretending that the libraries can be used as we expect.

auturgy:
ping @peterbarker

peterbarker:
The more I'm thinking about this the more I'm leaning towards, "we don't generate intermediate headers, we just generate header files for a single XML at a time, and the `mavlink.h` generated 

This is what we do for Python; if you generate for ardupilotmega.py you do *not* get "common.py" in addition to ardupilotmega.py.

In the suggested case of including the development headers in PX4's SITL implementation, an XML could be created which `<include>`s both `common.xml` and `development.xml` (notionally `px4_sitl.xml`)

An example of where allowing multiple includes from the one source falls down:

```
              A
            /   \
          B      C
            \    /
              D
````

If you generate from A  but don't *actually* want A's message definitions, then you might try to include the header files B.h and C.h.  That's not going to work as only one of them can define the enumerations (or the `MAVLINK_MESSAGE_CRC`s, or any one of a number of other things, I expect!)

@hamishwillee points out that the update_c_library.sh workflow which generates the pre-built headers already generates for all dialects (https://github.com/mavlink/mavlink/blob/master/scripts/update_c_library.sh#L63)

If we make this change then we will break people taking the advice generated into the include files (https://github.com/ArduPilot/pymavlink/blob//generator/mavgen_c.py#L89) about how to construct their own include ordering, but as described above the approach is flawed anyway, you can't just mix-and-match headers and get what you hoped for.

**Comment by olliw42** [comment]
I got a ping since I'm mentioned in the above

as a general comment I wonder why you are not trying to take advantage of the benefits which trying to learn from others may give. I claim fastmavlink has solved that completely.

The mentioned case A-B/C-D is no issue really, and the fact it is considered an issue just emphasizes the point. It's easily handable by #ifndef #define like with .h includes in the C code. This allows it that only the top-level cmd enum and crc table is included.

**Comment by hamishwillee** [comment]
@peterbarker As per @olliw42 comment allowing multiple includes from the one source works fine. 

The levels are build in most-deeply nested order first, so they are only built once. When B is built it is created with mavlink.h and B.h, where B.h contains the specific combined MAVLINK_MESSAGE_CRCs. Selecting the b/mavlink.h enables the right defines for that level.

> @hamishwillee points out that the update_c_library.sh workflow which generates the pre-built headers already generates for all dialects (https://github.com/mavlink/mavlink/blob/master/scripts/update_c_library.sh#L63)

It only has to do that because of the bug that makes this PR necessary. If this was merged everything would work and we could just build all.xml and any files that are not linked by all.xml.

Because of this bug I need to build each dialect in top down order so that the broken intermediate levels are built correctly. That works but is finiky.

The code should work it is just ugly.

File: generator/mavgen.py [hunk 1]
```diff
@@ -182,6 +221,9 @@ def update_oneiteration():
                         x.message_target_component_ofs.update(ix.message_target_component_ofs)
                         x.message_names.update(ix.message_names)
                         x.largest_payload = max(x.largest_payload, ix.largest_payload)
+                        if x.enum is not None and ix.enum is not None:
```
hamishwillee, line 224: This probably should be as shown - the enums are always defined as list, so they won't ever be `None`. We should merge if both of these are non-zero-length lists, which think equates to:

```suggestion
                        if x.enum and ix.enum:
```

generator/mavgen.py
```diff
@@ -182,6 +221,9 @@ def update_oneiteration():
...
+                            #print("merge %s into %s" % (ix.filename, x.filename) )
+                            merge_enums(x.enum, ix.enum) # Update duplicate x.enums with included values
```
hamishwillee, line 226: @peterbarker Following my investigation of https://github.com/mavlink/qgroundcontrol/pull/12410 I don't think this PR fully solves all issues with MAV_CMDs (though I think it is on the right track)

I think the current structure even without this PR makes it impossible to build or prebuild `all.xml` "reliably" to get commands from both development and ardupilotmega - the design is not robust for this kind of multiple inclusion.

----

Explanation

Say you build dialect.xml, you end up with dialect.h and mavlink.h that loads it. Dialect.h has definitions for all the enums defined within dialect.xml, such as this one for mav_commands.

```c
#ifndef HAVE_ENUM_MAV_CMD
#define HAVE_ENUM_MAV_CMD
typedef enum MAV_CMD
{
...

Dialect.h then includes header files for its dialects. However if it has defined any commands then `HAVE_ENUM_MAV_CMD` is already true, and none of the included commands get imported. This is why we need to make sure that the top level dialect merges MAV_CMDs from its lower level dialects.

Dialect.xml/.h does not define any commands then `HAVE_ENUM_MAV_CMD` is not defined, and the commands that get imported are those in the first included file that has `HAVE_ENUM_MAV_CMD` defined.

If you look at `all.h`, this does not define any commands, so the ones that get imported will depend on the order of include. As you can see, ardupilotmega.h, not development.h.

// base include
#include "../ardupilotmega/ardupilotmega.h"
#include "../ASLUAV/ASLUAV.h"
#include "../common/common.h"
#include "../development/development.h"
#include "../icarous/icarous.h"
#include "../minimal/minimal.h"
#include "../python_array_test/python_array_test.h"
#include "../standard/standard.h"
#include "../test/test.h"
#include "../ualberta/ualberta.h"
#include "../uAvionix/uAvionix.h"
#include "../storm32/storm32.h"
#include "../AVSSUAS/AVSSUAS.h"
#include "../cubepilot/cubepilot.h"
#include "../csAirLink/csAirLink.h"
```

So what should happen here?

I think this means that an intermediate dialect needs to have all its "inherited" definitions defined in a  `HAVE_ENUM_MAV_CMD` because there is no way to be sure otherwise that you won't fall though to an arbitrary other definition, depending on how the tree is defined.

This in turn means that we need to copy the definitions into the intermediate files for any enum that is defined more than once.

[Further, if we have a level that has no definitions there is no merging -  the next level will not know about the lower levels below that which were "missed" with the code here.]

Lastly, for this case all.xml should merge the enum definitions from development.xml and ardupilotmega.h, treating the first defined as the "master" if there are duplicates.

It is pretty easy to do this by modifying the code here to
```
                        if ix.enum and not x.enum:
                            x.enum = ix.enum
                        elif x.enum and not ix.enum:
                             pass
                        elif x.enum or ix.enum:
                            print(f"merge {ix.filename} into {x.filename}")
                            merge_enums(x.enum, ix.enum) # Update duplicate x.enums with included values
                        break
```
But the result is ugly as hell, because the top level ends up with all the enums of the sub levels, whether or not they are needed. 

We could improve this with care, but I think its going to end up looking not that different from the "single header file" library idea.

hamishwillee: PS We could hack this now by defining a dummy mav_cmd in all.xml

julianoes:
Unless I'm misunderstanding something, I think I'm on @peterbarker's camp of:

> "we don't generate intermediate headers, we just generate header files for a single XML at a time, and the mavlink.h generated

I don't think there is a way to generate a `dialect/mavlink.h` which includes `another_dialect.h` and then also include `dialect/another_dialect.h`, that's just not how enums work in C.

And as @hamishwillee says you could do it with ifdefs but it it gets ugly. In that case, what's wrong with disallowing this use case and just use the header that was specifically generated for a use case? I must be missing something :thinking:

hamishwillee:
> Ah that's what you wrote above @hamishwillee, just catching up 😄.

Yeah.

So this is "easily fixed" by each level copying any enums of the headers it includes (as shown in my last post). That ensures that each dialog parser object has information about all the enums that belong to it, rather than assuming that they will be picked up when we include the lower level header.

I suspect that MAV_CMDs are the only thing for which this has any "in real life" impact, because they are the only case where we usually have to merge enum values, so perhaps it is not worth fixing.
But the whole thing is a bit of a house of cards.

I tend to agree that we should rewrite the C implementation to be a single dialect level. But you would still have to do all these changes to make it work :-).

synther:
Hi, I encountered this issue, tried to debug it myself, and eventually came across this PR. I truly appreciate @hamishwillee’s dedication to this problem over the past four years!

Here I'd like to add that C++ (not C) generation is also affected.

----

Initially, it has the same `MAV_CMD` disappearance issue as in the C generator. It also depends on the order in which dialects are generated.

For instance, `ardupilotmega` ends up without `MAV_CMD`, while `storm32` does have it. It happens because `storm32` [includes](https://github.com/mavlink/mavlink/blob/677b9e3e25dbb1c606f87fa30b293a99c5a6f313/message_definitions/v1.0/storm32.xml#L20) `ardupilotmega`. And `storm32` is generated after `ardupilotmega` [alphabetically](https://github.com/mavlink/mavlink-gbp-release/blob/d30a86c5c370c002c7ea01bd5ce3fa2bfe861d64/patch/CMakeLists.txt#L124), effectively overriding the previously generated `ardupilotmega` headers.

image

So if you try to use `MAV_CMD` from `ardupilotmega`, you'll get a compile time error.

Interestingly, `common`'s `MAV_CMD` is unaffected, even it's included by nearly every other dialect. Its entries don't get overridden. That’s because its generation is [deliberately delayed until the end](https://github.com/mavlink/mavlink-gbp-release/blob/d30a86c5c370c002c7ea01bd5ce3fa2bfe861d64/patch/CMakeLists.txt#L119) as a workaround. This is the same strategy used for the pure C generator ([source](https://github.com/mavlink/mavlink/blob/677b9e3e25dbb1c606f87fa30b293a99c5a6f313/scripts/update_c_library.sh#L63)), but applied only to `common` for some reason.

----

C++ mavlink generator places everything in a namespace. When using a mavlink enum or message, you must specify the dialect explicitly ([example](https://github.com/mavlink/mavros/blob/aa534f1c57e2c6146372f8444fe17416e040c022/mavros_extras/src/plugins/mount_control.cpp#L362)):

```
using mavlink::common::MAV_CMD;
```

As an example, take `DO_SET_RESUME_REPEAT_DIST` MAV_CMD which is [defined in `ardupilotmega` specifically](https://github.com/mavlink/mavlink/blob/677b9e3e25dbb1c606f87fa30b293a99c5a6f313/message_definitions/v1.0/ardupilotmega.xml#L31).  However, because of the bug, it ends up in the `storm32` namespace instead.

image

As a result, you’re forced to write `using mavlink::storm32::MAV_CMD` to use a command that belongs to ardupilotmega, which is clearly counterintuitive.

----

I believe that once the C part is fixed properly, the C++ generator will inherit the fix automatically. This post is simply a reminder to verify the C++ generation as well.

generator/mavgen_c.py
```diff
@@ -747,6 +747,21 @@ def generate_one(basename, xml):
...
+    keepEnums = []
+    for e in xml.enum:
+        if not e.copied:
+            keepEnums.append(e) # keep: defined in this file
+        elif e.merged:
+            keepEnums.append(e) #keep: not defined in file but is merged (might be dropped)
+        else:
+            # Drop. Only defined in one file so will be included.
+            pass
+    xml.enum = keepEnums
```
hamishwillee, line 761: Each level xml has its own list of enums that correctly indicates whether it is merged or copied at that level (see https://github.com/ArduPilot/pymavlink/pull/544/files#r2101772580). 

For a C like language we could just include all of the enums in the top level, since they are `#defined` by name, and so you'd get the ones defined in the top level file.

But because this is C we don't have to do that. We include the enums in each level that were defined in that level. We also include those that were copied from another level and merged. The reason being that the merging could have been across parallel dialects - which is why all wasn't picking up both ardupilot and development enums.

Upshot, this code keeps the enums we need at each level. 

Comparison of the files shows this works perfectly for `C` build - all works without the dummy value.
This also proves that currently the MAV_CMD is the only risky enum.

generator/mavgen_python.py
```diff
@@ -11,8 +11,11 @@
 
 import os
 import sys
+import operator
 import textwrap
 from . import mavtemplate
+from . import mavparse
```
hamishwillee, line 17: Don't know how these changes got here. I don't think needed, but need to retest.

generator/mavgen_python.py
```diff
@@ -1135,11 +1138,10 @@ def generate(basename, xml):
         filename = basename + ".py"
 
     msgs = []
-    enums = []
+    enums = xml[0].enum #All enums merged in top level
```
hamishwillee, line 1141: So previously Python built up all enums by extending the enums at each level.
This essentially seemed to work.

But now that the top level has all the enums we can just include that from the top level file we are building. Shown to work.

generator/mavgen.py
```diff
@@ -82,6 +83,100 @@ def mavgen(opts, args):
...
+                    print("Merging enum: %s" % an_enum.name)
+                    an_enum.entry.extend(inc_enum.entry)
+                    # sort by value
+                    an_enum.entry = sorted(an_enum.entry,
+                                   key=operator.attrgetter('value'),
+                                   reverse=False)
+                    # Add description but only it has not been defined at top level.
+                    if not an_enum.description:
+                        an_enum.description = inc_enum.description
+        """
```
hamishwillee, line 157:
```suggestion
```

generator/mavgen.py
```diff
@@ -82,6 +83,85 @@ def mavgen(opts, args):
...
+                #print(f"Adding length enum: {len(only_inherited_new)}")
+                if len(only_inherited_new) > 0: # Only merge things from lower level
+                    #print("Merging enum: %s" % inc_enum.name)
+                    an_enum.merged = True
+                    an_enum.entry.extend(only_inherited_new)
+
+                    # sort by value
+                    an_enum.entry = sorted(an_enum.entry,
+                                       key=operator.attrgetter('value'),
+                                       reverse=False)
```
hamishwillee, line 133: Need to think about this bit a bit more.

This marks an enum as merged if any items are copied into it from a lower level.  That's cool, but the enum object will be the same the one in the lower list, so this will set the object in the lower list as merged, even if it isn't.

In practise this won't break anything, and with our current setup wouldn't change anything.
But we should be robust.

PR #545, opened 2021-05-28
Don't hardcode sysid and compid when packing

JonasVautherin:
Maybe I am missing something, but that surprised me. Typically this means that doing `message.unpack(); message.pack()` changes the sysid/compid, and I don't really see a reason for it :thinking:

### DIFF
--- a/generator/mavgen_java.py
+++ b/generator/mavgen_java.py
@@ -175,8 +175,8 @@ public class msg_${name_lower} extends MAVLinkMessage {
     @Override
     public MAVLinkPacket pack() {
         MAVLinkPacket packet = new MAVLinkPacket(MAVLINK_MSG_LENGTH,isMavlink2);
-        packet.sysid = 255;
-        packet.compid = 190;
+        packet.sysid = sysid;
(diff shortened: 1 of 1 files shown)


julianoes:
@auturgy I think this makes sense, right?

auturgy:
Seems to. It shouldn’t hardcode the ID’s.
I assume you guys hit this in testing and not just by code inspection? Curious as to whether they’re already overwritten somewhere else (doesn’t look like it, but I’ve never used the Java lib, and it’s strange this hasn’t been raised before).

JonasVautherin:
> I assume you guys hit this in testing and not just by code inspection?

Yes, I was unpacking and then repacking packets before forwarding them, and realized then :grin:. My guess is that most implementations do not unpack/repack, and instead call `pack()` __and__ set the IDs manually after that...

But it works well for me with those changes: now unpack does what I expect from it :+1:

peterbarker:
I've poked the commit message to be a bit less intimidating/more specific.

peterbarker:
Merged, thanks!

PR #547, opened 2021-06-02
Close the mmap object

auxym:
The inherited close() method from the mavlogfile class only closes the file object. However, the mmap object also keeps a handle on the file, which prevents it from being deleted or opened in read mode even after we call close() This commit closes both the file object (by calling the super method) and the mmap object (directly), ensuring that all handles are freed.

### DIFF
--- a/mavutil.py
+++ b/mavutil.py
@@ -1429,7 +1429,11 @@ class mavmmaplog(mavlogfile):
     def rewind(self):
         '''rewind to start of log'''
         self._rewind()
-        
+
+    def close(self):
+        super(mavmmaplog, self).close()
+        self.data_map.close()
+
     def init_arrays(self, progress_callback=None):
         '''initialise arrays for fast recv_match()'''
 


mavutil.py
```diff
@@ -1429,7 +1429,11 @@ def _rewind(self):
     def rewind(self):
         '''rewind to start of log'''
         self._rewind()
-        
+
+    def close(self):
+        super().close()
```
peterbarker, line 1434: This won't work on Python2 - and yes, sadly, we're going to continue to support it for a while yet.

```suggestion
        super(mavmmaplog, self).close()
```

peterbarker (changes requested):
This looks like an important fix, thanks for creating the PR.

@auxym are you interested in following this up?

mavutil.py
auxym: Sure thing --  I haven't used python2 personally in a while!

peterbarker: Lucky you :-)

peterbarker:
I've squashed the two commits together and adjusted the commit message a little.

peterbarker:
Merged, thanks!

PR #549, opened 2021-06-03
Update mavlink_signature_check code to be compiler agnostic

JXS2012:
The bitfield packing in mavlink_message_t behaves differently on gcc and windows compilers such as VS2019. This difference caused the mavlink_signature_check function unable to verify signature when built with VS2019. Update the mavlink_signature_check code to not depend on the bitfield packing behavior.

### DIFF
--- a/generator/C/include_v2.0/mavlink_helpers.h
+++ b/generator/C/include_v2.0/mavlink_helpers.h
@@ -133,7 +133,8 @@ MAVLINK_HELPER bool mavlink_signature_check(mavlink_signing_t *signing,
         
 	mavlink_sha256_init(&ctx);
 	mavlink_sha256_update(&ctx, signing->secret_key, sizeof(signing->secret_key));
-	mavlink_sha256_update(&ctx, p, MAVLINK_CORE_HEADER_LEN+1+msg->len);
+	mavlink_sha256_update(&ctx, p, MAVLINK_NUM_HEADER_BYTES);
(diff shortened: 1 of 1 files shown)


JXS2012:
Hi, this pull request originated from https://github.com/mavlink/mavlink/issues/1617 to fix the signature check issue on VS2019 compiler. It was in this pending approval state for a while. Anything I can do to move this ahead?

peterbarker:
Needs to be a single commit.  Let me know if that's going to be a problem for you to arrange.

peterbarker:
I've run ardupilot's Plane autotests against this change it passed just fine.

Note the two separate changes here.  The simple one is changing `MAVLINK_CORE_HEADER_LEN+1` to `MAVLINK_NUM_HEADER_BYTES`  - that's how the latter is `#define`d, so all good.

The second change here is that the checksum is being done in two chunks - all of the header bytes, and then the message payload.

The trick here is that we're using the "64-bit chunks" aspect of the message payload (`((const char *)(&((msg)->payload64[0])`) rather than across a (uint8_t*) cast of the message.

The linked issue states that vscode is storing the `uint32_t msgid:32` in 4 bytes rather than three.  This PR would seem a reasonable fix for that problem.

peterbarker:
Update: I've squashed the commits and fixed the commit message up a bit

tridge (approved):

tridge:
@JXS2012 nice find, thanks!

PR #550, opened 2021-06-10
DFReader.py: fix qnan detection under Python3

peterbarker:
On Py3:

Before:
```
1970-01-01 10:00:02.00: NANS {f : qnan, d : qnan, bf : qnan, bd : qnan}
```
After:
```
1970-01-01 10:00:02.00: NANS {f : qnan, d : qnan, bf : nan, bd : nan}
```

Py2 was not broken

### DIFF
--- a/DFReader.py
+++ b/DFReader.py
@@ -208,13 +208,17 @@ class DFMessage(object):
         return self.fmt.name
 
     def __str__(self):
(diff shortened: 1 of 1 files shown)


rmackay9:
great, thanks for this..

PR #555, opened 2021-06-18
Test on the current versions of Python

cclauss:
https://devguide.python.org/#status-of-python-branches v.s.
https://devguide.python.org/devcycle/#end-of-life-branches

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false  # don't cancel if a job from the matrix fails
       matrix:
-        python-version: [2.7, 3.5, 3.6, 3.7, 3.8]
+        python-version: [2.7, 3.5, 3.6, 3.7, 3.8, 3.9]
 
     steps:
       - uses: actions/checkout@v2


peterbarker:
Any reason not to leave 3.5 in there?

It might not be current, but there will be people out there using it and not in a position to move forward.  If we start to fail 3.5 tests we could then re-evaluate, yes?

cclauss:
Asyncio and f-strings came in Python 3.6 so at some point moving forward can help.

Perhaps we should test on legacy Python (until #556 lands) plus 3.5, 3.7, and 3.9 so our testing does not cook the planet.  It quite safe to say that if our tests pass on the odd versions then they should also run on the even version numbers between them.  Thoughts?

cclauss:
Actually looking the [Python minor version axis](https://pypistats.org/packages/pymavlink) shows almost no Python 3.5 usage.

peterbarker (approved):

peterbarker:
Squashed the two commits together.

And that's a damned cool stats page.

peterbarker:
Merged, thanks!

PR #567, opened 2021-06-26
Tools: magfit_WMM: add option to save plot and param file

IamPete1:
This adds the option to save a param file and/or plot. Goal being to reduce the pain of mag-fitting several logs. 

My changing of the file extensions is somewhat clunky, `.BIN` rather than `.bin` would break it. Basically I'm not very familiar with python.


IamPete1:
Should add mag index to saved file names.

tridge (approved):

PR #568, opened 2021-07-03
mavgen_python: fix to_string decoding of arrays with garbage

hendjoshsr71:
Found while doing this PR https://github.com/ArduPilot/ardupilot/pull/16167 when `to_string()` tries to decode strings with any sort of garbage in the arrays it silently fails and returns `_xxx`

There may be better ways to do this and I'm sure there are reasons for why the code was what it was...
And better ways to handle the exception but printing is better than silent...

Exceptions caught silently
L89: bytes objects have no decode method
L98: on the second loop can't concatenate a str to bytes

Also not sure what is up with the `x = u"%s" % r2` except maybe to catch another type of exception??

Tested in Python3 only


alehed:
The issue this is solving should be solved now in master with the merge of #666.

PR #569, opened 2021-07-13
generator: don't check min message length in MAVLink2

sypaq-nexton:
MAVLink2 Empty-Byte Payload Truncation can reduce the size of a valid message. As such, we can only check if a message has a valid minimum length on MAVLink1 links.

See: https://github.com/mavlink/mavlink/issues/1654


generator/C/include_v2.0/mavlink_helpers.h
```diff
@@ -703,15 +703,6 @@ MAVLINK_HELPER uint8_t mavlink_frame_char_buffer(mavlink_message_t* rxmsg,
 		} else {
 			status->parse_state = MAVLINK_PARSE_STATE_GOT_PAYLOAD;
 		}
-#ifdef MAVLINK_CHECK_MESSAGE_LENGTH
```
hamishwillee, line 706: @peterbarker `rxmsg->len`is the actual payload length (zero truncated). `mavlink_min_message_length` compares this to a minimal length based on the size of the memory required (i.e. untrucated size). This means any truncated message will fail this test and be dropped.

It is possible I am wrong, but I don't think it is "logically" possible to generically check the minimal case. I'm also not sure it is necessary - if the message is truncated below the zero truncation point you'll generally get a CRC error. 

Further, I'm betting we don't have a MAVLink 2 test with `MAVLINK_CHECK_MESSAGE_LENGTH` enabled?

Upshot, I think this fix makes sense.

sypaq-nexton:
@peterbarker: Any interest in this? Is there anything else you need from me?

PR #574, opened 2021-08-09
[proposal] mavutil.py: support multiple clients with a timeout for udpin

Williangalvani:
The reasoning behind this is that I had multiple simple services using endpoints in mavproxy and each of them required their own endpoint or they couldn't run in parallel. With multiple endpoints on a Raspberry Pi the cpu usage got very high. This saves cpu and bandwidth by allowing multiple connections to an endpoint and only sending data to active clients.

It allows multiple services/clients to share the same endpoint. The clients time out after 3 seconds.
There's an exception for the case when there's only one client left. This is never "disconnected" in order to keep the current behavior. 

Tested with https://github.com/ArduPilot/MAVProxy/pull/944

### DIFF
--- a/mavutil.py
+++ b/mavutil.py
@@ -1004,7 +1004,7 @@ class mavserial(mavfile):
 
 class mavudp(mavfile):
     '''a UDP mavlink socket'''
-    def __init__(self, device, input=True, broadcast=False, source_system=255, source_component=0, use_native=default_native):
+    def __init__(self, device, input=True, broadcast=False, source_system=255, source_component=0, use_native=default_native, timeout=0):
         a = device.split(':')
         if len(a) != 2:
             print("UDP ports must be specified as host:port")
@@ -1023,6 +1023,9 @@ class mavudp(mavfile):
         set_close_on_exec(self.port.fileno())
         self.port.setblocking(0)
         self.last_address = None
+        self.timeout = timeout
+        self.clients = set()
+        self.clients_last_alive = {}
         self.resolved_destination_addr = None
         mavfile.__init__(self, self.port.fileno(), device, source_system=source_system, source_component=source_component, input=input, use_native=use_native)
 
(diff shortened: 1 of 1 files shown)


khancyr (commented):
Why not use the broadcast connection if you need multiple client?

Williangalvani:
> Why not use the broadcast connection if you need multiple client?

it actually behaves the same. The advantage of udpbcast is that you don't have to hardcode the IP, but still, according to the docs:

> A UDP broadcast can be made via the udpbcast:<ip>:<port> that starts as UDP broadcast then locks onto the first client that connects.

I just tested udpbcast with two mavproxy instances. When you connect a new instance, the old one stops receiving data.

khancyr:
My bad, I also confuse broadcast and multicast : https://github.com/ArduPilot/pymavlink/blob/master/mavutil.py#L1076 . that is what I use on swarm. Everybody is able to connect on it. But the disadvantage is that everybody received what other send so that may not be what you want.

So, yep, the udp server could be interesting too.

mavutil.py
```diff
@@ -1036,15 +1039,25 @@ def recv(self,n=None):
...
         try:
             if self.udp_server:
-                if self.last_address:
-                    self.port.sendto(buf, self.last_address)
+                current_time = time.time()
+                to_remove = set()
+                for address in self.clients:
+                    if self.clients_last_alive[address] + self.timeout > current_time:
+                        self.port.sendto(buf, address)
+                    elif len(self.clients) > 1:
```
ES-Alexander, line 1057:
```suggestion
                    elif len(self.clients) > 1 and len(to_remove) < len(self.clients - 1):
```
If all clients time out at the same time the current behaviour would boot all of them, then once/if one reconnects it would get an endless stream until another one reconnected (assuming the old behaviour of single connections keeping an endless stream is restored, as above). Maybe that's the desired behaviour, but it seems a bit odd. This approach at least keeps the last one, but not sure if such an arbitrary choice is particularly better 🤷

mavutil.py
```diff
@@ -1004,7 +1004,7 @@ def reset(self):
 
 class mavudp(mavfile):
     '''a UDP mavlink socket'''
-    def __init__(self, device, input=True, broadcast=False, source_system=255, source_component=0, use_native=default_native):
+    def __init__(self, device, input=True, broadcast=False, source_system=255, source_component=0, use_native=default_native, timeout=3):
```
peterbarker, line 1007:
```suggestion
    def __init__(self, device, input=True, broadcast=False, source_system=255, source_component=0, use_native=default_native, timeout=None):
```

mavutil.py
```diff
@@ -1036,15 +1039,26 @@ def recv(self,n=None):
...
 
     def write(self, buf):
         try:
             if self.udp_server:
-                if self.last_address:
-                    self.port.sendto(buf, self.last_address)
+                current_time = time.time()
+                to_remove = set()
+                for address in self.clients:
+                    if len(self.clients) == 1 or self.clients_last_alive[address] + self.timeout > current_time:
```
tridge, line 1055: timeout <= 0 should be no timeout
then we added a udptimeout setting in MAVProxy to allow this to be set

tridge (changes requested):

peterbarker (commented):
I'd prefer to *NOT* have the first setting magical.

mavutil.py
Williangalvani: I did 0 meaning no timeout instead

Williangalvani:
@tridge I don't think this error is related to the pr. Maybe something else was trying to access the board?
Could you give it another go?

I tried the same thing with a Pixhawk 1 here and it worked fine.

tridge (approved):

PR #577, opened 2021-08-15
[proposal] mavactive: util to send heartbeats in the background

ES-Alexander:
According to the [heartbeat microservice docs](https://mavlink.io/en/services/heartbeat.html#:~:text=A%20component%20is%20considered%20,messages%20are%20not%20received.):

> A component is considered to be connected to the network if its HEARTBEAT message is regularly received, and disconnected if a number of expected messages are not received.

Accordingly, for active mavlink connections it's useful to be able to send regular heartbeats without needing to set that up manually as part of every script/program.

This PR provides a `mavactive` class for managing an active mavlink connection, which:
- Sets up a thread for sending regular heartbeat messages
- Allows live control of heartbeat parameters (e.g. to represent changes of autopilot state)
- Allows killing (and reviving) the heartbeat thread at any point
- Swaps out the internal file object of the managed mavlink connection with a wrapped equivalent that's thread-safe for writing

### DIFF
--- a/mavactive.py
+++ b/mavactive.py
@@ -1,0 +1,66 @@
+#!/usr/bin/env python
+
+from builtins import object
+
+from time import sleep
+from threading import Thread, Event
+from pymavlink import mavutil
+mavlink = mavutil.mavlink
+
+class mavactive(object):
+    """ A class for managing an active mavlink connection. """
+    def __init__(self, connection, type_=mavlink.MAV_TYPE_GENERIC, autopilot=mavlink.MAV_AUTOPILOT_INVALID, base_mode=0, custom_mode=0,
+                 mavlink_version=0, heartbeat_period=0.95):
+        """ Initialises the program state and starts the heartbeat thread. """
+        self.connection = connection
+        self.type = type_
+        self.autopilot = autopilot
+        self.base_mode = base_mode
+        self.custom_mode = custom_mode
+        self.mavlink_version = mavlink_version
+        self.heartbeat_period = heartbeat_period
+        self._kill = Event()
+        
+        self._birth()
+
+    def _birth(self):
+        """ Creates and starts the heartbeat thread. """
+        self._kill.clear()
+        self.heartbeat_thread = Thread(target=self.heartbeat_repeat)
+        self.heartbeat_thread.start()
+
+    @property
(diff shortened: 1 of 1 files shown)


ES-Alexander:
It's possible this is better placed within `mavutil.py`, for example [here](https://github.com/ArduPilot/pymavlink/blob/master/mavutil.py#L1736) (just after the `mavlink_connection` definition), but I'm not sure so have proposed it in its own file for now.

peterbarker:
On Sat, 14 Aug 2021, ES-Alexander wrote:

> Accordingly, for active mavlink connections it's useful to be able to send regular heartbeats without needing to set that up manually as part of every script/program.

Doesn't this lose the whole idea of a heartbeat - that your component is 
actually alive?

Sending the heartbeat as part of your main control loop (rather than in a 
thread) would seem to be rather more preferable as it is a much stronger 
indication your component is alive and kicking.

ES-Alexander:
> Sending the heartbeat as part of your main control loop (rather than in a thread) would seem to be rather more preferable as it is a much stronger indication your component is alive and kicking.

Interesting take. I'm seeing it more from the perspective of if you had something like a slow sensor or processing step then having a separate thread for heartbeats enables frequency consistency and simplifies the main program logic (i.e. our hearts keep pumping even if we're busy - it doesn't always make sense to focus on it 'consciously'). A similar idea applies for a passive observer, where you want to keep getting information so the heartbeat is just there to say "hey I'm still here, please keep sending messages" - simplified logic is appreciated.

I'd agree that a thread-based heartbeat would be a worse indication of aliveness in the specific case of the main program getting into a blocked state that it can't get out of. For cases where the main program ends (e.g. via completion or exception) both multi-threaded and single-threaded methods are equivalent, and for cases where the main process is subject to long enough waits/calculations to be significant on the heartbeat timescale then a threaded approach is perhaps preferable.

I'd also make the case that having separate threads for different functionalities can be helpful for simplifying logic in each functionality (they can be programmed and operate independently). I'm not saying this approach is _always_ better than a single-threaded approach, but it provides a choice that (at least to me) seems sufficiently useful that it should be available, so users can make their own decision about what is most appropriate for their particular use-case 🙂

PR #578, opened 2021-08-17
mavlink: compile external dialects

dayjaby:
Some of the new commits in mavlink include external dialects from third parties (https://github.com/mavlink/mavlink/pull/1628 and https://github.com/mavlink/mavlink/pull/1653), which caused errors when trying to install pymavlink from source.

With this PR, those external dialects get added and compiled as well. The dialects can be imported in python for example via

```
from pymavlink.external.v10 import storm32
```


dayjaby:
The insane size of this PR comes from the new `all.xml` and `development.xml` in mavlink. Nothing to worry about 8)

khancyr:
Could you remove the generate py file and the xml ? They don't belong to pymavlink

dayjaby:
@khancyr I put that into gitignore now, so this accident doesn't happen if someone just does a `git add .`

dayjaby:
@khancyr Could you review again?

dayjaby:
Considering this fixes a breaking issue, can we please move this forward?

peterbarker:
On Fri, 1 Oct 2021, David Jablonski wrote:

> Considering this fixes a breaking issue, can we please move this forward?

I'm going to discuss this with tridge.

My immediate thought on seeing this was "external dialects are more 
painful than what it gains us" and that the dialects should simply be 
moved into the standard message definitions folder.  However, we need to 
know the justifications for "external" and whether in light of the pain 
the system is causing those justifications outweight the pain....

dayjaby:
I thought that was decided already? But yeah, I agree. We had some "external" dialects in the message definitions all the time already, like uAvionix.

hamishwillee:
@dayjaby This was for non-stakeholder dialects/dialects that aren't ready to start working with the standard. We were hesitant to accept them but we decided that as long as there was no management cost it would be OK  because it might provide a path for eventual engagement/collaboration. Later on I think we saw some benefits of having them in all.xml so we can keep our ids from clashing.

@peterbarker Is the problem that the dialects are out of tree? Could we fix this problem "technically" by putting them in a folder under the "approved" messages?

Otherwise we should discuss the options in the dev call - having them in the main folder or removing them altogether.

peterbarker:
On Tue, 5 Oct 2021, Hamish Willee wrote:

> Otherwise we should discuss the options in the dev call - having them in the main folder or removing them altogether.

Just put them in the main folder IMO.

We already want maintainer's contact information at the top, it could 
perhaps be augmented with "not maintained by mavlink developers as part of 
mavlink development" in the top, perhaps.

I don't think they should be removed; ArduPilot would like to have some 
information about soaring coming back to the GCS, and either swiping 
somebody else's message (or learning from their mistakes!) is good.

hamishwillee:
@peterbarker (& @auturgy) FYI we discussed in the dev call and decided we needed a little more context on the problem before we might be able to say anything useful on this topic. 

The context for this is that the error came up when building [mavlink-gbp-release](https://github.com/mavlink/mavlink-gbp-release) which is described as: "git-buildpackage repository for releasing mavlink as 3-rd party library for ROS".

The build problem comes IFF you build pymavlink - ie `python3 setup.py install` as shown in the comment above ^^^. 

If you build directly using mavgen from mavlink/mavlink, libraries are built for python or C, and appear to be correct (modulo the enum generation issues you know about). My concern though is that perhaps this only "appears" to work and was not properly validated.
```
python3 -m pymavlink.tools.mavgen --lang=Python --wire-protocol=2.0 --output=generated/include/mavlink/v2.0 message_definitions/v1.0/all.xml
```

**Does the normal Pymavlink build work?**

The PR has two parts - changes to setup.py to find the external files, and changes to mavgen.py to _explicitly_ support external files. If we decide to support external files we will need both bits. IMO though it would have been better to split them into two to make review easier.

FWIW I like being able to build all.xml in order to be able to manage the id space. On balance I feel that we should support this.

PR #591, opened 2021-10-20
mavwp: improve WP display after NAV_LAND and NAV_VTOL_LAND

tridge:
these items should "end" a mission sequence


khancyr:
would this prevent mission with onloading payload and autotakeoff ?

tridge:
> would this prevent mission with onloading payload and autotakeoff ?

no

PR #592, opened 2021-11-10
XSD: WIP tags in enums

hamishwillee:
Enums can also be work in progress. This enables them to be marked as such.

Further, this allows a "since" attribute to be added to the wip tag (but does not mandate it). The idea is that since allows you to easily see when the WIP tag was created, and thereby more easily assess whether the tag should be removed (or the whole element if implementation did not end up happening)

FYI @amilcarlucas - as per discussion in  https://github.com/mavlink/mavlink/issues/1732#issuecomment-960388791

### DIFF
--- a/generator/mavschema.xsd
+++ b/generator/mavschema.xsd
@@ -208,6 +208,7 @@
         <xs:sequence>
             <xs:element ref="description" minOccurs="0"/>
         </xs:sequence>
+        <xs:attribute ref="since" />
     </xs:complexType>
 </xs:element>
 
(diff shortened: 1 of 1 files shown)


hamishwillee:
@amilcarlucas Can we get this in? It makes sense unless we decide we don't want WIP at all.

amilcarlucas:
Looks good to me. @tridge can I merge this?

amilcarlucas (approved):

peterbarker:
Merged, thanks!

PR #600, opened 2021-12-01
Use symbolic links to message definitions

dayjaby:
Instead of copying the message definition xml files, we create symbolic
links to them, which preserves the folder structure of the original
message definitions. This fixes the setup.py install to work again,
while still allowing external include paths in the message definitions.

Unlike #578 this PR is less invasive, but does not support standalone
mavlink generation for external dialects via setup.py install.
External dialects are, however, correctly included in the "all" dialect.

Tested on Ubuntu. Not sure how well symlinks behave under windows. Any feedback would be very welcome!

PR #601, opened 2021-12-01
fix setup.py by making it use mavgen instead of mavgen_python_dialect

dayjaby:
A replacement for #600. As already well documented (https://mavlink.io/en/getting_started/generate_libraries.html), using tools/mavgen is the proper way of creating dialects. Still, `sudo python3 setup.py install` remains a convenient and powerful installation tool for pymavlink, so I made setup.py rely on the same mechanics as tools/mavgen.

Advantages:

- setup.py works again
- external dialects are compiled and put into dialects/v10 (or v20)
- does not rely on possibly not correctly working symbolic links

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,7 @@
 from __future__ import absolute_import, print_function
 from setuptools.command.build_py import build_py
+from argparse import Namespace
+
 # Work around mbcs bug in distutils.
 # http://bugs.python.org/issue10945
 import codecs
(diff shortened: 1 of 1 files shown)


peterbarker:
I'm intrigued by your patch here to use mavgen rather than the python generator.  Unfortunately it's mixed with some external-dialect changes here - perhaps contrary to intent considering the commit message.

peterbarker:
One issue with doing this would be that the dynamic generation (https://github.com/ArduPilot/pymavlink/blob//mavutil.py#L121) would give you a different codepath than installing them as part of `setup.py`.

PR #608, opened 2021-12-17
magfit_WMM: support new compass param naming

IamPete1:
This adds magfit support for the new param naming scheme in https://github.com/ArduPilot/ardupilot/pull/19490

Tested with current AP and built from that branch.

### DIFF
--- a/tools/magfit_WMM.py
+++ b/tools/magfit_WMM.py
@@ -47,11 +47,6 @@ import numpy
 earth_field = None
 declination = None
 
-if args.mag > 1:
-    mag_idx = str(args.mag)
-else:
-    mag_idx = ''
-
 class Correction:
     def __init__(self):
         self.offsets = Vector3(0.0, 0.0, 0.0)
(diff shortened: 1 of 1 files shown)


tools/magfit_WMM.py
```diff
@@ -245,21 +241,32 @@ def remove_offsets(MAG, BAT, c):
...
+    global earth_field, declination, new_param_format
 
     global data
     data = []
 
     ATT = None
     BAT = None
 
-    mag_msg = 'MAG%s' % mag_idx
+    if args.mag == 1:
```
tridge, line 266: this fails for old logs with MAG, MAG2, MAG3

tridge (dismissed):

tools/magfit_WMM.py
IamPete1: still works, tested on a 4.0 log MAG and MAG2

IamPete1:
There must be some clever stuff in the `recv_match` function to handle the instance logging because this script does not deal with it. Hence this change did not break it.

PR #616, opened 2022-01-12
mavparms.py: join with tabs for qgc

peterbarker:


durka:
As discussed on Discord, the issue with this is it marks everything as a float, but QGC has its own XML of all possible parameters that it uses to decide that certain parameters are actually ints. Then, it uses its own idea of types to convert the strings in the file (ignoring the type field in the file!). A string such as 2.0 doesn't parse as int, so any non-float parameters are read from the file as zero.

peterbarker:
@durka I guess this one isn't actually particularly useful, then?

PR #618, opened 2022-01-13
def/undef of MAVLINK_THIS_XML_HASH to MAVLINK_FILENAME_XML_…

hamishwillee:
Follows on from #537. As suggested by @peterbarker "Can you think of a reason we couldn't emit `MAVLINK_COMMON_XML_HASH` and `MAVLINK_MINIMAL_XML_HASH` instead of having to undef/redef MAVLINK_THIS_XML_HASH all the time?"

The aim of the def is to ensure that we load the primary and do some processing, then we process the include files but we DONT include the stuff at the end. We only include it when we come back to the primary file after doing all the includes. This is still what is happening.  I like it more because it is clear that we're asking "is common (say) the primary". I find it more readable.

Just FYI, the before and after zip is attached showing the file changes.

[compare-hash_variants.zip](https://github.com/ArduPilot/pymavlink/files/7859814/compare-hash_variants.zip)


peterbarker:
Tested and merged - thanks!

PR #622, opened 2022-01-16
mavgen_javascript.py: small bugfixes

davidbuzz:
- allow loading as both common.js (Node), and/or vanilla javascript in-browser
 - include customised jspack and long  as part of node package.json - 'npm install' now works in generated output folder without further tweaks.
-  bump version no in generated output as its never happened before.
- fix package.json urls and keywords.


davidbuzz:
also tested here:  https://github.com/davidbuzz/MAVAgent/commit/effb4f791145ed88dbf26550279ebcd9d8c7eb31

PR #624, opened 2022-01-20
Mavutil: Get vehicle data from HIGH_LATENCY2

stephendade:
This PR allows mavutil to get system information from the HIGH_LATENCY2 message, if available.

This is useful when the HEARTBEAT message is not being sent (ie most high latency links).


mavutil.py
```diff
@@ -404,6 +404,14 @@ def post_message(self, msg):
                 self.sysid_state[self.sysid].armed = (msg.base_mode & mavlink.MAV_MODE_FLAG_SAFETY_ARMED)
                 self.sysid_state[self.sysid].mav_type = msg.type
                 self.sysid_state[self.sysid].mav_autopilot = msg.autopilot
+        elif type == 'HIGH_LATENCY2':
+                self.flightmode = mode_string_v10(msg)
```
peterbarker, line 408: Shouldn't we really do the sysid-locking in here like we do for heartbeat?

stephendade: Quite right - I've now added that in

PR #626, opened 2022-02-07
mavgen: use MAVLink_unknown messages for unknown XML

tridge:
this makes it possible for pymavlink tools like MAVProxy to forward
unknown messages


davidbuzz:
Looks plausible

peterbarker (approved):
The fact that we don't add a header to the return message means we can't see this in MAVProxy, which is kind of sad.

Also copying in from Skype:
```
Hmmm.  Not working exactly as I was expecting here.  Not getting into mpstatus.msg in MAVProxy, so not visible in status output

.... there's also the fact the patch forwards unsigned packets with message IDs it doesn't recognise, even if signing is purportedly in use.
```

But better than what we've currently got :-)

Issue #615, opened 2022-01-12
Problem with generator for C (message without fields)


Kazeshirou:
comma before `${{arg_fields: ${array_const}${type} ${array_prefix}${name}, }}` breaks code (syntax error) for empty messages (without fields)

For example
https://github.com/ArduPilot/pymavlink/blob/8c72af35d782224b8886e8f64205ac358e2ea056/generator/mavgen_c.py#L358

This error appears in few places in file.

It's interesting that generating test code processes empty messages correctly  `${{arg_fields:, packet1.${name} }}`  

https://github.com/ArduPilot/pymavlink/blob/8c72af35d782224b8886e8f64205ac358e2ea056/generator/mavgen_c.py#L486

peterbarker:
@Kazeshirou is this an important thing to fix?

I wouldn't be surprised to find out a lot of the generators are broken.

Generally supporting empty messages can be a problem in some languages as zero-length objects can't have memory addresses (not a problem in our code, but perhaps someone else generating headers will have a problem) - so perhaps we should just say that a mavlink message has to have at least one field?

Kazeshirou:
"mavlink message has to have at least one field" warning in documentation is also good solution, why not. This fact has to be obvious from docs. Right now, it isn't

peterbarker:
Where would you expect to see this noted?

I love making things documentations problems (@hamishwillee ) :-)

Kazeshirou:
https://ardupilot.org/dev/docs/mavlink-basics.html
In Message definition, right after "Messages are no more than 263 bytes" =) 

And in readme in the end of section [Documention](https://github.com/ArduPilot/pymavlink#documentation)

Kazeshirou:
and  Data (1 - 255 byte not 0 ) in https://github.com/ArduPilot/ardupilot_wiki/blob/master/dev/source/docs/mavlink-basics.rst  that

hamishwillee:
@peterbarker If we agree the "messages must have at least one field rule", the best way to help users would be to add it to validation in https://github.com/ArduPilot/pymavlink/blob/master/generator/mavschema.xsd . @amilcarlucas Could you make that happen?
From my perspective this is fine, because it was a limitation I knew about "by osmosis". 

From my perspective that would get documented in https://mavlink.io/en/guide/xml_schema.html and https://mavlink.io/en/guide/define_xml_element.html

[...]

PR #630, opened 2022-02-08
XSD validation requires messages have at least one field

hamishwillee:
This enforces in the XSD/validator that every message must have at least one field.

@tridge @julianoes @auturgy Are you OK with this restriction? 
The cases I can think of where you might want to send an empty message are all commands - in almost every case you will want an ACK so you probably want to use a MAV_CMD. Also it is pragmatic.

This originates from #615, which demonstrates that mavgen_c generated code is broken if no field is specified. More generally, @peterbarker notes that:
- many of the generators might be broken
- empty messages may be a problem in some languages as zero-length objects can't have memory addresses.

### DIFF
--- a/generator/mavschema.xsd
+++ b/generator/mavschema.xsd
@@ -272,7 +272,7 @@
                     <xs:element ref="wip"/>
                 </xs:choice>
                 <xs:element ref="description" minOccurs="1" maxOccurs="1"/>
-                <xs:element ref="field" minOccurs="0" maxOccurs="unbounded"/>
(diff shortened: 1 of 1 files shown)


julianoes (approved):
Intuitively I would say this makes sense.

hamishwillee:
@peterbarker @amilcarlucas Can you please review this?

peterbarker:
@hamishwillee will bring up on DevCallEU tonight.

peterbarker:
Merged after testing that it fails CI if I require 100 fields in every message.

PR #636, opened 2022-02-17
Feature/add metadata to Java MAVLink message class fields

rrr6399:
This is a pull request that just adds a Units and Description annotation to the MAVLink message classes for Java so that they can be retrieved at runtime for plotting and such similar to the C# class generator.


rrr6399:
@amilcarlucas @peterbarker @magicrub  
Did you all have any comment on this approach? Can it be merged?

amilcarlucas:
I am all in favor of adding units everywhere. 🙌 
If it works, you have my vote

amilcarlucas:
The enum values are a bit duplicated, what happens when a new unit gets introduced in the .xml files?

rrr6399:
> The enum values are a bit duplicated, what happens when a new unit gets introduced in the .xml files?

I added the enum class to make it easy to use the known supported values when retrieving the units, but the text string can be used as well since the annotation takes a string and not an annotation as an argument. I decided against using the enum for the annotation argument for the reason you mentioned here.

tridge:
@rrr6399 I'm not a java expert, but this seems like a good idea. Can you rebase to fix the conflict?

rrr6399:
@tridge I merged in the master changes, but the recent change (https://github.com/ArduPilot/pymavlink/issues/638 ) in master that changed the enum type to long from int is a non-optimal change that breaks existing code and forces developer to cast the command enums from long to int to use them in the MAVLINK command messages. The command type in those messages is actually an unsigned short, but the Java code promoted the type to an int. 

The offending enums in the new MAVLink messages are actually masks. I can see the mask type being promoted to longs to make the change less disruptive to existing code. Unfortunately, the masks aren't annotated as masks. All the enum types (masks and command options) get lumped into one category. Anyway, I think there needs to be a more in depth discussion on the right way to accommodate these large masks in Java code before wholesale changing all enums to longs. (Also, it appears that the new offending version of the MAVLink message definitions aren't being used in the ArduPilot version of pymavlink right now.)

rrr6399:
@tridge @amilcarlucas Any opinion on my comment above?

amilcarlucas:
I agree with your comments, but I do not use java at all. So my opinion does not count.

amilcarlucas:
I now passes the checks, and has no conflicts, but there are still merge commits when there should be none.

rrr6399:
> I now passes the checks, and has no conflicts, but there are still merge commits when there should be none.
Is there anyway to squash or remove the merge commits or do I need to start a new branch?

rrr6399:
@amilcarlucas Would it be best just to create another merge request with a clean branch? Or, is there a simple way to remove the merge from this branch? I removed the one from the head of the branch, but I also did a merge a few commits ago as well.

amilcarlucas:
It is called pull-request, not merge request. No need for a new one.
All you need to do is issue:
```
git pull origin master
git rebase origin/master
git push -f
```

rrr6399:
@amilcarlucas i thought that's what I did, but it didn't get rid of one of the merges I did early on. I'll try again. Thanks.

rrr6399:
@amilcarlucas I'm going to close this pull request. I opened a new one with a cleaner history since I couldn't figure out how to remove the merge history from this one. The new pull request is here:
https://github.com/ArduPilot/pymavlink/pull/694

Issue #638, opened 2022-02-24
[Java generator] error: integer number too large


TSC21:
Hey there,

I just tried to generate and compile the Java generated library from the the `common` dialect and the compiler fails with the following error:
```sh
output_dir_java/enums/MAV_SYS_STATUS_SENSOR.java:44: error: integer number too large
   public static final int MAV_SYS_STATUS_EXTENSION_USED = 2147483648; /* 0x80000000 Extended bit-field are used for further sensor status bits (needs to be set in onboard_control_sensors_present only) | */
                                                           ^
output_dir_java/enums/MAV_SYS_STATUS_SENSOR.java:45: error: integer number too large
   public static final int MAV_SYS_STATUS_SENSOR_ENUM_END = 2147483649; /*  | */
                                                            ^
output_dir_java/enums/HIL_SENSOR_UPDATED_FLAGS.java:27: error: integer number too large
   public static final int HIL_SENSOR_UPDATED_RESET = 2147483648; /* Full reset of attitude/position/velocities/etc was performed in sim (Bit 31). | */
                                                      ^
output_dir_java/enums/HIL_SENSOR_UPDATED_FLAGS.java:28: error: integer number too large
   public static final int HIL_SENSOR_UPDATED_FLAGS_ENUM_END = 2147483649; /*  | */
                                                               ^
4 errors
```

Steps to reproduce:

1. Use `mavgen.py` to generate the V2 library for Java in an `output_dir`
2. _javac -d output_dir `find output_dir -name "*.java"`_
3. The error above will pop.

This is confirmed to happen with both Java 11 and Java 14.

Thanks in advance for the assistance solving this!

abroekhof:
This should be re-opened since the PR above was reverted. The bug still present in current versions.

PR #639, opened 2022-02-24
mavgen_java.py: generate_enums: assume all integers to be 64 bits

TSC21:
Fixes https://github.com/ArduPilot/pymavlink/issues/638. Though I wonder if it is acceptable to use a larger data struct by default to enums just because of a single enumeration. Anyway we might eventually have other cases where this will happen.


TSC21:
@peterbarker is this something you can review? Thanks in advance!

peterbarker:
@TSC21 I haven't really dealt with these bindings much.

Perhaps @len0rd or @magicrub could comment?

TSC21:
@len0rd @magicrub any chance you can help move this forward? Thanks!

magicrub (approved):
The one line change looks ok, but there's a ton of whitespace changes. They're good changes.. so that's ok.. but still kind of annoying. Anyway, it's likely worth merging.

rrr6399:
This change breaks everybody's Java code and forces a int cast to the command enum when using it in a command. The large values are related to masks. The longs should only be used for masks.

PR #642, opened 2022-02-28
speed up parsing of bin logs

tridge:
use a dictionary of sets to record which instances of a message have
been seen before. This avoids us having to parse all instance based
messages fully during the load.

as we have added more instance messages the time to load the log has
become progressively longer. With this change it now becomes short
again

on a recent 160MByte quadplane log the time to load the log with this
change drops from 39.9s to 6.5s


peterbarker (approved):

davidbuzz:
i haven't tested it, but def like the idea.

PR #650, opened 2022-03-15
mavutil.py: default to using 'all' dialect in place of ardupilotmega

peterbarker:
I've been running this locally for a while and it seems to be fine


Ryanf55 (approved):
Needs rebase

tridge (approved):

PR #653, opened 2022-03-25
Fix enum generation

peterbarker:
This is a rework of https://github.com/ArduPilot/pymavlink/pull/544/files

@hamishwillee


tools/mavgen.py
```diff
@@ -25,7 +25,7 @@
 parser.add_argument("--no-validate", action="store_false", dest="validate", default=mavgen.DEFAULT_VALIDATE, help="Do not perform XML validation. Can speed up code generation if XML files are known to be correct.")
 parser.add_argument("--error-limit", default=mavgen.DEFAULT_ERROR_LIMIT, help="maximum number of validation errors to display")
 parser.add_argument("--strict-units", action="store_true", dest="strict_units", default=mavgen.DEFAULT_STRICT_UNITS, help="Perform validation of units attributes.")
-parser.add_argument("definitions", metavar="XML", nargs="+", help="MAVLink definitions")
+parser.add_argument("definitions", metavar="XML", nargs=1, help="MAVLink definitions")
```
hamishwillee, line 28: Does this mean you only allow one definition file now?

peterbarker: On input, yes.  I did have some code to create an file that just included multiple.

This was the original "shove two XML files into one another" code which was around before `<include>` was, IIUC

hamishwillee: Thanks. I don't have a problem with the change - I think it is better. My concern is that we need to document these changes   - not just in https://mavlink.io/en/getting_started/generate_libraries.html but in the usage section.

PR #654, opened 2022-03-25
Add python3 target with type annotations

alehed:
This adds a second generation target for python.
It is meant to complement the existing target and be useful if somebody wants more modern
generated python code at the cost of dropping support for older Python versions.

Main differences to the current target:

* Full type annotations (mypy passes with the most strict settings)
* Dropped support for Python 2 (and probably Python <3.6)
* No runtime dependency on pymavlink (The x25crc function is directly embedded into the file)
* Dropped support for mavnative
* `enums` is a single dict literal
* `name` property of MAVLink_message was renamed `msgname` due to conflicts with the name property used in common.xml
* More generous spacing increases readability (the output is quite close to `black`-formatted code)

All bullet points apart from the first two are negotiable.


khancyr (commented):
That is interesting but as we cannot drop py2 for now, this will create a duplicated generator to maintain.

But I would be in favor to style code style !

alehed:
I was thinking that maybe it could be like JavaScript where there is "JavaScript_Stable" and "JavaScript_NextGen".

I recognize that this increases maintenance burden. Given that this is something we are actively using at my company, there is interest in maintaining this target.

alehed:
If it is preferred I can:

* Backport the non-python2-specific changes to the python2 target (each change in its own PR)
* Add a flag to the existing python generator that it optionally emits type annotations

alehed:
Closed in favor of smaller PRs that modify `mavgen_python.py`.

PR #662, opened 2022-04-20
mavgen_python: Reformat output

alehed:
This PR is the first in a series to add a python3 target based off of #654, but in a way that doesn't increase the maintenance burden.

Right now the generated python code contains many instances of inconsistent indentation (a mix of 4 spaces, 8 spaces and 2 spaces), this changes all the indentation to 4 spaces.

This contains the following formatting changes:

* The code itself is slightly reformatted for more space
* Whenever possible, double quotes are used instead of single quotes (in the code and the output)
* The output code consistently uses 4 spaces of indentation and doesn't produce whitespace linter errors
* The enum in the output code is one big literal instead of being constructed

There is also the option to use more vertical space in long lists and parameters, but I'm not sure if this is desired.

This PR is probably the most controversial. I'm happy to adjust parts of it for it to get merged. Let me know what you think.


alehed:
The following gist shows a comparison between the generated code before and after: <https://gist.github.com/alehed/ab70682d44fe4149a974a79d7f16d802>

alehed:
@khancyr I added a CI check. Is this what you were thinking about? Also can you approve the CI workflow?

PR #664, opened 2022-04-27
mavgen_python: Multiple improvements and bug fixes

alehed:
This PR is best reviewed commit-by-commit starting where the reformat_output branch leaves off.

This lays the groundwork for the following PR which adds type annotations in Python3 code.


alehed:
@peterbarker I modified the PayLoadTrimZeros test because I think it is wrong.

I'm a bit confused about the test that failed. It uses code that is generated for mavlink 1.0 and sets the version to 2. Because the version is set to an int and not a string, in the past only parts of the generated code used to assume mavlink 1.0. Should code generated with protocol 1.0 allow sending mavlink 2.0 messages?

tridge:
on the name attribute problem, it would be nice to cope with .name in existing code (eg. in msgrates module in mavproxy)
maybe use a __getattr__() to check dictionary for the field, if field is 'name' and if no field and have msgname then return msgname
and a comment that this is a temporary workaround, can be removed in 1 years time

alehed:
As discussed, added the name class variable property back, but in such a way that using it will raise an exception when using on an instance and printing a warning when using on a class. It is also added in such a way that mypy doesn't know about it (so it doesn't print warnings about it or suggests using it). Like this it should be ready for merging.

tridge (approved):

PR #665, opened 2022-04-28
mavgen_python: Only generate code needed for mavlink version

alehed:
This PR is based on #664, hence is only one commit.

It removes a bunch of runtime checks that can actually be done at generation time, simplifying the generated code.


alehed:
This causes issues with the the tests, closing this PR.

PR #666, opened 2022-04-28
mavgen_python: Add python3 target with type annotations

alehed:
This PR is based on #664  and contains 4 commits. It adds type annotation support in way that should not break mavlink for people stuck on Python2. Everyone else should benefit from the type annotations.

### DIFF
--- a/.github/workflows/mypy-generated.ini
+++ b/.github/workflows/mypy-generated.ini
@@ -1,0 +1,49 @@
+[mypy]
+python_version = 3.6
+namespace_packages = False
+ignore_missing_imports = False
+follow_imports = normal
+no_site_packages = True
+no_silence_site_packages = False
+
+disallow_any_unimported = True
+# disallow_any_expr = True
+disallow_any_decorated = True
+# disallow_any_explicit = True
+disallow_any_generics = True
+disallow_subclassing_any = True
+
+disallow_untyped_calls = True
+disallow_untyped_defs = True
+disallow_incomplete_defs = True
+check_untyped_defs = True
+disallow_untyped_decorators = True
+
+no_implicit_optional = True
+strict_optional = True
+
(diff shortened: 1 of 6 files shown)


alehed:
Note that the type of file should be

```python3
class BinaryWriter(typing.Protocol):
    def write(self, b: bytes) -> int:
        ...
```

and not Any, but typing.Protocol is a python 3.8 thing and I want to support Python 3.6+ with the type annotations.

tridge:
notes from dev call:
 - need to pay special attention to mavproxy (including MAVExplorer), ardupilot autotest suite (Tools/autotest/autotest.py) and mavros
 - need CI (github actions) tests to ensure this works both for py2 and py3
 - need a plan for deployment, so existing scripts automatically get the new generator if running on py3
 - note that most scripts access pymavlink via mavutil

vooon:
Mavros doesn't use `--lang=Python` in it's build, but may use as a weak dependency for console scripts.
I'd suggest leave `--lang=Python` and decide which implementation to use based on interpreter version. E.g. if major version == 3 => use Python3.

alehed:
> Mavros doesn't use `--lang=Python` in it's build, but may use as a weak dependency for console scripts. I'd suggest leave `--lang=Python` and decide which implementation to use based on interpreter version. E.g. if major version == 3 => use Python3.

In this PR if you specify "Python" or nothing (the default is "Python") it will only generate the type annotations if you are running it with python >= python3.6

alehed:
> getting this with py2 build:
> python2 setup.py build install --user

While these errors are annoying, I don't see anything short of disabling bytecode compiling for those versions that fixes these warnings. Note that, like this all files that can be byte-compiled, are byte-compiled.

tridge (approved):

tridge (approved):

PR #671, opened 2022-05-06
Fix Wireshark parsing logic

rotu:
Previously, fields sometimes showed the wrong value, byte offsets were from the beginning of the IP packet, and truncated fields were omitted.
Now, the correct value is shown, byte offsets are from the beginning of the MAVLink packet, and truncated values are kept.

image

image


peterbarker:
Merged, thanks!

hamishwillee:
@rotu In case you don't get notifications from mavlink/mavlink, I was wondering if you could advise on https://github.com/mavlink/mavlink/issues/1978 (as you were the last person to modify the Wirshark plugin, and will know a lot more than I do)! If not, is there anyone else you recommend we might ask for help?

PR #678, opened 2022-05-20
Enum values larger than Java max int

radarku:
Fix to the Java generator for enum values that are larger than Java maximum integer size.

Without this PR, mavgen would produce MAVLink library code that cannot compile.  This change preserves existing interoperability (i.e., current source code that expects integer enums and int types for Java switch statements), while also allowing larger enum values.


radarku:
@peterbarker and @magicrub - @TSC21 mentioned that this might be a slight improvement (compatibility-wise) to https://github.com/ArduPilot/pymavlink/pull/639/files

magicrub:
no objection here. If it compiles and works then ok

radarku:
Hi @magicrub - I tested against the current [RAS-A IOP Specification](https://github.com/Dronecode/air-iop-definitions/blob/master/message_definitions/v1.0/all.xml) and it generates Java message definitions that compile.

magicrub:
@radarku woah, didn't realize this is still pending.. sorry about that! Thanks for updating it with master with a merge but we don't merges in PRs. Please rebase on master instead.

WickedShell (approved):

tridge:
@radarku needs rebase due to conflicts

peterbarker:
@radarku I've merged another PR which was essentially identical to this one - simply because this one needs rebasing and the other didn't.

Thanks for looking at this!

(and thanks for @shancock884 for sorting all this out again)

PR #679, opened 2022-05-31
Better ci

khancyr:
This add pre-commit as part of CI. 
this tool allow to set check easy and add them localy on git-hook for better coding and not wait CI to trigger failure.

The PR also add : 
pyproject.toml to have commom place for python checker/linter tools
.editorconfig: to try to have all ide configured the same 
add .idea on .gitignore for those that are using pyCharm or Clion
add .flake8 config file for local developpement
add shellcheck to CI : it will only fail if the shell script got fatal error.

Correct Line Ending mixing on files : we should have only LF
Correct permission on files
Remove python shebang on python library as they aren't executable


tridge (changes requested):
needs to be info on this in the README.md, and add a run_code_checks.sh which runs the checks, and gives a sensible error message if a required tool isn't available

peterbarker (changes requested):
There's a couple of these "APM"s

Sorry, didn't get far here.

tridge:
lets do this on EU dev call

peterbarker:
Looked briefly at this and it definitely needs a rebase.  Also tridge marked it but wasn't available.

PR #683, opened 2022-06-08
mavgen_python: add method of last resort for displaying something abo…

TunaLobster:
…ut the incoming string bytes

Found it a little annoying while debugging a Lua driver that MAVProxy would only show "_XXX" if the byte values where outside of utf-8 range. This will give the best effort to display ascii characters and then just show the hex value when that fails. To me that would be better than nothing happening.

Similar goals as #568

### DIFF
--- a/generator/mavgen_python.py
+++ b/generator/mavgen_python.py
@@ -105,11 +105,22 @@ def to_string(s):
     # so it's a nasty one. Let's grab as many characters as we can
     r = ""
     try:
-        for c in s:
-            r2 = r + c
-            r2 = r2.encode("ascii", "ignore")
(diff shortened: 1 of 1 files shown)


alehed:
Note that the entire to_string function is removed in PR #666. The correct thing to do is to have the object hold a bytes in python3, not a string. For displaying it in that PR, `MAVLink_message.format_attr` currently uses `decode(error="ignore")`.

The "nice" thing there would probably be to escape all non-printable (not non-unicode) characters with `\x`, but then again, format_attr is used in to_dict, where you probably don't want this behavior...

alehed:
PR #666 was just merged. This issue should be resolved now.

Issue #684, opened 2022-06-09
mavlink_msg_xxx_decode by mavgen_c does not handle extensions of message properly


jnippula:
In mavlink_msg_${name_lower}_decode() function, there is a msg length check and memset(0) in "#else" section to take care of possibly trimmed messages and fill the last removed fields with zeros:
https://github.com/ArduPilot/pymavlink/blob/ba8ca05b2b905c6af2ae20aafa12d7bd09ce9195/generator/mavgen_c.py#L403-L404

However, in the "#if MAVLINK_NEED_BYTE_SWAP || !MAVLINK_ALIGNED_FIELDS" section this check is missing:
https://github.com/ArduPilot/pymavlink/blob/ba8ca05b2b905c6af2ae20aafa12d7bd09ce9195/generator/mavgen_c.py#L400

In case the message has extension fields which have been trimmed out by sender (message length are reduced accordingly), the receiver may just copy those fields from memory outside of the message data area and filling with garbage.

This has caused issues at least with HIL_GPS message, where last two fields (id and yaw) were set zero by sender and the receiver got some random values in id field causing it to detect multiple GPS sensors.

```
    hil_gps->id = mavlink_msg_hil_gps_get_id(msg);
    hil_gps->yaw = mavlink_msg_hil_gps_get_yaw(msg);
```

jnippula:
I created PR for fix proposal: https://github.com/ArduPilot/pymavlink/pull/700

hamishwillee:
@peterbarker Does this make sense?

jlaitine:
Any update on this? We have to do ugly workarounds around the issue, exactly for HIL_GPS message handling. It is also theoretically possible to get a crash due to this in MMU enabled environments (risc-v nuttx KERNEL_BUILD), if the message data area happens to be in the very end of the mapped memory. Unlikely, perhaps, but possible. It is not nice to receive garbage for the extension bytes when the sender has sent 0s.

hamishwillee:
Not from me. @peterbarker ?

jnippula:
Actually, this may affect also to COMMAND_ACK message handling.
https://mavlink.io/en/messages/common.html#COMMAND_ACK

In case the target_component is zero, it can be something garbage in receiver end. E.g. MAVSDK rejects the ack message if target_component does not match with it's own component_id.

jnippula:
Or actually, MAVSDK is using C headers (not CPP ones), so it is affected.

[...]

PR #700, opened 2022-07-01
mavgen_c.py: fix trimmed message decode

jnippula:
This is related to Issue I have created: https://github.com/ArduPilot/pymavlink/issues/684

In case of byte swap or alignment fields, the 'mavlink_msg_xxx_get_' functions
called by message decode functions shall return actual message field value
only if the field is included in the received message. In case the field is
trimmed out it should return zero.

Example output:
```
static inline uint16_t mavlink_msg_hil_gps_get_yaw(const mavlink_message_t* msg)
{
    if (msg->len > 37)
        return _MAV_RETURN_uint16_t(msg,  37);
    return (uint16_t) 0;
}
```

### DIFF
--- a/generator/mavgen_c.py
+++ b/generator/mavgen_c.py
@@ -396,12 +396,12 @@ static inline ${return_type} mavlink_msg_${name_lower}_get_${name}(const mavlink
  */
 static inline void mavlink_msg_${name_lower}_decode(const mavlink_message_t* msg, mavlink_${name_lower}_t* ${name_lower})
 {
+    memset(${name_lower}, 0, MAVLINK_MSG_ID_${name}_LEN);
 #if MAVLINK_NEED_BYTE_SWAP || !MAVLINK_ALIGNED_FIELDS
 ${{ordered_fields:    ${decode_left}mavlink_msg_${name_lower}_get_${name}(msg${decode_right});
 }}
 #else
-        uint8_t len = msg->len < MAVLINK_MSG_ID_${name}_LEN? msg->len : MAVLINK_MSG_ID_${name}_LEN;
-        memset(${name_lower}, 0, MAVLINK_MSG_ID_${name}_LEN);
+    uint8_t len = msg->len < MAVLINK_MSG_ID_${name}_LEN? msg->len : MAVLINK_MSG_ID_${name}_LEN;
     memcpy(${name_lower}, _MAV_PAYLOAD(msg), len);
 #endif
 }
--- a/generator/C/include_v2.0/protocol.h
+++ b/generator/C/include_v2.0/protocol.h
@@ -90,45 +90,45 @@ static inline uint16_t mavlink_msg_get_send_buffer_length(const mavlink_message_
 }
 
 #if MAVLINK_NEED_BYTE_SWAP
-static inline void byte_swap_2(char *dst, const char *src)
+static inline void byte_swap_2(char *dst, const char *src, uint8_t len)
 {
-	dst[0] = src[1];
-	dst[1] = src[0];
+	if (len > 1) dst[0] = src[1];
+	if (len > 0) dst[1] = src[0];
 }
-static inline void byte_swap_4(char *dst, const char *src)
+static inline void byte_swap_4(char *dst, const char *src, uint8_t len)
 {
-	dst[0] = src[3];
-	dst[1] = src[2];
-	dst[2] = src[1];
-	dst[3] = src[0];
+	if (len > 3) dst[0] = src[3];
+	if (len > 2) dst[1] = src[2];
+	if (len > 1) dst[2] = src[1];
+	if (len > 0) dst[3] = src[0];
 }
-static inline void byte_swap_8(char *dst, const char *src)
+static inline void byte_swap_8(char *dst, const char *src, uint8_t len)
 {
-	dst[0] = src[7];
-	dst[1] = src[6];
-	dst[2] = src[5];
-	dst[3] = src[4];
-	dst[4] = src[3];
-	dst[5] = src[2];
-	dst[6] = src[1];
-	dst[7] = src[0];
+	if (len > 7) dst[0] = src[7];
+	if (len > 6) dst[1] = src[6];
+	if (len > 5) dst[2] = src[5];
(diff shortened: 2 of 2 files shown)


JonasVautherin (commented):
This seems like it makes sense to me, but I'm not super familiar with the codebase.

One question: is it correct that `${wire_offset}` is the position of the field `${name}` in the message? Seems like it, but just to be sure :blush:

jnippula:
> One question: is it correct that `${wire_offset}` is the position of the field `${name}` in the message? Seems like it, but just to be sure 😊

Yes, it is.
Just realized that this could be optimized a bit by adding that length check only if the message type contains v2 extension fields (MAVLINK_MSG_ID_XXX_LEN > MAVLINK_MSG_ID_XXX_MAX_LEN).

jnippula:
Checked also if there would be easy way to modify empy template script a bit to add the length check only in case the message really has extension fields. However, I couldn't find quick solution for that, would require more changes in python generator functions, maybe not worth to optimize.

JonasVautherin (approved):

generator/mavgen_c.py
```diff
@@ -384,7 +384,9 @@ def generate_message_h(directory, m):
  */
 static inline ${return_type} mavlink_msg_${name_lower}_get_${name}(const mavlink_message_t* msg${get_arg})
 {
-    return _MAV_RETURN_${type}${array_tag}(msg, ${array_return_arg} ${wire_offset});
+    if (msg->len > ${wire_offset})
+        return _MAV_RETURN_${type}${array_tag}(msg, ${array_return_arg} ${wire_offset});
+    return (${return_type}) 0;
```
peterbarker, line 389:
```suggestion
    if (msg->len <= ${wire_offset}) {
        // zero-bytes have been truncated from this message
        return (${return_type}) 0;
    }
    return _MAV_RETURN_${type}${array_tag}(msg, ${array_return_arg} ${wire_offset});
```

The braces because that's the style used throughout this file, even on single-line bodies - and that many bugs have been introduced by not having them :-)

I reordered to do early return and to make a nice place to put an extra comment about what's going on.

peterbarker (changes requested):
Really good find.

generator/mavgen_c.py
jnippula: I agree, that is better solution. Patch updated.

generator/mavgen_c.py
```diff
@@ -384,6 +384,10 @@ def generate_message_h(directory, m):
  */
 static inline ${return_type} mavlink_msg_${name_lower}_get_${name}(const mavlink_message_t* msg${get_arg})
 {
+    if (msg->len <= ${wire_offset}) {
```
peterbarker, line 387: I wonder if the spec says that if you trim some zeroes then you must trim them all.  My guess is that it doesn't.

I thought this was previously checking if the field offset was past the end of the message?

jnippula: I'm not sure if I understood your point/question here..
Yes, I think nothing prevents the sender to trim only some of the zero bytes in the end, instead of trimming all of them.

The trimming process reduces the actual message length accordingly, so the receiver then just append the remaining missing zeros into the end of the received message in case the length of the received message is less than what is defined/specified for the particular message type.
Do you think there is still something missing/wrong?

peterbarker: Apologies for the slow response.

My concern is that the check you have here doesn't account for the length of the field.  The wire offset may be within the length but the wire-offset-plus-field-length may be outside.  I was thinking in terms of arrays, but I guess anything longer than a byte might be a problem.

jnippula: You are absolutely right. It does not work well in case of longer field types that have zeros in latter bytes (MSB or LSB depending on endianess)

jnippula: Right, this is a bit tricky, implementing partial copy for all possible types and arrays would not sound very handy..

I can think of some more options:

1. Leave the mavlink_msg_<name>_get_<name> function untouched and do an intermediate copy from original msg to full length zero initialized tmp_msg and perform the above function call from that tmp_msg. Then the zeros are already in place and we are not reading from the end of the allocated space.
What to do if malloc is not successful (out of memory case)?

2. Define compiler flag to have possibility to disable sender trim-feature in case system has receiver with MAVLINK_NEED_BYTE_SWAP || !MAVLINK_ALIGNED_FIELDS defined. This is not bullet-proof, because the sender may not be configurable by developer of the receiver.

3. Get rid of the trim feature altogether. How much benefit it gives? At least there are not so many messages with "trimmable" extended fields and not so many bytes to be trimmed from them either. The feature decreases the amount of data transferred over some media, but causes extra load due to memcopy when byte swap not needed and possible incorrect data when byte swap needed.

4) Some new idea to fix this?

Personally I would vote for the option 3).

jnippula:
Ok, I had some time to look at this again. I made another proposal for this.
It should now take all different types into account, also if they are partially or fully trimmed.
What do you think about it?

jnippula:
Added some missing changes from "put" functions.

Now when I run some more tests, I realized that the trimming does not affect only the message extension fields, but all the fields. There is no checking whether the payload field is extension one or not, it is trimmed if it contains zero bytes in the end. So, the trimming affects basically all the mavlink2.0 messages.

jnippula:
Well, not quite fixed yet. Still issues with byte arrays where length is not checked. Working on it.

jnippula:
..and fixed the MAV_RETURN_ macros for 1 byte length fields

jnippula:
I think the pymavlink test scripts shall verify also the case where "MAVLINK_ENDIAN = MAVLINK_BIG_ENDIAN". If I fetch the latest pymavlink (master branch), modify test Makefile to run MAVLINK_ENDIAN as MAVLINK_BIG_ENDIAN, the test_generator.sh manages to test 6 messages and then crash.

I made another PR to enable big_endian test case also:

jnippula:
At least this big endian enabled test set (https://github.com/ArduPilot/pymavlink/pull/812) runs now successfully with this new trimming fix implementation, so it should be in better shape now.

PR #705, opened 2022-07-12
Fix/imports

dforsi:
This PR removes some imports that are not used according to pylint.
I'm creating this PR as draft because while it's easy to check that these changes do not break the tests, it's not so easy for generators and for tools.


dforsi:
Some checks were not successful because of networking issues on the CI server.

PR #706, opened 2022-07-13
isb_extract.py - unit and reading corrections

Pokornz:
- The script isb_extract.py outputs the IMU values as integers (int divided by int = int) -> fixed
- My bin files have the us timestamps in the field TimeUS, not Sample US -> should be verified by other people

### DIFF
--- a/tools/isb_extract.py
+++ b/tools/isb_extract.py
@@ -31,7 +31,7 @@ def mavfft_fttd(logfile):
             self.instance = ffth.instance
             self.sample_rate_hz = ffth.smp_rate
             self.multiplier = ffth.mul
-            self.SampleUS = ffth.SampleUS
+            self.SampleUS = ffth.TimeUS
             self.data = {}
             self.data["X"] = []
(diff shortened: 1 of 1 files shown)


peterbarker:
SampleUS should be correct - it's the sample time rather than the time we stuck it into the log.

The float() changes look useful - could you put those in a separate PR, please?

Pokornz:
The issue with SampleUS is that it is filled with zeros in my logs. 
[2022-07-07 14-55-55_batch_sample.zip](https://github.com/ArduPilot/pymavlink/files/9180188/2022-07-07.14-55-55_batch_sample.zip)
Can we confirm that this is an error in my Ardupilot setup rather than a "new" way Ardupilot handles batch sampling?

peterbarker:
@andyp1per your view?

andyp1per:
SampleUS is correct

Pokornz:
> SampleUS is correct

Any idea why it is filled with zeros in the bin I attached?

PR #708, opened 2022-07-16
mavlogdump: added --reduce-rate msg

tridge:
for reducing large logs for analysis


tools/mavlogdump.py
```diff
@@ -129,6 +130,26 @@ def reduce_msg(mtype, reduction_ratio):
         return False
     return True
 
+last_msg_rate_t = {}
+
+def reduce_rate_msg(m, reduction_rate):
+    '''return True if this msg should be discarded by reduction'''
+    global last_msg_rate_t
+    mtype = m.get_type()
+    if mtype in ['PARM','MSG','FMT','FMTU','MULT','MODE','EVT']:
```
peterbarker, line 139: Speed-wise - and you're working with big logs - might be better to use a `frozenset`

PR #712, opened 2022-07-23
Make the mavfft_isb accessible to other programs

Gone4Dirt:
mavfft_isb is a pretty useful tool.  I like to use it as part of other tools.  I also wanted to integrate it into the MavExplorer partly to add the functionality of this tool into MavExplorer and to reduce duplication.  This PR will enable that.

Functionality should be unchanged so it can be used as it was for everyone that used it previously.


andyp1per (approved):
@Gone4Dirt its impossible for me to review 😄 presumably you have tested this on some logfiles in which case I approve

It's unfortunate this will massively conflict with my FFT heatmap change, but I am not going to finish that anytime soon.

Gone4Dirt:
Yeah, sorry the diff is horrendous.

Basically moved the function into a class.  Made all of the args variables withing that class then added a main == name check and moved the arg parser after that.

Tested all args and all are working as before.

peterbarker:
> Yeah, sorry the diff is horrendous.

Hiding whitespace changes makes it much more palatable.

peterbarker:
Merged, thanks!

PR #717, opened 2022-08-12
Tools: Magfit: throttle based

IamPete1:
Allows throttle based (on copter anyway), this means you can fit to throttle based and also remove a existing throttle based correction from flight data prior to calculating a current based one. Fit with `--Tmot`. You will almost certainly need to allow a larger CMOT limits when doing throttle based, at least a factor of 50 `--max-cmot 500`. 

Also allows to set a larger number of max iterations with `--iter 200` for example.

### DIFF
--- a/tools/magfit_WMM.py
+++ b/tools/magfit_WMM.py
@@ -31,6 +31,7 @@ parser.add_argument("--lon", type=float, default=0, help="longitude")
 parser.add_argument("--att-source", default=None, help="attitude source message")
 parser.add_argument("--save-plot", action='store_true', default=False, help="save plot to .png file")
 parser.add_argument("--save-params", action='store_true', default=False, help="save params to .param file")
+parser.add_argument("--iter", type=int, default=100, help="max optimization iterations")
 
 parser.add_argument("log", metavar="LOG")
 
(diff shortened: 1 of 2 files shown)


tools/magfit_WMM.py
```diff
@@ -24,12 +24,14 @@
 parser.add_argument("--no-offset-change", action='store_true', help="don't change offsets")
 parser.add_argument("--no-cmot-change", action='store_true', help="don't change cmot")
 parser.add_argument("--elliptical", action='store_true', help="fit elliptical corrections")
-parser.add_argument("--cmot", action='store_true', help="fit compassmot corrections")
+parser.add_argument("--cmot", action='store_true', help="fit compassmot corrections using current")
+parser.add_argument("--Tmot", action='store_true', help="fit compassmot corrections using throttle")
```
peterbarker, line 28:
```suggestion
parser.add_argument("--cmot-using-throttle", action='store_true', help="fit compassmot corrections using throttle")
```

tools/magfit_WMM.py
```diff
@@ -145,15 +155,19 @@ def wmm_error(p):
         c.diag = Vector3(1.0, 1.0, 1.0)
         c.offdiag = Vector3(0.0, 0.0, 0.0)
 
-    if args.cmot:
+    if args.cmot or args.Tmot:
         c.cmot = Vector3(p.pop(0), p.pop(0), p.pop(0))
+        if args.cmot:
+            c.cmot_mode = 2
```
peterbarker, line 161:
```suggestion
            c.cmot_mode = CMOT_MODE_CURRENT_BASED
```

tools/magfit_WMM.py
```diff
@@ -186,7 +205,7 @@ def fit_WWW():
         for i in range(3):
             bounds.append((args.min_offdiag,args.max_offdiag))
 
-    if args.cmot:
+    if args.cmot or args.Tmot:
```
peterbarker, line 208:
```suggestion
    if c.cmot_mode != CMOT_MODE_NONE:
```

peterbarker (changes requested):

peterbarker (approved):

peterbarker:
I've tested this as far as "it doesn't grossly break magfit_wmm.py

PR #718, opened 2022-08-14
Update requirements.txt

davidbuzz:


davidbuzz:
wheel is needed for lxml and pymavlink under most circumstances.

khancyr (approved):
Yep, make things easier

PR #728, opened 2022-09-12
Disable mavnative by default

khancyr:
Disable  mavnative building and loading by default.
Update docs to reflet how to use mavnative now.

Tested  with ArduPilot SITL and mavproxy and mavexplorer.


peterbarker:
Hmm.  I thought there was a proposal to kill mavnative entirely?  No mavlink2 support and all.

khancyr:
I can also do the full kill. That is even simplier.
Notice that mavnative depreaction isnot only for the lack of mavlink2 support : the C code doesn't match the python side anymore as we put some fix on python but not on mavnative.

Here I just put mavnative as not build nor run by default but keep the config  to  build and run it  if asked  for.

tridge:
I think we should kill mavnative completely

khancyr:
Ok I will update the pr to remove it completely!

khancyr:
@tridge @peterbarker removal done.
I have only keep `use_native=False` in MAVLink class to provide backward compatibility. Otherwise, all those have use this in the past will get and error on updating.

generator/mavgen_python.py
```diff
@@ -795,18 +765,7 @@ def parse_char(self, c):
...
-                m = self.__parse_char_native(self.test_buf)
-                m2 = self.__parse_char_legacy()
-                if m2 != m:
-                    logger.error("Native: %s\\nLegacy: %s\\n", m, m2)
-                    raise Exception("Native vs. Legacy mismatch")
-            else:
-                m = self.__parse_char_native(self.buf)
-        else:
-            m = self.__parse_char_legacy()
+        m = self.__parse_char_legacy()
```
alehed, line 768: Thanks for this PR!

Not sure if it makes sense to still call it `__parse_char_legacy`, instead of renaming it to something like `__parse_char` or `__parse_char_internal`.

README.md
```diff
@@ -78,12 +77,7 @@ sudo python -m pip install --upgrade pymavlink
 
 #### Mavnative
 
-By default, pymavlink will try to compile and install mavnative which is a C extension for parsing mavlink. Mavnative only supports mavlink1.
-To skip mavnative installation and reduce dependencies like `gcc` and `python-dev`, you can pass `DISABLE_MAVNATIVE=True` environment variable to the installation command:
-
-```bash
-sudo DISABLE_MAVNATIVE=True python -m pip install --upgrade pymavlink
-```
+Starting from September 2022, mavnative, a C extension for parsing mavlink, was deprecated and removed. Mavnative developpement was stalled for long time, it only supports MAVLink1 and doesn't get any fix on the protocol.
```
peterbarker, line 80:
```suggestion
Starting from September 2022, mavnative, a C extension for parsing mavlink, was deprecated and removed. Mavnative development was stalled for long time, it only supports MAVLink1 and doesn't get any fix on the protocol.
```

peterbarker (approved):

PR #737, opened 2022-10-07
DFReader: extract defaults from 4.3.x logs

tridge:


DFReader.py
```diff
@@ -666,6 +666,14 @@ def _add_msg(self, m):
             self.flightmode = mavutil.mode_string_px4(m.MainState)
         if type == 'PARM' and getattr(m, 'Name', None) is not None:
             self.params[m.Name] = m.Value
+            if hasattr(m,'Default'):
+                if not hasattr(self,'param_defaults'):
+                    self.param_defaults = {}
+                if math.isnan(m.Default):
+                    if not m.Name in self.param_defaults:
+                        self.param_defaults[m.Name] = m.Value
```
IamPete1, line 674: NaN just means no value, not value is defualt. We only add the defaults on the initial param dump where we walk the table, the ones with NaN are from subsequent saves where its not worth searching the table to get the default.

PR #744, opened 2022-10-26
[Java] Use longs for values too big to store in int

abroekhof:


abroekhof:
This fixes #638

shancock884:
Note that this PR appears to be duplicated by #678. I believe this (or 678) would fix #825.

peterbarker:
Merged, thanks!

PR #746, opened 2022-11-03
cope with recent changes in param_id string handling

### LINKED ISSUE #745: string issues
issues from the recent changes to string handling:

mavparmdiff.py using b'XXXX'
image

NAMED_VALUE_FLOAT string indexes show as b'XXXX'
image

tridge:
ping @alehed 
fixes #745

### DIFF
--- a/mavparm.py
+++ b/mavparm.py
@@ -61,7 +61,7 @@ class MAVParmDict(dict):
                 if ack is None:
                     time.sleep(0.1)
                     continue
-                if str(name).upper() == str(ack.param_id).upper():
+                if str(name) == ack.param_id.decode('ascii'):
                     got_ack = True
                     self.__setitem__(name, float(value))
                     break
--- a/mavutil.py
+++ b/mavutil.py
@@ -94,6 +94,9 @@ def add_message(messages, mtype, msg):
         messages[mtype] = msg
         return
     instance_value = getattr(msg, msg._instance_field)
+    if isinstance(instance_value, bytes):
+        # cope with string instance values, such as NAMED_VALUE_FLOAT
+        instance_value = instance_value.decode(errors="backslashreplace").rstrip("\\x00")
     if not mtype in messages:
         messages[mtype] = copy.copy(msg)
         messages[mtype]._instances = {}
@@ -432,7 +435,7 @@ class mavfile(object):
         elif type == 'PARAM_VALUE':
             if not src_tuple in self.param_state:
                 self.param_state[src_tuple] = param_state()
-            self.param_state[src_tuple].params[msg.param_id] = msg.param_value
+            self.param_state[src_tuple].params[msg.param_id.decode('ascii')] = msg.param_value
         elif type == 'SYS_STATUS' and mavlink.WIRE_PROTOCOL_VERSION == '0.9':
             self.flightmode = mode_string_v09(msg)
         elif type == 'GPS_RAW':
--- a/tools/AccelSearch.py
+++ b/tools/AccelSearch.py
@@ -48,7 +48,7 @@ def AccelSearch(filename):
                 logcount += 1
             return False
         if m.get_type() == 'PARAM_VALUE':
-            if m.param_id.startswith('INS_PRODUCT_ID'):
+            if m.param_id.decode('ascii').startswith('INS_PRODUCT_ID'):
                 if m.param_value not in [0.0, 5.0]:
                     return False
         if m.get_type() == 'RAW_IMU':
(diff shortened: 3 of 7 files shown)


tools/AccelSearch.py
```diff
@@ -48,7 +48,7 @@ def AccelSearch(filename):
                 logcount += 1
             return False
         if m.get_type() == 'PARAM_VALUE':
-            if m.param_id.startswith('INS_PRODUCT_ID'):
+            if m.param_id.decode('ascii').startswith('INS_PRODUCT_ID'):
```
alehed, line 51: Here I would do `m.param_id.startswith(b'INS_PRODUCT_ID')`. But semantically it is the same.

tools/MPU6KSearch.py
```diff
@@ -82,7 +82,8 @@ def IMUCheckFail(filename):
                     print('UNCALIBRATED: %s' % m)
                     return False
         elif mtype == 'PARAM_VALUE':
-            if m.param_id.startswith('INS_ACCOFFS_') or m.param_id.startswith('INS_ACC2OFFS_'):
+            param_id = m.param_id.decode('ascii')
+            if param_id.startswith('INS_ACCOFFS_') or param_id.startswith('INS_ACC2OFFS_'):
```
alehed, line 86: Same here

tools/magfit_motors.py
```diff
@@ -93,7 +93,7 @@ def magfit(logfile):
         m = mlog.recv_match(condition=args.condition)
         if m is None:
             break
-        if m.get_type() == "PARAM_VALUE" and m.param_id == "RC3_MIN":
+        if m.get_type() == "PARAM_VALUE" and m.param_id.decode('ascii') == "RC3_MIN":
```
alehed, line 96: Here I would also use a bytes literal.

tools/magfit_rotation_gyro.py
```diff
@@ -129,7 +129,7 @@ def magfit(logfile):
                 name = str(m.Name)
                 value = int(m.Value)
             else:
-                name = str(m.param_id)
+                name = m.param_id.decode('ascii')
```
alehed, line 132: Since this is for display, you probably want to also set set errors argument to ignore.

tools/mavparms.py
```diff
@@ -32,7 +32,7 @@ def mavparms(logfile):
         except Exception:
             return
         if m.get_type() == 'PARAM_VALUE':
-            pname = str(m.param_id).strip()
+            pname = m.param_id.decode('ascii')
```
alehed, line 35: Same here.

mavparm.py
```diff
@@ -61,7 +61,7 @@ def mavset(self, mav, name, value, retries=3, parm_type=None):
                 if ack is None:
                     time.sleep(0.1)
                     continue
-                if str(name).upper() == str(ack.param_id).upper():
+                if str(name) == ack.param_id.decode('ascii'):
```
alehed, line 64: name probably doesn't need the str(). So I would do `if name.upper() == ack.param_id.decode('ascii', errors='ignore').upper()`

mavutil.py
```diff
@@ -432,7 +435,7 @@ def post_message(self, msg):
         elif type == 'PARAM_VALUE':
             if not src_tuple in self.param_state:
                 self.param_state[src_tuple] = param_state()
-            self.param_state[src_tuple].params[msg.param_id] = msg.param_value
+            self.param_state[src_tuple].params[msg.param_id.decode('ascii')] = msg.param_value
```
alehed, line 438: Probably it makes sense to also add the error here or index the params by bytes internally.

alehed (commented):
Sorry for the breakage. In general I would always assume that decoding can fail.

mavutil.py
alehed: Also note that in python2 decode produces a unicode value not a string value, so this might also cause potential issues.

mavparm.py
alehed: This should work in python2 because `'123' == u'123'` produces `True`.

tools/mavparms.py
alehed: You can also add the strip in the end, unless that change was intentional.

tridge:
@alehed we're going to need to change back to producing string for char[] arrays in the XML. I'm just hitting far too many errors, hit several more today. Changing char[] from being a string to being bytes is too much of an API breakage
If you want bytes, declare a uint8_t[] array

alehed:
Ok, I'll make another PR.

alehed:
I will probably make the PR sometime this weekend. The tricky bit is going to handle python2/3 compatibility properly.

tridge:
> I will probably make the PR sometime this weekend

thanks!

> The tricky bit is going to handle python2/3 compatibility properly.

yes, quite tricky. thanks for offering to look at this!

alehed:
See #747.

PR #747, opened 2022-11-04
Use strings for char array mavlink fields again

alehed:
This should fix the backward compatibility issue with the type annotation PR.

### DIFF
--- a/generator/mavgen_python.py
+++ b/generator/mavgen_python.py
@@ -554,15 +554,21 @@ def generate_classes(outf, msgs, enable_type_annotations):
 
         init_fields = []
         for f in m.fields:
-            init_fields.append("self.%s = %s" % (f.name, f.name))
+            if f.type == "char":
+                init_fields.append("self._%s_raw = %s" % (f.name, f.name))
+                init_fields.append('self.%s = %s.decode("ascii", errors="replace")' % (f.name, f.name))
+            else:
+                init_fields.append("self.%s = %s" % (f.name, f.name))
 
         pack_fields = []
         for field in m.ordered_fields:
-            if field.type != "char" and field.array_length > 1:
+            if field.type == "char":
+                pack_fields.append("self._{0:s}_raw".format(field.name))
+            elif field.array_length == 0:
+                pack_fields.append("self.{0:s}".format(field.name))
+            else:
                 for i in range(field.array_length):
                     pack_fields.append("self.{0:s}[{1:d}]".format(field.name, i))
-            else:
-                pack_fields.append("self.{0:s}".format(field.name))
 
         t.write(
             outf,
--- a/generator/mavgen_python.py
+++ b/generator/mavgen_python.py
@@ -556,7 +556,7 @@ def generate_classes(outf, msgs, enable_type_annotations):
         for f in m.fields:
             if f.type == "char":
                 init_fields.append("self._%s_raw = %s" % (f.name, f.name))
-                init_fields.append('self.%s = %s.decode("ascii", errors="replace")' % (f.name, f.name))
+                init_fields.append('self.%s = %s.split(b"\\x00", 1)[0].decode("ascii", errors="replace")' % (f.name, f.name))
             else:
                 init_fields.append("self.%s = %s" % (f.name, f.name))
 


alehed:
Should fix https://github.com/ArduPilot/pymavlink/issues/745

WickedShell:
I haven't investigated the python code fully here, but have you checked what happens when you get a char array from the aircraft that contains invalid ASCII, or invalid UTF-8? I've seen some interesting crashes out of Java recently relating to this.

alehed:
Due to the fact that we are decoding them as utf-8, all we can do here is raise a MAVError. The official answer for having non-string data in a message is using an array of uint8_t (see https://github.com/ArduPilot/pymavlink/pull/746#issuecomment-1302800170).

tridge:
@alehed thanks! I've installed and will test today

> Due to the fact that we are decoding them as utf-8, all we can do here is raise a MAVError

it would be better if we could do backslashreplace or similar, but giving a MAVError is better than breaking the existing API. I thought it would just be a small number of places to fix, but I found that too many tools had errors, and realised that there will be too many users of pymavlink with their own code that would break
I think we need a test case for this to see what the old code did with bad UTF-8 and to ensure whatever we decide on is locked in

tridge:
found first issue:
```
RTL> param set INS_GYRO_FILTER 30
RTL> ERROR in command ['set', 'INS_GYRO_FILTER', '30']: 'bytes' object has no attribute 'encode'
  File "/home/tridge/.local/lib/python3.10/site-packages/MAVProxy-1.8.59-py3.10.egg/EGG-INFO/scripts/mavproxy.py", line 768, in process_stdin
  File "/home/tridge/.local/lib/python3.10/site-packages/MAVProxy-1.8.59-py3.10.egg/MAVProxy/modules/mavproxy_param.py", line 647, in cmd_param
    self.pstate[sysid].handle_command(self.master, self.mpstate, args)
  File "/home/tridge/.local/lib/python3.10/site-packages/MAVProxy-1.8.59-py3.10.egg/MAVProxy/modules/mavproxy_param.py", line 381, in handle_command
    self.mav_param.mavset(master, uname, value, retries=3, parm_type=ptype)
  File "/home/tridge/.local/lib/python3.10/site-packages/pymavlink-2.4.37-py3.10.egg/pymavlink/mavparm.py", line 57, in mavset
    mav.param_set_send(name.upper(), vfloat, parm_type=parm_type)
  File "/home/tridge/.local/lib/python3.10/site-packages/pymavlink-2.4.37-py3.10.egg/pymavlink/mavutil.py", line 581, in param_set_send
    self.mav.param_set_send(self.target_system, self.target_component,
  File "/home/tridge/.local/lib/python3.10/site-packages/pymavlink-2.4.37-py3.10.egg/pymavlink/dialects/v20/ardupilotmega.py", line 22190, in param_set_send
    self.send(self.param_set_encode(target_system, target_component, param_id, param_value, param_type), force_mavlink1=force_mavlink1)
  File "/home/tridge/.local/lib/python3.10/site-packages/pymavlink-2.4.37-py3.10.egg/pymavlink/dialects/v20/ardupilotmega.py", line 19441, in send
    buf = mavmsg.pack(self, force_mavlink1=force_mavlink1)
  File "/home/tridge/.local/lib/python3.10/site-packages/pymavlink-2.4.37-py3.10.egg/pymavlink/dialects/v20/ardupilotmega.py", line 9744, in pack
    return self._pack(mav, self.crc_extra, self.unpacker.pack(self.param_value, self.target_system, self.target_component, self.param_id.encode(), self.param_type), force_mavlink1=force_mavlink1)
AttributeError: 'bytes' object has no attribute 'encode'
```

alehed:
> it would be better if we could do backslashreplace or similar,

On second though decoding to utf-8 here was probably too liberal. When a message is said to contain 2 ints, and 20 chars, returning fewer (in case non-ascii characters are in the string, or we choose errors=ignore) or more characters (if doing backslashreplace) than specified is probably going to cause issues. So there I would either decode as ascii and raise Maverrors if we don't get ascii, or decode as ascii and use `errors="replace"` so at least indexing still works properly.

> found first issue:

We're in a bit of a pickle here. The thing is that the old behavior was to decode into a string on receiving the message such that a consumer of the message would get a python string.

Now that means that if you took a received message that had some chars inside and tried to send it again, it would raise an exception because you can't give strings to `struct.pack`. The reason this issue probably never came up is that messages are either sent via the `send_` helper methods or received via `parse_char()`.

So I guess what you had to do is deal with strings when you received a message, but pass the XXX_send helper method a `bytes` for the same field. What I could do to maintain backwards compatibility is decode the arguments to strings in the XXX_send methods. I'm not sure about the XXX_encode method though that is called from send. I'm assuming the backwards compatible thing would be to also accept bytes there.

Another, maybe less nasty way, of doing things would be to store both the string and the bytes value in the message (the bytes being prefixed with `_` and maybe suffixed with `_raw`). Then encoding and decoding would do the correct thing and the user can still see the string value if necessary by accessing the attribute.

alehed:
@tridge can you give it another spin? I like the fix better now and it should be more compatible now (when creating a message object manually now it raises an error if a string is passed in, but I don't expect this to be the case in any existing codebase).

alehed:
In theory I could even change the signature of the initializer of the mavlink message subclass to take either a string or a bytes, but I would like to avoid that if possible as the semantics would be slightly different and probably confusing.

tridge:
@alehed I've installed this PR plus #748 and am actively testing. Good so far, will update you on EU dev call this evening if you're around
thanks!

PR #751, opened 2022-11-15
fix_instance_field

### LINKED ISSUE #750: _instance_field doesn't exists
https://github.com/ArduPilot/pymavlink/blob/850ef26117159808e7c8421600b21eb8c1b101cc/mavutil.py#L95

In some cases, a received mavlink message object doesn't contain the _instance_field member.
I believe the quoted code snippet was meant to deal with such a situation (by using the getattr function) but not properly (first there's an attempt to access the _instance_field member, and then in the second part of the if statement, the second argument of getattr is the attribute itself instead of the attribute name).

OhadNir9:
Fixes #750

### DIFF
--- a/mavutil.py
+++ b/mavutil.py
@@ -92,11 +92,11 @@ class location(object):
 
 def add_message(messages, mtype, msg):
     '''add a msg to array of messages, taking account of instance messages'''
-    if msg._instance_field is None or getattr(msg, msg._instance_field, None) is None:
+    if getattr(msg, '_instance_field', None) is None:
         # simple case, no instance field
         messages[mtype] = msg
         return
-    instance_value = getattr(msg, msg._instance_field)
+    instance_value = getattr(msg, '_instance_field')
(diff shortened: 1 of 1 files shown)


mavutil.py
```diff
@@ -92,11 +92,11 @@ def __str__(self):
 
 def add_message(messages, mtype, msg):
     '''add a msg to array of messages, taking account of instance messages'''
-    if msg._instance_field is None or getattr(msg, msg._instance_field, None) is None:
+    if getattr(msg, '_instance_field', None) is None:
         # simple case, no instance field
         messages[mtype] = msg
         return
-    instance_value = getattr(msg, msg._instance_field)
+    instance_value = getattr(msg, '_instance_field')
```
tridge, line 99: this breaks instance fields. Say for example msg._instance_field is "FOO" we want msg.FOO not msg._instance_field

tridge (changes requested):

PR #752, opened 2022-11-23
generator/C: fix node test in CI

julianoes:
The node test in mavlink/mavlink CI fails because of:

```
standard/testsuite.h:29:13: error: ‘mavlink_test_standard’ defined but not used [-Werror=unused-function]

   29 | static void mavlink_test_standard(uint8_t system_id, ...
      |             ^~~~~~~~~~~~~~~~~~~~~
cc1: all warnings being treated as errors
```

Ignoring the warning fixes CI.

See as an example:
https://github.com/mavlink/mavlink/actions/runs/3504071299/jobs/5869516184

FYI @davidbuzz, @hamishwillee, @auturgy


davidbuzz:
Seems reasonable.

davidbuzz (approved):

julianoes:
Thanks @davidbuzz and @tridge!

PR #698, opened 2022-06-24
added reserved words check in mavgen_cpp11.py

Macavitycode:
Appends a '_' while generating in case of reserved words.
This is a workaround to the issue posted in this issue: [mavlink/mavlink#1853](https://github.com/mavlink/mavlink/issues/1853)

PR #761, opened 2022-12-19
fix EOF error for cpp11

spectre-solutions:
Fix issue: https://github.com/ArduPilot/pymavlink/issues/762

Changes identical to [this fix](https://github.com/mavlink/mavlink-gbp-release/blob/master/patch/pymavlink/generator/mavgen_cpp11.py#L44-L47)

Related to PR: https://github.com/ArduPilot/pymavlink/pull/698
We prefer solution from this PR, but both works and both are OK.


spectre-solutions:
Thanks @tridge

PR #763, opened 2022-12-28
gpslag: cope with GPS instances

tridge:


peterbarker (approved):
Commit message is a bit misleading in that we're also losing support for `GPS2`.  I don't think that's a big issue.

PR #773, opened 2023-02-02
tools: use instancing for MAG msg in onboard logs

robertlong13:
The onboard messages `MAG` and `MAG2` have been replaced with a single `MAG` message with an instance variable. Code that was intended to read from the primary mag was reading from all mags, and code that was intended to read from `MAG2` wasn't working.

I have only tested magfit_WMM.py, but while I was there, decided to search for any other outdated usages of `MAG`

### DIFF
--- a/tools/magfit.py
+++ b/tools/magfit.py
@@ -105,14 +105,15 @@ def magfit(logfile):
             mag = Vector3(m.xmag, m.ymag, m.zmag)
             # add data point after subtracting the current offsets
             data.append(mag - offsets + noise())
-        if m.get_type() == "MAG" and not args.mag2:
-            offsets = Vector3(m.OfsX,m.OfsY,m.OfsZ)
-            mag = Vector3(m.MagX,m.MagY,m.MagZ)
-            data.append(mag - offsets + noise())
-        if m.get_type() == "MAG2" and args.mag2:
-            offsets = Vector3(m.OfsX,m.OfsY,m.OfsZ)
-            mag = Vector3(m.MagX,m.MagY,m.MagZ)
-            data.append(mag - offsets + noise())
+        if m.get_type() == "MAG":
+            if m.I==0 and not args.mag2:
+                offsets = Vector3(m.OfsX,m.OfsY,m.OfsZ)
+                mag = Vector3(m.MagX,m.MagY,m.MagZ)
+                data.append(mag - offsets + noise())
+            elif m.I==1 and args.mag2:
+                offsets = Vector3(m.OfsX,m.OfsY,m.OfsZ)
+                mag = Vector3(m.MagX,m.MagY,m.MagZ)
+                data.append(mag - offsets + noise())
 
     print("Extracted %u data points" % len(data))
     print("Current offsets: %s" % offsets)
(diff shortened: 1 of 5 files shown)


peterbarker:
Thanks @robertlong13 .  Do you think we should instead remove this as there's a copy in MAVProxy which is in much more common use?  It knows about the different mags already: https://github.com/ardupilot/MAVProxy/blob/master/MAVProxy/modules/lib/magfit.py#L1

tridge (approved):

robertlong13:
@peterbarker Sorry for the delay in the reply. I find myself preferring this script, partly because I don't use MAVExplorer much, and this feels more lightweight; and partly because it has the ability to add filter options (correct me if I'm wrong, but I don't think the alternative offers that. If we would rather unify, perhaps I should open a PR to add that)

tridge:
@robertlong13 the big advantage of using MAVExplorer is it only uses the samples from the region of the last graph zoom, which is great for selecting the right part of a flight

robertlong13:
Ahhh, I missed that. I will have to check that out again

WickedShell:
I tried using the magfit_WMM.py from here, and I'm having trouble getting it to approximate the results I'm seeing with the tool in MAVExplorer. I like having the script as a backup, at least since it seems to actually read the data in, before creating a bad result. I've also had some issues in the past with using MAVExplorer on machines that are missing graphics, or are suffering with python graphics bindings (this has caused me a number of issues in the past randomly).

Anyways specifically I'm seeing this script produce this result:

```
New: Vector3(-295.49, 102.82, 20.07) diag: Vector3(0.80, 0.80, 0.89) offdiag: Vector3(-0.10, -0.06, -0.15) cmot_mode: 0 cmot: Vector3(0.00, 0.00, 0.00) scale: 1.00
COMPASS_OFS_X -295
COMPASS_OFS_Y 102
COMPASS_OFS_Z 20
COMPASS_DIA_X 0.800
COMPASS_DIA_Y 0.800
COMPASS_DIA_Z 0.888
COMPASS_ODI_X -0.102
COMPASS_ODI_Y -0.058
COMPASS_ODI_Z -0.149
COMPASS_MOT_X 0.000
COMPASS_MOT_Y 0.000
COMPASS_MOT_Z 0.000
COMPASS_SCALE 1.00
COMPASS_MOTCT 0
```

while MAVExplorer generates:

```
New: Vector3(-175.51, 135.39, 51.77) diag: Vector3(0.92, 1.11, 1.03) offdiag: Vector3(0.13, 0.01, -0.03) cmot: Vector3(0.00, 0.00, 0.00) scale: 1.00
COMPASS_OFS_X -175
COMPASS_OFS_Y 135
COMPASS_OFS_Z 51
COMPASS_DIA_X 0.916
COMPASS_DIA_Y 1.111
COMPASS_DIA_Z 1.035
COMPASS_ODI_X 0.128
COMPASS_ODI_Y 0.006
COMPASS_ODI_Z -0.026
COMPASS_MOT_X 0.000
COMPASS_MOT_Y 0.000
COMPASS_MOT_Z 0.000
COMPASS_SCALE 1.00
```

andyp1per:
I too am getting weird results

PR #774, opened 2023-02-02
generator-C: added last_status field for signing

tridge:
this allows the caller to know why signing is being rejected, to provide better feedback to the user


peterbarker (approved):
LGTM

peterbarker:
Could potentially add an "MAVLINK_SIGNING_STATUS_UNKNOWN" at 0 and then make "OK a different number

PR #779, opened 2023-02-16
mavutil: Return mode as 'Mode(msg.custom_mode)' for high-latency streams

dernatsch:
Before, when reading high-latency messages that came from PX4, mavutil threw an exception because after the check for PX4, `msg.base_mode` was read immediately. It was not regarded that the `HIGH_LATENCY2` message does not contain a `base_mode` field.

This introduces a fix where the mode string for _all_ `HIGH_LATENCY2` messages is returned as `Mode(msg.custom_mode)`.

### DIFF
--- a/mavutil.py
+++ b/mavutil.py
@@ -2273,7 +2273,11 @@ def mode_mapping_bynumber(mav_type):
 def mode_string_v10(msg):
     '''mode string for 1.0 protocol, from heartbeat'''
     if msg.autopilot == mavlink.MAV_AUTOPILOT_PX4:
+        if msg.get_type() == "HIGH_LATENCY2":
+            return "Mode(%u)" % msg.custom_mode
+
         return interpret_px4_mode(msg.base_mode, msg.custom_mode)
+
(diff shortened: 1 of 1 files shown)


peterbarker:
This also looks like we will no longer map the base mode to a string on ArduPilot.

Could you put your two new lines just within the PX4 bloc there, please - just stop us reading those attributes within that block if the message is a `HIGH_LATENCY2`

Apologies for the breakage - we really should have picked this up.

@stephendade FYI

dernatsch:
@peterbarker Just changed it. Is this what you had in mind? :)

stephendade:
FYI in ArduPilot, we do send the ``base_mode`` in the ``custom0`` field of ``HIGH_LATNENCY2``, as we need ``base_mode`` to determine if the vehicle's armed.

peterbarker (approved):
LGTM

peterbarker (changes requested):
Ah, sorry - still have  the f-string :-/

dernatsch:
@peterbarker No worries, changed it to the way it's done at the bottom of the function. Hope it's alright now. :)

PR #788, opened 2023-03-30
tools: print progress in magfit_WMM

kd0aij:
adds a per-iteration callback for the least-squares routine

### DIFF
--- a/tools/magfit_WMM.py
+++ b/tools/magfit_WMM.py
@@ -181,6 +181,9 @@ def wmm_error(p):
 
     return ret
 
+def printProgress(x):
+    print("offsets:", x[0:3], " scale:", x[3])
+
 def fit_WWW():
     from scipy import optimize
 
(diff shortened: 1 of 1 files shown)


tools/magfit_WMM.py
```diff
@@ -181,6 +181,9 @@ def wmm_error(p):
 
     return ret
 
+def printProgress(x):
+    print("offsets:", x[0:3], " scale:", x[3])
```
peterbarker, line 185: This isn't the way one usually prints stuff in Python now-adays :-)
```suggestion
    print(f"offsets: {x[0:3]} scale: {x[3]}")
```
would be the modern way

peterbarker (approved):

peterbarker:
Merged, thanks!

peterbarker:
On Thu, 13 Apr 2023, Mark Whitehorn wrote:

> thanks, I remember seeing that format once... does pymavlink now require Python >= 3.6?

ish :-)

We're not longer afraid of breaking things, I think.

I wasn't suggesting you change this, BTW.

PR #791, opened 2023-04-05
Add support for relative path to generated lua mavlink module

bugobliterator:


rotu:
What's the deal with this PR? It looks like it was cherry-picked in? https://github.com/ArduPilot/pymavlink/commit/aba3c41422a04466a0edcc287589cfc8ab29ffd3

Also, while we're at it, why do both Lua and WLua exist, and why is Lua omitted from the tests?

https://github.com/ArduPilot/pymavlink/blob/e4eb7fab82a3a342f98535cbf168bf5ed82174f2/test_generate_all.sh

bugobliterator:
> What's the deal with this PR? It looks like it was cherry-picked in? [aba3c41](https://github.com/ArduPilot/pymavlink/commit/aba3c41422a04466a0edcc287589cfc8ab29ffd3)
> 
> Also, while we're at it, why do both Lua and WLua exist, and why is Lua omitted from the tests?
> 
> https://github.com/ArduPilot/pymavlink/blob/e4eb7fab82a3a342f98535cbf168bf5ed82174f2/test_generate_all.sh

This is for ardupilot lua script generation. The test for the generation is done in ardupilot

rotu:
> This is for ardupilot lua script generation. The Lua needs to be added to that list.

Makes sense. Probably could also document how to use the generated files and remove this misleading comment:
https://github.com/bugobliterator/pymavlink/blob/2e0d4d0d121cf3d7eee716c1b454218bb54dd4b6/generator/mavgen_lua.py#LL3C1-L3C73

PR #797, opened 2023-04-21
Added slots

stryngs:
Slots is a great way to increase efficiency without an in-depth code change.  The only caution I would have in accepting this is the following line of thought with regards to those who may try to write a wrapper and create an object within the Class that wasn't slotted ahead of time.

```
class Bob(object):
    __slots__ = ('foo')
    def __init__(self):
        self.foo = 1
        
b = Bob()
```

```
In [1]: %run ./test.py

In [2]: b.bar = 2
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-2-9fce950889ff> in <module>
----> 1 b.bar = 2

AttributeError: 'Bob' object has no attribute 'bar'
```
If this is accepted I will continue submitting until everything that can be slotted is.

### DIFF
--- a/mavutil.py
+++ b/mavutil.py
@@ -81,6 +81,7 @@ def u_ord(c):
 
 class location(object):
     '''represent a GPS coordinate'''
+    __slots__ = ('lat', 'lng', 'alt', 'heading')
     def __init__(self, lat, lng, alt=0, heading=0):
         self.lat = lat  # in degrees
         self.lng = lng  # in degrees
@@ -148,6 +149,8 @@ set_dialect(os.environ['MAVLINK_DIALECT'])
 
 class mavfile_state(object):
     '''state for a particular system id'''
+    __slots__ = ('messages', 'flightmode', 'vehicle_type', 'mav_type',
+                 'mav_autopilot', 'base_mode', 'armed')
     def __init__(self):
         self.messages = { 'MAV' : self }
         self.flightmode = "UNKNOWN"
@@ -191,7 +194,7 @@ class mavfile(object):
         # param state for each sysid/compid tuple
         self.param_state = {}
         self.param_state[self.param_sysid] = param_state()
-        
+
         # status of param fetch, indexed by sysid,compid tuple
         self.source_system = source_system
         self.source_component = source_component
@@ -218,7 +221,7 @@ class mavfile(object):
     @property
     def target_component(self):
         return self.param_sysid[1]
-    
+
     @target_system.setter
     def target_system(self, value):
         self.sysid = value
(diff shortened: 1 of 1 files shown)


Gone4Dirt:
This is interesting, was not aware of slots! learnt a thing :-)
From reading about it the main benefit to this that it is more memory efficient.  @stryngs do you agree or is there something else to it?

stryngs:
> This is interesting, was not aware of slots! learnt a thing :-) From reading about it the main benefit to this that it is more memory efficient. @stryngs do you agree or is there something else to it?

There is a speed boost to be gained as well.  In a way you give up some of the dynamic capabilities Python3 offers in exchange for a faster outcome.

khancyr:
speed boost is going to be negligeable in most case unless you have a really big dict. So the real gain from this will be on message usage I think.
For the use in this PR, that will provide a bit of type safety but the speed will be totally negligeable. 
I can be wrong, but until I see some real data show speed boost, I will trust Python docs on this !

davidbuzz:
I like slots, i'm in favour,if that help sway anyone.

peterbarker:
On Sun, 30 Apr 2023, Buzz wrote:

> I like slots, i'm in favour,if that help sway anyone.

Big questions to be answered
  - does adding slots to each message save time/memory?
  - does it kill any users of pymavlink that we know of?
   - e.g. dronekit-python
   - e.g. MAVProxy
   - e.g. pymavlink itself :-)
   - e.g. the ArduPilot autotest suite?

stryngs:
slots is a different way of thinking.  The biggest straw to chew on would be to know how many projects out there customize the existing libraries.  When you add slots to the mix the dynamic nature for that given class is gone.  You can do workarounds to where this is not the case but then you're reducing the effectiveness of it and thus I wouldn't.

All a matter of choice as anyone who would be writing wrappers is probably capable of modifying the core Class to their needs, etc.

My experience with slots has been positive but I am limited to my worldview where I sniff packets.  I write in Python and so any speed increase I can get is a want.  It will "always?" lag behind a compiled language.   When you add pypy3 to the mix Python starts to get really really snappy and it keeps up with Enterprise loads.

I have no idea if slots is the right fit for this project but if it is wanted I will happily do the PRs.

I look forward to the decision =)

davidbuzz:
great summary of what implications slots have here:  https://wiki.python.org/moin/UsingSlots 
key takeaways:
"slots are more efficient in terms of memory space and speed of access, and a bit safer than the default Python method of data access."
"a __slots__ declaration uses less memory than a dictionary, and direct memory access is faster than dictionary lookups"
"__slots__ is a class variable. If you have more than one instance of your class, any change made to __slots__ will show up in every instance."

"you can add '__dict__' or '__weakref__' as the last element in the __slots__ declaration" ...to enable ... "your class to use dynamic attribute creation or weak references".

PR #803, opened 2023-04-26
add PID tuning based on Bayesian optimization

zeng-hq:
Add a Bayesian optimization program in examples that attempts to minimize the maximum angular velocity of a copter.


tridge (changes requested):
very interesting project!
I tried this on the builtin SITL quadcopter and ended up getting bad oscillation, with very high D gains and high slew rates. I think this would need a few changes, even as an example:
 - a big warning that this is highly experimental
 - a way to abort out and restore original PIDs

You may find it is better to implement this as a MAVProxy module, which allows for more interaction with the user. We can accept it here as an example if it has a big warning on it though.

PR #808, opened 2023-05-04
WLUA - add new port for WSL 18570

hamishwillee:
This adds a new dissector port for 18570. This is the port that is exposed for PX4 running in WSL2.
I'm not sure if it more widely defined than that, or why it works and not the other ports.


peterbarker:
Merged :-)

hamishwillee:
Thanks very much. Should make bisecting the other error easier :-)

PR #809, opened 2023-05-04
mavschema.xsd: add superseded tag

hamishwillee:
This adds a `superseded` tag. Once this is in. my intent is to divide the existing deprecated tags into those that are actually deprecated by the standard (and we hope to remove) and those that are just superseded by a new message, command, enum that we hope one day everyone will upgrade too.

The difference being that superseded is as much a hint to autopilots as users that there is something that might we worth upgrading to. 

This is a precondition to any kind of release process.

### DIFF
--- a/generator/mavschema.xsd
+++ b/generator/mavschema.xsd
@@ -62,14 +62,14 @@
     </xs:restriction>
   </xs:simpleType>
 </xs:attribute>
-<xs:attribute name="since"> <!-- used in deprecated elements -->
+<xs:attribute name="since"> <!-- used in elements: wip, superseded, deprecated -->
   <xs:simpleType>
     <xs:restriction base="xs:string">
-      <xs:pattern value="(20)\d{2}-(0[1-9]|1[012])"/>  <!-- restrict dates to 20YY-MM format -->
+      <xs:pattern value="(20)\d{2}-(0[1-9]|1[012])"/> <!-- restrict dates to 20YY-MM format -->
     </xs:restriction>
   </xs:simpleType>
 </xs:attribute>
-<xs:attribute name="replaced_by" type="xs:string"/> <!-- used in deprecated elements -->
+<xs:attribute name="replaced_by" type="xs:string"/> <!-- used in elements: superseded, deprecated -->
 
 <!-- mavlink message IDs are unsigned 24-bit values. -->
 <xs:simpleType name="mavlinkMsgId" id="mavlinkMsgId">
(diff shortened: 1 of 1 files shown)


generator/mavschema.xsd
```diff
@@ -202,7 +202,16 @@
     </xs:complexType>
 </xs:element>
 
-<xs:element name="deprecated">
+<xs:element name="wip">
```
hamishwillee, line 205: FYI, just a reorder for wip and deprecate, which is what makes this a bit messier. Sorry.

hamishwillee:
@auturgy @peterbarker Can we get some movement on this? ONly a year and a half late ....

Plan is to move current deprecated to superseded unless they are "really" deprecated. I think that is a little less confusing.

peterbarker:
@hamishwillee could you provide some concrete examples on where this would be applied, please?

Is this part of a life-cycle of some description?  So do things move from superseded to deprecated, for example?

Do we outline a life-cycle anywhere?  That would be wip -> <no tag> -> superseded -> deprecated -> removed, I think with this current proposal?

hamishwillee:
Hi @peterbarker 

The intent here is to reduce confusion in the naming so that we can move towards a lifecycle. Currently we use the term deprecated to mean both:
- `superseded`: there is a newer way of doing this same thing we'd like flight stacks to implement, and users to use instead of the older alternative. Does not mean that it is implemented or the old version is not in use.
- `deprecated`: this is something you should not use because it is planned for removal.

So yes, the intent is to have a lifecycle. We don't outline the lifecycle because it is a mess.

Yes, my thinking was: wip -> "active" -> superseded -> deprecated -> removed.
- Though something might be deprecated without being superseded.

It isn't perfect, because it doesn't help a user know whether anything can be used - i.e. I come to superseded, navigate to the new API and discover it isn't implemented anywhere. It does however help us know what we'd like flight stacks to be moving towards and deprecating.

Does that make sense?

hamishwillee:
> could you provide some concrete examples on where this would be applied, please?

Initially I would move everything that is deprecated to superseded in one PR. Then we would discuss which are really deprecated.

I'd be doing some testing and propose.

peterbarker:
> > could you provide some concrete examples on where this would be applied, please?
> 
> Initially I would move everything that is deprecated to superseded in one PR. Then we would discuss which are really deprecated.

I was more after specific messages.

I'd love to get rid of messages which are ancient and crap!  These aren't zero-cost, they all push out the storage costs and have upkeep overhead.

> But there is no point working out what can really be deprecated if there is no way of me marking it.

Yep.

hamishwillee:
> I was more after specific messages.

Yes I know. OK, then I will mark every single message that is currently deprecated as superseded :-).

The ones that will then become deprecated would be anything that is removed by all current ardupilot/PX4 versions back a few years. I think `MISSION_ITEM` might be a good candidate. Essentially though we'll need some criteria and then move each individually by PR using the normal process.

Once you merge this we can move forward.

julianoes (approved):

PR #799, opened 2023-04-24
Better ci2

khancyr:
This add pre-commit as part of CI.
this tool allow to set check easy and add them localy on git-hook for better coding and not wait CI to trigger failure.

Correct Line Ending mixing on files : we should have only LF
Correct permission on files

Add test_local_change.sh to help people to use pre-commit.

update from https://github.com/ArduPilot/pymavlink/pull/679

PR #810, opened 2023-05-04
Commit line ending changes required to make checkouts clean

peterbarker:
This just changes line endings.

We recently had a commit which says that all checkouts must have the correct line endings, so fresh checkouts always have these files modified.

Thanks to @hamishwillee for reporting


hamishwillee (approved):
I don't know if I have approver rights here, but if so, this kind of thing is non-harmful, and as far as I can tell the code hasn't changed other than the line endings

khancyr (approved):
good to go.
They will be fixed as part of https://github.com/ArduPilot/pymavlink/pull/799 too, and checked on CI

PR #816, opened 2023-05-14
Replace validator with xmlschema (XSD1.1 compliant)

hamishwillee:
The current validator is XSD 1.0 compliant. That means you can validate for order of elements, and number and type of elements. 
I'd like to update to an XSD 1.1 compliant validator, as XSD 1.1 allows validation using patterns, and inter-element validation. For example, with XSD 1.0 you can require that a command has 7 parameters, and that their index is between 1 and 7, but with XSD 1.1 you can also require that they be unique.

My main interest at this point is to add range checking - allowing us to validate in XSD the allow range(s) for the command and message IDs, and to add exclusion ranges so that we can reserve items more obviously inside the allocated ranges.

The only Python XSD1.1 validate I can find is this one: https://github.com/sissaschool/xmlschema

This update is pretty much a direct replacement. Right now it drops one feature - the ability to turn off strict checking of the types. I've raised a question here to see if there is a way to do this: https://github.com/sissaschool/xmlschema/issues/346

In many ways I prefer the existing parser, but there is no indication if or when it will get v1.1 parsing. Note that the existing parser is capable of doing the same kind of tests that I want, but only in Python, not in XSD. That is not really OK because we want others to be able to validate in their own programming languages if they so wish - so best to embed in the XSD file.

### DIFF
--- a/generator/mavgen.py
+++ b/generator/mavgen.py
@@ -249,12 +249,6 @@ def mavgen(opts, args):
                 for fieldItem in messageItem.iter('field'): #
                     if forbidden_names_re.search(fieldItem.get('name')):
                         xmlvalid = False
-                        #print(f"messageItem value: {entryItem.get('value')}")
-                        #print(f"messageItem name: {entryItem.get('name')}")
-                        #print(f"Fielditem name: {fieldItem.attrib}")
-                        #print(f"Fielditem name: {fieldItem.text}")
-                        #print(f"Entry name: {fieldItem.tostring(root, encoding='utf8').decode('utf8')}")
-                        #TODO: Print out the whole field?
                         raise Exception(
                             "Validation Error: Message : "
                             f"{messageItem.get('name')} ({messageItem.get('id')}) "
--- a/generator/mavgen.py
+++ b/generator/mavgen.py
@@ -31,6 +31,10 @@ import re
 import sys
 from . import mavparse
 
+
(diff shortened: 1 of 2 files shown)


khancyr:
I would said that should be ok. But we need to ensure that Windows can use the new lib and we will need to put a python3.7+ restriction on pymavlink.
That can be an issue depending on the project using it . for example ArduPilot still require python3.6+ so that can have some issue. Even if 3.6 is already from the past.

hamishwillee:
Thanks. I've been running it on Windows (only) so far. Good catch on Python restriction. I still want to see if I can turn off strict checking. Apparently the right way to do this is a second XSD file. If I can do it elegantly I will.

hamishwillee:
Ubuntu 18.04 is running Python 3.6, which might be a blocker

julianoes:
Python 3.6 is end of life, so not even security supported anymore, since December 2021 (see https://devguide.python.org/versions/#unsupported-versions).

My strong opinion is to remove it from CI here and focus any efforts on what is actually maintained. If we try to maintain each and every old version it creates a burden for maintainers and hence slows everything down.

The only use case of Python 3.6 would be someone building MAVLink directly on a Jetson on Ubuntu 18.04 with stock Python. I would say to ignore that one. Worst case Python can be updated manually in that case.

Issue #556, opened 2021-06-18
Drop support for Python 2


cclauss:
Python 2 died on more than 500 days ago on 1/1/2020 and should be considered an insecure platform.  Our installation and development processes are both more complicated because we still support legacy Python.

Current Python is v3.9.
https://devguide.python.org/#status-of-python-branches v.s.
https://devguide.python.org/devcycle/#end-of-life-branches

hamishwillee:
@auturgy @peterbarker Didn't you say we had dropped support for Python 2? I think we should formalise that across the project, starting by removing the Python 2.7 checks from CI.

What can I do to make this happen?

While we're at it, would also be good to drop Python 3.6, but that's another story.

hamishwillee:
Our customers and those of the Python core team are different, and have different lifecycles. Our usage of python is not one where a million % increase in performance would make all that much difference. So I don't think we can draw the lines quite that hard.

But you're right that once the older versions start costing us to maintain, then a decision has to be made about whether those old versions are worth the cost of carrying. Definitely not for Python 2.7, but and hopefully not for Python 3.6. I've pinged people offline so I am sure this will get properly discussed at various dev calls in the coming weeks.

cclauss:
Your customers are doing more on their computers than just running pymavlink.  If they do not care about performance then perhaps they care about safety.  Running software that no longer receives security fixes can often mean that they are working on an insecure platform.
https://www.cvedetails.com/vulnerability-list.php?vendor_id=10210&product_id=18230

hamishwillee:
They care about safety, but for the most part old Python versions do not affect that in the drone domain. The code on the flight stack isn't running Python - we just use it as part of the build toolchain. Some drones with companion computers will be running Python to send messages, but that code isn't really exposed to the same kinds of attacks that most people worry about - say in websites.

Maybe I'm wrong. But since I want this to happen I'm concentrating on the thing that I know (from experience) is most likely to affect the outcome.

rotu:
GitHub CI just deprecated python 2 so our CI workflow is now failing: https://github.com/actions/setup-python/issues/672

rotu:
Here's some things I noticed. Some of this might be in scope for #829.

1. There is some documentation that needs to be updated to remove support for python 2:
   - https://mavlink.io/en/guide/scripts.html 
[...]

PR #817, opened 2023-05-15
Fix Wireshark dissector and add snapshot tests

### LINKED ISSUE #726: Wireshark dissect bit fields fails
Most notably this affects HEARTBEAT dissection.

looks like this change https://github.com/ArduPilot/pymavlink/commit/4a8b143c071ee2e937bba36411ba1f8fb9c1a7ab
broke the bit field dissection in wireshark.
The line 
   _subtree, value = tree:add_le(f.${fvar}, padded(field_offset, ${fbytes}))_
in generated lua does not actually return the value, it returns nil instead.
here is a generated example
    _subtree, value = tree:add_le(f.HEARTBEAT_base_mode, padded(field_offset, 1))_

As a work around I explicitly extracted the value to pass to the generated dissect_flags function which resolved the issue.
in mavgen_wlua.py, function generate_field_dissector. line 319

```
        if flag_enum is not None:
            t.write(outf,
"""
    value = padded(field_offset, ${fbytes}):uint()
    dissect_flags_${enumname}(subtree, "${fvar}", value)
""", {'enumname': flag_enum.name, 'fvar': field_var, 'fbytes': size})

```

Of course this is assuming all bit fields are an uint

### LINKED ISSUE #815: Wireshark Wlua does not dissect flags
An error is shown if you try to inspect a message, such as HEARTBEAT, that has flags, and you can't inspect any fields after that flag. For example in HEARTBEAT this is the mode flag:

There are two issues. Firstly, the `value` returned by `tree:add_le` is `nil`, so what gets passed into `dissect_flags_MAV_MODE_FLAG` cannot be dissected. The docs say `tree:add_le` should return a value. My assumption is that this is failing for the reason given after the code. Note that this is the only place the value is actually used, which is why we don't see the error elsewhere.

```
    field_offset = offset + 6
    subtree, value = tree:add_le(f.HEARTBEAT_base_mode, padded(field_offset, 1))
    dissect_flags_MAV_MODE_FLAG(subtree, "HEARTBEAT_base_mode", value)
```

I "fixed" this error by passing in `padded(field_offset, 1)` to `dissect_flags_MAV_MODE_FLAG` instead of `value`. Wireshark now fails with the error that the data passed in is "userData" - i.e. an arbitrary C data that it can't interpret. That's correct, and probably why `nil` got returned from the `tree:add_le`.

The solution would be to cast the data to a number before passing it to `tree:add_le`. HOwever I don't know how to do this elegantly.

This is cross posted (and summarised) from https://github.com/mavlink/mavlink/issues/1978

rotu:
Handle bit flags in Wireshark dissector (fix #815, fix #726)
Add tests for Wireshark (via `tshark` binary and `pytest` via the `syrupy` snapshot library)
Remove defunct mav2pcap.py script


hamishwillee:
@rotu Tried this out. Certainly fixes the error we were seeing. I particularly like how clear the bitflags are rendered:

image

I also looked at many other message, including commands, and nothing seems broken:

image

Excellent to see test code of this. Would be good to expand it in future as a validation test. Out of scope for this probably.

The generator code changes make sense to me, and I think this also makes the generated code much more clear. 

However I am not expert. Once you have this to a state you are happy with, we'll need to  work with @peterbarker to get this in.

rotu:
I'm not an expert either. Thanks for the screenshots! I think they highlight some other bugs!

The second message you show is a [`COMMAND_INT`](https://mavlink.io/en/messages/common.html#COMMAND_INT), where param2 is a float.

The value shown is `1065353216` which I think is supposed to be the floating point value `1.0f`, corresponding to `1`; i.e. `MAV_DO_REPOSITION_FLAGS_CHANGE_MODE: True`.

Also the longitude/latitude values are payload parameters 5/6 so I think they should be `int32_t`, not `float`, hence the wonky-looking values.

Could you please attach some pcap data I can test against?

hamishwillee:
Looks like you're right. I ran this for a few seconds and then did a save-as. Is this what you needed?

[pcapdata.zip](https://github.com/ArduPilot/pymavlink/files/11548904/pcapdata.zip)

hamishwillee:
PS Also a CI error

rotu:
I see the CI errors and but the alternative is to use the [obsolete `pkg_resources`](https://setuptools.pypa.io/en/latest/pkg_resources.html) and I hate writing new legacy code. When are we dropping python 2.7 and 3.6 support per #556?

hamishwillee:
@rotu Could not agree more. 
PX4 have dropped support for Python 2, and I understand that ArduPilot either have done so, or will be doing so very shortly. If those dependencies are gone, we should be able to switch over "very soon". I'll push for that, so ignore the error for now. 
Wr.t. dropping Python 3.6 that is harder - we are in discussion.

As I understand it you're looking at the other issues with the generated output. Since I'm very happy for you to do so I'm going to just keep quiet now :-)

hamishwillee:
@rotu This isn't waiting on me for anything is it?

hamishwillee:
Hi @rotu 
Looks like you are still progressing. Let me know when you need more review.

rotu:
@hamishwillee Yep. I'm chipping away slowly. I'm a little hung up on getting the output to match when there are long lines (not sure how to get the output of tshark to wrap to a consistent line length)

rotu:
Alright I'm sufficiently sick of this PR, so it's time to get it out there!

hamishwillee:
@rotu I can imagine. Time for me to get a technical review on this?

rotu:
> @rotu I can imagine. Time for me to get a technical review on this?

I don't understand what you mean. Do whatever you need, loop in whoever you like, and ask whatever questions to satisfy your curiosity.

hamishwillee:
@peterbarker Can you review this please. I think it is important to get in.

I have done basic tests on it by running the instructions in https://mavlink.io/en/guide/wireshark.html and also specifically looking at the HEARTBEAT message to check that the known issue was fixed. 
This definitely fixes the original problem in a logic way, and the resulting output is also good.

@rotu Has also added a heap of test code. I have not tested that, since I have no merge rights, and you will have to do this anyway.

requirements.txt
```diff
@@ -1,3 +1,7 @@
 lxml>=3.6.0
 future>=0.15.2
 wheel>=0.37.1
+
+# dev dependencies:
+pytest
```
peterbarker, line 6: Should we add these dependencies for everyone on every install of pymavlink?

The vast majority of people won't want to run the test suite.

Perhaps we should add this as a requirements.txt in the `snapshottests` directory?

rotu: > Should we add these dependencies for everyone on every install of pymavlink?

No, and this PR doesn't. When you `pip install` pymavlink, you *don't* get the stuff in requirements.txt. You get the (totally unversioned :shrug:) dependencies in `setup.py`.

hamishwillee: @peterbarker Response please.

hamishwillee: @peterbarker Ping.

peterbarker: Cool, thanks.

peterbarker (changes requested):
@rotu could you fix the commit list, please?  Should be a single commit

rotu:
> ... should have said I'm happy enough to merge this with a good commit list.

@peterbarker
Feel free to squash merge. https://docs.github.com/en/repositories/configuring-branches-and-merges-in-your-repository/configuring-pull-request-merges/configuring-commit-squashing-for-pull-requests

hamishwillee:
Hurray. Thanks!

PR #822, opened 2023-05-29
Check for messages[mtype]._instances is None

### LINKED ISSUE #807: Exception in mavutil.py after bad_data multi-instance message
Summary:

For multi-instance message types, e.g., https://mavlink.io/en/messages/common.html#DISTANCE_SENSOR, a call to `add_message` with a `bad_data` message will fail to initialize the `messages[mtype]._instances` dictionary. The next good message after that will cause an exception.

Traceback:

```
$ mavlogdump.py --quiet --types DISTANCE_SENSOR ~/Desktop/sa_tests/2023_04_18_surftrak3/2023-04-18\ 11-03-34.tlog
  File "/home/clyde/.local/bin/mavlogdump.py", line 242, in <module>
    m = mlog.recv_match(blocking=args.follow, type=match_types)
  File "/home/clyde/.local/lib/python3.10/site-packages/pymavlink/mavutil.py", line 1633, in recv_match
    m = self.recv_msg()
  File "/home/clyde/.local/lib/python3.10/site-packages/pymavlink/mavutil.py", line 481, in recv_msg
    self.post_message(msg)
  File "/home/clyde/.local/lib/python3.10/site-packages/pymavlink/mavutil.py", line 1436, in post_message
    super(mavlogfile, self).post_message(msg)
  File "/home/clyde/.local/lib/python3.10/site-packages/pymavlink/mavutil.py", line 378, in post_message
    add_message(self.sysid_state[src_system].messages, type, msg)
  File "/home/clyde/.local/lib/python3.10/site-packages/pymavlink/mavutil.py", line 99, in add_message
    messages[mtype]._instances[instance_value] = msg
TypeError: 'NoneType' object does not support item assignment
```

A possible fix is to [modify the if statement](https://github.com/ArduPilot/pymavlink/blob/v2.4.37/mavutil.py#L93):
```
if mtype not in messages or messages[mtype]._instances is None:
```

Reference this conversation:
https://discuss.bluerobotics.com/t/mavlogdump-py-crashes-while-reading-distance-sensor-msgs-in-some-tlog-files/14228

clydemcqueen:
This is a small change to fix #807. I've tested this against quite a few tlog files over the past 2 weeks and it works great. Running pytest comes up clean.

See the longer discussion in https://discuss.bluerobotics.com/t/mavlogdump-py-crashes-while-reading-distance-sensor-msgs-in-some-tlog-files/14228

Thanks,
/Clyde

### DIFF
--- a/mavutil.py
+++ b/mavutil.py
@@ -2538,40 +2538,41 @@ px4_map = { "MANUAL":        (mavlink.MAV_MODE_FLAG_CUSTOM_MODE_ENABLED | mavlin
 
 
 def interpret_px4_mode(base_mode, custom_mode):
+    # Dispatch on custom_main_mode (authoritative per PX4's px4_custom_mode.h);
+    # base_mode is no longer reliable for mode identity on PX4 v1.12+ (#793).
+    del base_mode
     custom_main_mode = (custom_mode & 0xFF0000)   >> 16
     custom_sub_mode  = (custom_mode & 0xFF000000) >> 24
 
-    if base_mode & mavlink.MAV_MODE_FLAG_MANUAL_INPUT_ENABLED != 0: #manual modes
-        if custom_main_mode == PX4_CUSTOM_MAIN_MODE_MANUAL:
-            return "MANUAL"
-        elif custom_main_mode == PX4_CUSTOM_MAIN_MODE_ACRO:
-            return "ACRO"
-        elif custom_main_mode == PX4_CUSTOM_MAIN_MODE_RATTITUDE:
-            return "RATTITUDE"
-        elif custom_main_mode == PX4_CUSTOM_MAIN_MODE_STABILIZED:
-            return "STABILIZED"
-        elif custom_main_mode == PX4_CUSTOM_MAIN_MODE_ALTCTL:
-            return "ALTCTL"
-        elif custom_main_mode == PX4_CUSTOM_MAIN_MODE_POSCTL:
-            return "POSCTL"
-    elif (base_mode & auto_mode_flags) == auto_mode_flags: #auto modes
-        if custom_main_mode & PX4_CUSTOM_MAIN_MODE_AUTO != 0:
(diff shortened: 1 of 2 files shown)


mavutil.py
```diff
@@ -97,7 +97,7 @@ def add_message(messages, mtype, msg):
         messages[mtype] = msg
         return
     instance_value = getattr(msg, msg._instance_field)
-    if not mtype in messages:
+    if mtype not in messages or messages[mtype]._instances is None:
```
peterbarker, line 100: Why not check for `msg.get_type() == 'BAD_DATA'` at the top of this method, and not do silly things with invalid data?

clydemcqueen: Thanks for the review! @peterbarker

That fixes the bug, and pytest succeeds. But has the side effect of never adding BAD_DATA messages to mavfile sysid_state.messages, which is a slight change to behavior. I looked for all usages in this repo, and I don't see any problems, and in fact I imagine it might fix an obscure class of bugs. If you are OK w/ this I'll make the change.

clydemcqueen: I made this change and rebased. Tests pass locally. Thanks.

tridge (approved):

mavutil.py
```diff
@@ -92,6 +92,9 @@ def __str__(self):
 
 def add_message(messages, mtype, msg):
     '''add a msg to array of messages, taking account of instance messages'''
+    if msg.get_type() == 'BAD_DATA':
+        # avoid dealing with instance field
```
peterbarker: Have you made sure that `BAD_DATA` still appears in MAVProxy's `messages` command after this change?

clydemcqueen: Hi @peterbarker -- I added the stack trace and some notes to the linked issue.

Yeah, I didn't test the "bad comms" case. I will need to rig something up to test this.

I need to spend more time thinking about the appropriate fix. Feel free to close this PR.

Happy New Year!

PR #824, opened 2023-06-07
Add altitudeOnly attribute

hamishwillee:
This adds an attribute in order to indicate a command has an altitude in param 7, but not a lat/lon in param 5/6. It also adds comments to note exactly what the other attributes mean (i.e. a lat/lon in param 3, 4 would not "count").

This allows more precise documentation about issues with using a altitude-only command.

This comes out of MAV call discussion summarised here: https://github.com/mavlink/mavlink/pull/1994#issuecomment-1581629423 . There was thought of modifying these to be hasLatLon, and hasAltitude, but decided to reduce churn.

### DIFF
--- a/generator/mavschema.xsd
+++ b/generator/mavschema.xsd
@@ -197,7 +197,7 @@
 <xs:attribute name="reserved" type="xs:boolean" default="false"/>      <!-- parameter is reserved -->
 
 <!-- definition entry elements attributes (like the ones used on MAV_CMD for example) -->
-<xs:attribute name="hasLocation" type="xs:boolean" default="false"/>    <!-- entry is an ROI or other non-destination (lat/lon/alt location in params 5, 6, 7). -->
+<xs:attribute name="hasLocation" type="xs:boolean" default="false"/> <!-- lat/lon/alt location in params 5, 6, 7 (entry contains location information). -->
 <xs:attribute name="isDestination" type="xs:boolean" default="false"/>  <!-- entry is a waypoint/destination (lat/lon/alt in params 5, 6, 7) -->
 <xs:attribute name="hasAltitudeOnly" type="xs:boolean" default="false"/>  <!-- alt in param7 and lat/lon not in param 5/6 -->
 <xs:attribute name="missionOnly" type="xs:boolean" default="false"/>  <!-- entry only makes sense in missions, not commands -->
--- a/generator/mavschema.xsd
+++ b/generator/mavschema.xsd
@@ -197,8 +197,9 @@
 <xs:attribute name="reserved" type="xs:boolean" default="false"/>      <!-- parameter is reserved -->
 
 <!-- definition entry elements attributes (like the ones used on MAV_CMD for example) -->
-<xs:attribute name="hasLocation" type="xs:boolean" default="true"/>    <!-- entry has lat/lon/alt location -->
-<xs:attribute name="isDestination" type="xs:boolean" default="true"/>  <!-- entry is a destination -->
+<xs:attribute name="hasLocation" type="xs:boolean" default="true"/>    <!-- entry is an ROI or other non-destination (lat/lon/alt location in params 5, 6, 7). -->
+<xs:attribute name="isDestination" type="xs:boolean" default="true"/>  <!-- entry is a waypoint/destination (lat/lon/alt in params 5, 6, 7) -->
+<xs:attribute name="altitudeOnly" type="xs:boolean" default="true"/>  <!-- entry contains an altitude but no lat/lon (alt in param 7) -->
 <xs:attribute name="missionOnly" type="xs:boolean" default="false"/>  <!-- entry only makes sense in missions, not commands -->
 
 <!-- definition of complex elements -->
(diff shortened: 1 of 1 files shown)


hamishwillee:
I've decided to close this as "not all that useful". The reason being that while there are commands where we might like to indicate altitudeOnly, they use inconsistent params for the altitude. Given that, there is no way a GCS can reliably use the information. We could revisit, but we'd do so next time we create another altitude command (and suggest we put it in same location as others so this becomes useful.

hamishwillee:
@peterbarker This is back because Don thinks it can be useful - see https://github.com/mavlink/mavlink/pull/1994#issuecomment-5259386300

- Adds XSD to allow  `altitudeOnly` in MAV_CMD XSD, indicating that param7 contains altitude information (and that param 5/6 do not have latitude.
- Switch this and isDestination/hasLocation to default `false` rather than default `true`. This means that if something isn't a location we don't have to apply the attribute (and that we can remove a bunch of these from the XML). 
  - Claude says this reflects the default in Pymavlink, so won't break anything. XML that is broken now will become more accurate.
  
  Corresponding change to XML for just the ones that use  `altitudeOnly` in https://github.com/mavlink/mavlink/pull/2566

generator/mavschema.xsd
```diff
@@ -287,9 +288,10 @@
         </xs:sequence>
         <xs:attribute ref="value" />
         <xs:attribute ref="name" use="required"/>
-        <xs:attribute ref="hasLocation"/>
+        <xs:attribute ref="hasLocation"/> <!-- should be choice with altitudeOnly, but can't enforce in XSD1.0 -->
+        <xs:attribute ref="altitudeOnly"/>
```
hamishwillee: Falls out of https://github.com/mavlink/mavlink/pull/2566#discussion_r3763487164

```suggestion
        <xs:attribute ref="hasAltitudeOnly"/>
```

generator/mavschema.xsd
```diff
@@ -287,9 +288,10 @@
...
         <xs:attribute ref="isDestination"/>
-        <xs:attribute ref="missionOnly"/>
+        <xs:attribute ref="missionOnly"/>
```
hamishwillee:
```suggestion
        <xs:attribute ref="missionOnly"/>
```

generator/mavschema.xsd
```diff
@@ -197,8 +197,9 @@
 <xs:attribute name="reserved" type="xs:boolean" default="false"/>      <!-- parameter is reserved -->
 
 <!-- definition entry elements attributes (like the ones used on MAV_CMD for example) -->
-<xs:attribute name="hasLocation" type="xs:boolean" default="true"/>    <!-- entry has lat/lon/alt location -->
-<xs:attribute name="isDestination" type="xs:boolean" default="true"/>  <!-- entry is a destination -->
+<xs:attribute name="hasLocation" type="xs:boolean" default="false"/>    <!-- entry is an ROI or other non-destination (lat/lon/alt location in params 5, 6, 7). -->
+<xs:attribute name="isDestination" type="xs:boolean" default="false"/>  <!-- entry is a waypoint/destination (lat/lon/alt in params 5, 6, 7) -->
+<xs:attribute name="altitudeOnly" type="xs:boolean" default="false"/>  <!-- entry contains an altitude but no lat/lon (alt in param 7) -->
```
hamishwillee:
```suggestion
<xs:attribute name="hasAltitudeOnly" type="xs:boolean" default="false"/>  <!-- alt in param7 and lat/lon not in param 5/6 -->
```

generator/mavschema.xsd
```diff
@@ -287,9 +288,10 @@
         </xs:sequence>
         <xs:attribute ref="value" />
         <xs:attribute ref="name" use="required"/>
-        <xs:attribute ref="hasLocation"/>
+        <xs:attribute ref="hasLocation"/> <!-- should be choice with altitudeOnly, but can't enforce in XSD1.0 -->
```
hamishwillee:
```suggestion
        <xs:attribute ref="hasLocation"/> <!-- Mutually exclusive with hasAltitudeOnly (can't enforce in XSD1.0) -->
```

generator/mavschema.xsd
```diff
@@ -197,8 +197,9 @@
 <xs:attribute name="reserved" type="xs:boolean" default="false"/>      <!-- parameter is reserved -->
 
 <!-- definition entry elements attributes (like the ones used on MAV_CMD for example) -->
-<xs:attribute name="hasLocation" type="xs:boolean" default="true"/>    <!-- entry has lat/lon/alt location -->
-<xs:attribute name="isDestination" type="xs:boolean" default="true"/>  <!-- entry is a destination -->
+<xs:attribute name="hasLocation" type="xs:boolean" default="false"/>    <!-- entry is an ROI or other non-destination (lat/lon/alt location in params 5, 6, 7). -->
```
hamishwillee: The param values is what is most important here.

```suggestion
<xs:attribute name="hasLocation" type="xs:boolean" default="false"/> <!-- lat/lon/alt location in params 5, 6, 7 (entry contains location information). -->
```

peterbarker (approved):

hamishwillee:
Thanks @peterbarker !!

PR #829, opened 2023-06-25
Test on the current version of Python

cclauss:
### These are the minimal steps required to run all tests successfully while adding Python 3.11.
https://docs.python.org/3.11/whatsnew/3.11.html
> Python 3.11 is between 10-60% faster than Python 3.10. On average, we measured a 1.25x speedup on the standard benchmark suite.

GitHub Actions no longer supports running tests on legacy Python as discussed at https://github.com/ArduPilot/pymavlink/issues/556#issuecomment-1605736268

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false  # don't cancel if a job from the matrix fails
       matrix:
-        python-version: ['2.7', '3.5', '3.6', '3.7', '3.8', '3.9', '3.10']
+        python-version: ['2.7', '3.5', '3.6', '3.7', '3.8', '3.9', '3.10', '3.11']
 
     steps:
       - uses: actions/checkout@v3
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -69,6 +69,7 @@ jobs:
             flake8 --ignore='W503,E203,E501,E741' --statistics dialects/
             mypy --config-file ./.github/workflows/mypy-generated.ini dialects/v10/*.py dialects/v20/*.py
           fi
+
       - name: Generate messages
         run: |
           ./test_generate_all.sh
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,7 +12,7 @@ jobs:
     strategy:
       fail-fast: false  # don't cancel if a job from the matrix fails
       matrix:
-        python-version: ['2.7', '3.5', '3.6', '3.7', '3.8', '3.9', '3.10', '3.11']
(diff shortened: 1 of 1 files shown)


rotu:
Can you please make it more explicit whether the intent is ending support for Python 2.7 or merely removing it from CI coverage? The title makes it unclear and the project readme (and code!) definitely makes it sound like Python2 is still in the project's support scope.

cclauss:
Done above.

rotu:
> Added above.

The title is "Test on the current version of Python"
You added the description "These are the minimal steps required to run all tests successfully while adding Python 3.11.".

Neither mention Python 2.7.

That leaves open the two questions:
1. "As a user, should I expect this project to still work under Python 2.7?"
2. "As a developer, what do I need to do differently to test things moving forward?"

cclauss:
> Neither mention Python 2.7.

Sorry.  Perhaps I stuttered… _Python 2 died 1,273 days ago on 1/1/2020._

I see the answer to both of those questions to be self evident.
1. As a user, I should use a version of Python that is supported by the Python Core Team.
2. As a programmer, I should ensure that my modifications pass the project’s automated tests and I should add more useful tests.

rotu:
> > Neither mention Python 2.7.
> 
> Sorry. Perhaps I stuttered… _Python 2 died 1,273 days ago on 1/1/2020._

I know it did.

The title of this PR buries the lede. This PR adds a Python version to CI and drops a version from CI. It is not clear why the change is being made - it could be because any of the following reasons:

1. Python 2.7 was already out of scope for this project and it was just never removed from CI
2. Python 2.7 is newly out of scope as of this PR
3. Python 2.7 is still in scope for this project and it's just out of scope for automated testing because of new changes to GitHub CI.

and that should be made clear in the title and description of this PR.

> I see the answer to both of those questions to be self evident.
>     1. As a user, I should use a version of Python that is supported by the Python Core Team.

Python 3.5 and 3.6 are also out of support. Why are these staying in?

>     2. As a programmer, I should ensure that my modifications pass the project’s automated tests and I should add more useful tests.

Then why not *fix* CI for Python 2.7? (to be clear, I don't think this should be done)

cclauss:
> The title of this PR buries the lede.

Legacy Python is not the _lede_ and it has not been for several years.
It’s further use is dangerous and other that, it is completely irrelevant.

Closing.

rotu:
> > The title of this PR buries the lede.
> 
> Legacy Python is not the _lede_ and it has not been so for several years. It’s further use is dangerous and other that, it is completely irrelevant.
> 
> Closing.

It was not my intent to discourage you. I'm sorry. I agree with your feelings about Python 2 and that we should immediately discontinue its use.

I'll continue your efforts in PR #831 and #832.

PR #830, opened 2023-06-26
Remove Python 2 and Python 3.5 supporting code

rotu:
Code cleanup in furtherance of #556


setup.py
```diff
@@ -104,7 +97,6 @@ def run(self):
                     'Intended Audience :: Science/Research',
                     'License :: OSI Approved :: GNU Lesser General Public License v3 (LGPLv3)',
                     'Operating System :: OS Independent',
-                    'Programming Language :: Python :: 2.7',
                     'Programming Language :: Python :: 3.6',
```
alehed, line 100: I would specify here that 3.10 as well as 3.11 are supported. Otherwise pip will refuse to install it on these platforms.

rotu: I absolutely should add 3.10 and 3.11 and I'll push that in a different PR.

I tried to install the published version of pymavlink on python 3.11.3 with pip 23.1.2 and it worked fine despite the classifiers, so I think these are informational only.

Thanks for the comment! In looking it up, I learned about the `setup(python_requires=...)` argument which *does* enforce versions via pip https://python3statement.org/practicalities/

cclauss:
Status?!?
* #556
* #831
* #1025

PR #831, opened 2023-06-27
Drop support for python 2.7 and python 3.5

rotu:
As of this commit, pymavlink requires Python 3.6 or newer.

With the recent breaking of setup-python on github ci, this is as good a time as any to drop nominal support for old Python versions. As of today, Python 3.7 is end-of-life, but Ardupilot maintains support for Ubuntu Bionic (18.04)., which uses Python 3.6.

See #556


.github/workflows/test.yml
```diff
@@ -58,17 +54,12 @@ jobs:
 
       - name: Install pymavlink
         run: |
-          python setup.py build install
+          pip install .
```
peterbarker, line 57: Isn't this supposed to always be `pip3`?

README.md
```diff
@@ -15,7 +15,7 @@ Examples can be found [in the repository](examples/) or in the [ArduSub book](ht
 
 # Installation 
 
-Pymavlink supports both Python 2 and Python 3.
+Pymavlink supports Python 3.
```
peterbarker, line 18:
```suggestion
Pymavlink only supports Python 3.
```

README.md
```diff
@@ -56,7 +53,7 @@ sudo apt-get install python3-numpy python3-pytest
 Using pip you can install the required dependencies for pymavlink :
 
 ```bash
-sudo python -m pip install --upgrade future lxml
```
peterbarker, line 59: I believe these were deliberately `sudo` to install them for the entire system rather than just the current user. 

*I* install just for the local user - but not everybody does that.

I don't see any reason to conflate other Py2-related changes in this PR with changing these instructions..

README.md
```diff
@@ -56,7 +53,7 @@ sudo apt-get install python3-numpy python3-pytest
...
+python -m pip install --upgrade future lxml
```
peterbarker, line 56: SHould be `python3`?

peterbarker (changes requested):

.github/workflows/test.yml
rotu: I meant to write `pip`. The distinction can be important when you have a side-by-side installation of `python2` and `python3` to use `pip` or `pip3`, but frankly part of the reason for this PR is to stop *thinking* about python2 vs python3.

`pip install` is the recommendation as per every package page on PyPI (e.g. see below the package title on https://pypi.org/project/pytest/ ).

README.md
rotu: Here's the warning pip gives when I install with sudo:

> WARNING: Running pip as the 'root' user can result in broken permissions and conflicting behaviour with the system package manager. It is recommended to use a virtual environment instead: https://pip.pypa.io/warnings/venv

It's a bad recommendation here to use `sudo pip install`. It should be either non-sudo or installed with the package manager as `apt install python3-lxml python3-future`. The reason this is important in this PR is if the user has multiple python environments (e.g. installed with `pyenv`), `sudo` is going to change the variable environment, likely causing the package to be installed in the wrong place.

README.md
rotu: All the script shebangs use `python` so if this needs `python3`, the whole package needs other widespread updates as well.

rotu:
closed since this contribution was not merged in a timely manner.

PR #832, opened 2023-06-27
Add support for Python 3.10, 3.11

rotu:
I wanted to add support for 3.12 but currently am having build problems with lxml


cclauss:
> having build problems with lxml

Try adding a step...
```
      - if: "contains(matrix.python-version, '-dev')"
        run: sudo apt-get install -y libxml2 libxslt-dev
```

rotu:
> > having build problems with lxml
> 
> Try adding a step...
> 
> ```
>       - if: "contains(matrix.python-version, '-dev')"
>         run: sudo apt-get install -y libxml2 libxslt-dev
> ```

I already [tried with libxml2-dev](https://github.com/ArduPilot/pymavlink/actions/runs/5396673491/jobs/9800564018) but here's my try with libxml:

https://github.com/rotu/pymavlink/actions/runs/5477114606/jobs/9975667201

I'll give it a few more shots but I think that:
1. lxml is broken
2. we shouldn't be manually installing transitive dependencies anyway - those should be built in to the the wheel (if installed with pip) or installed as package managers (if installing lxml with apt)

rotu:
> [numpy/numpy#23808](https://github.com/numpy/numpy/issues/23808)

Thanks! I wish I had seen your comment earlier :-).

At any rate this is blocked on our dependency on the `future` package https://github.com/rotu/pymavlink/actions/runs/5477845537/jobs/9977492531#step:10:45

cclauss:
You could try creating an empty file named `imp.py`

rotu:
> You could try creating an empty file named `imp.py`

Mocking up missing modules is something I'd like to avoid. Especially since the code that's entailing it exists only for compatibility across the python 2->3 transition!

cclauss:
Even Py2 and Py3.1 [`importlib`](had https://docs.python.org/2/library/importlib.html)

Just drop Py3.12...  It is still in beta and can be attacked in a separate PR.

julianoes:
`imp` is imported by `future`, right? Could we drop `future`?

peterbarker:
Looks like 3.10 is now supported in master, but not 3.11. 

Rebase required.

 @rotu do you intend to pursue this PR?

rotu:
@peterbarker nope

Issue #813, opened 2023-05-08
How to call mavgen?


julianoes:
I'm confused what the right way should be to call mavgen.

According to various docs, I would assume the canonical way should be to install pymavlink via pip:

```
python -m pip install pymavlink  
Defaulting to user installation because normal site-packages is not writeable
Collecting pymavlink
  Using cached pymavlink-2.4.38-py3-none-any.whl (11.3 MB)
Requirement already satisfied: lxml in /usr/lib/python3/dist-packages (from pymavlink) (4.8.0)
Requirement already satisfied: future in ./.local/lib/python3.10/site-packages (from pymavlink) (0.18.3)
Installing collected packages: pymavlink
Successfully installed pymavlink-2.4.38
```

And then run it like this:

```
python -m pymavlink.tools.mavgen
/usr/bin/python: Error while finding module specification for 'pymavlink.tools.mavgen' (ModuleNotFoundError: No module named 'pymavlink.tools')
```

However, that doesn't seem to work. It doesn work when pymavlink is in tree, like when in the mavlink repo where pymavlink is a submodule, however, I would challenge whether pymavlink really needs to be a submodule there or whether that could be simplified.

@peterbarker any hints? Am I using it wrong?

JonasVautherin:
> It doesn work when pymavlink is in tree

You mean it _does work_ in that case, right?

julianoes:
@JonasVautherin yes thanks. Fixed it.

julianoes:
Or what about adding tools as a package in `setup.py`, would that work?

```
diff --git a/setup.py b/setup.py
index 6ccec166..86c97638 100644
--- a/setup.py
+++ b/setup.py
@@ -128,6 +128,7 @@ setup (name = 'pymavlink',
                         'pymavlink'              : ['message_definitions/v*/*.xml']
                         },
        packages = ['pymavlink',
+                   'pymavlink.tools',
                    'pymavlink.generator',
                    'pymavlink.dialects',
                    'pymavlink.dialects.v10',
```

julianoes:
Yes, it is installed as a script, and if you have e.g. `~/.local/bin/mavgen.py` in your path, then that works, but that's generally not something that's easy to explain to users in documentation.

[...]

PR #835, opened 2023-07-05
autodiscover packages in setup.py

rotu:
Revert to using setuptools' default package discovery. This restores, e.g. `python -m pytest.tools.mavgen` to a working state.

Fixes regression reported in #813


julianoes:
Awesome, thanks @rotu for figuring this out.

rotu:
> Awesome, thanks @rotu for figuring this out.

Judging by CI, you patted me on the back too soon! Looking into it some more :-)

PR #847, opened 2023-08-03
Added an http proxy for gps

stryngs:
This code may be useful in situations where a companion computer is connected to the ArduPilot board and is able to run mavproxy with an output on the IP and port as defined in the runtime arguments.


tridge:
this uses a separate http request per GPS fix? Seems an odd way to get GPS in! What you say "external" GPS, you mean one that is connected to the companion computer?

tridge:
@stryngs have you got a flight log of using this?
I would expect a lot of timing jitter, and poor EKF performance

PR #861, opened 2023-10-03
mavgen_lua: Fix for mavlink string fields

EosBandi:
Encode/Decode functions did not process string fields in mavlink message definitions correctly
{ param_id, "<c", 16 } was expanded to <ccccccccccccccc instead of <c16

Same went to ardupilot with https://github.com/ArduPilot/ardupilot/pull/24889


IamPete1:
Fixed by https://github.com/ArduPilot/pymavlink/pull/868

Issue #862, opened 2023-10-05
tools/mavgen.py can silently fail if an enum definition is missing


nexton-winjeel:
Running `mavgen.py` in our CI tool can report success even if the generation fails. In this particular case, we had a message defined, but the associated enum was missing.

* Using `master` at: https://github.com/ArduPilot/pymavlink/commit/19e385ccb4b6cbe2d08a8689dfcc586389fe14ce
* Using Python v3.8

```
(venv) nexton@laptop:~/Repositories/mavlink$ python --version
Python 3.8.10
(venv) nexton@laptop:~/Repositories/mavlink$ ./pymavlink/tools/mavgen.py --output ./build --lang C --wire-protocol 2.0 message_definitions/v1.0/all.xml
Validating message_definitions/v1.0/all.xml
Parsing message_definitions/v1.0/all.xml
[...]
Parsing /home/nexton/Repositories/mavlink/message_definitions/v1.0/cubepilot.xml
Validating /home/nexton/Repositories/mavlink/message_definitions/v1.0/csAirLink.xml
Parsing /home/nexton/Repositories/mavlink/message_definitions/v1.0/csAirLink.xml
Merged enum MAV_CMD
[...]
Merged enum MAV_CMD
Enum ESC_FAILURE_FLAGS in ESC_INFO.failure_flags does not exist
(venv) nexton@laptop:~/Repositories/mavlink$ echo $?
0
```

nexton-winjeel:
Fixed in #863.

PR #863, opened 2023-10-05
tools: Add option to mavgen.py to return exit code if generation fails

nexton-winjeel:
Addresses #862.


nexton-winjeel:
FYI: @tridge, @peterbarker

peterbarker (approved):
LGTM

nexton-winjeel:
Fixed in #863.

PR #875, opened 2023-11-20
generators: js: fix sinon dependency to 15.2.0, add node tests

Williangalvani:
this fixes a node12 issue with optional chaining in Sinon lib. check [recent mavlink actions](https://github.com/ardupilot/mavlink/actions)

I'm adding the js tests here because it is weird that it only gets to fail on the other repo.


khancyr (approved):

peterbarker (approved):

davidbuzz:
lgtm

PR #885, opened 2023-12-07
Add multiplier value for GPS_STATUS.satellite_azimuth where 255 means 360deg

shancock884:
Relates to https://github.com/mavlink/mavlink/issues/2062, point 2.

> 2/ Satellite Azimuth unit doesn't consider scale factor
The satellite_azimuth field of the GPS_STATUS message has a unit of "deg", but the description says "0: 0 deg, 255: 360 deg".
To be consistent with other units, and allow automatic value decoding, I would propose that a specific unit is used here too.
e.g. "degS8" (meaning "Degree Scaled to 8 bit integer"), or "deg/360*255" or any sensible simplification thereof.
I would also question if 255 is really supposed to be 360deg - as this means a value of 0 and 255 represents the same angle.
Perhaps what was intended is that 255 represents 360-(360/256) = 358.59375 ?

This PR adds the "degS8" unit, and an associated comment to the mavschema.xsd file.

### DIFF
--- a/generator/mavschema.xsd
+++ b/generator/mavschema.xsd
@@ -178,6 +178,7 @@
 <xs:simpleType name="factor">
   <xs:restriction base="xs:string">
     <xs:enumeration value="1E-2"/>        <!-- actual value = stated value / 100 -->
+    <xs:enumeration value="360/255"/>     <!-- actual value = stated value * 360/255, as used for GPS_STATUS.satellite_azimuth -->
   </xs:restriction>
 </xs:simpleType>
 


shancock884:
Following merging of #894, this now proposes to add a multiplier value of 360/255 to be used by satellite_azimuth, rather than create a new not-a-real-unit.
I specified this literally as "360/255" rather than a value to be more accurate and readable, but happy to change if it doesn't seem right.
@peterbarker, @hamishwillee

generator/mavschema.xsd
```diff
@@ -178,6 +178,7 @@
 <xs:simpleType name="factor">
   <xs:restriction base="xs:string">
     <xs:enumeration value="1E-2"/>        <!-- actual value = stated value / 100 -->
+    <xs:enumeration value="360/255"/>     <!-- actual value = stated value * 360/255, as used for GPS_STATUS.satellite_azimuth -->
```
hamishwillee, line 181: @peterbarker How about
```suggestion
    <xs:enumeration value="360div255"/>     <!-- actual value = stated value * 360/255, as used for GPS_STATUS.satellite_azimuth -->
```

hamishwillee: Or to put it another way, I have no strong opinions, but good to put this in.

shancock884: Personally, I'd prefer "360/255" because, it is both:
- directly evaluatable in pretty much any programming language for when the actual value is required for maths.
- humanly understandable as text if printed as-is (e.g. web page, Wireshark dissector, code comments, etc).

hamishwillee: Let's see what @peterbarker says - I dislike the slash sorry!

peterbarker: I like the slash.  It means we can continue to just eval these things.

peterbarker (approved):
Marked this for DevCall to make sure tridge is good with this.

amilcarlucas:
360/256 seams more logical
0° == 360°
8 LSBs of 0 bin == 8 LSBs of 256 bin

generator/mavschema.xsd
tridge: i prefer the slash

hamishwillee: Great!

hamishwillee: Thanks for pushing this @shancock884

shancock884:
> 360/256 seams more logical
> 0° == 360°
> 8 LSBs of 0 bin == 8 LSBs of 256 bin

yes I made the same suggestion when raising the original issue https://github.com/mavlink/mavlink/issues/2062, but in the end it seemed safer to keep the range in line with the historic description.

Issue #571, opened 2021-07-14
Problem with latest pymavlink - mavutil


mtbsteve:
When I import pymavlink with `from pymavlink import mavutil`
I am getting the following error:
```
Traceback (most recent call last):
  File "/home/apsync/ws/src/zed_yolo/nodes/zed_ros_to_mavlink.py", line 34, in <module>
    from pymavlink import mavutil
  File "/home/apsync/.local/lib/python3.7/site-packages/pymavlink-2.4.15-py3.7-linux-aarch64.egg/pymavlink/mavutil.py", line 2005, in <module>
    mavlink.MAV_TYPE_DECAROTOR:   mode_mapping_acm,
AttributeError: module 'pymavlink.dialects.v20.ardupilotmega' has no attribute 'MAV_TYPE_DECAROTOR'
```
I cloned the repro, and built it with `python3 setup.py install --user` from within the pymavlink directory. No errors or warnings during build.
Any idea anyone? Thanks!

peterbarker:
On Wed, 14 Jul 2021, mtbsteve wrote:

> AttributeError: module 'pymavlink.dialects.v20.ardupilotmega' has no attribute 'MAV_TYPE_DECAROTOR'

Update your mavlink definitions.  If that's not a possiblity, we can 
potentially work around this.

peterbarker:
@mtbsteve did you get this fixed?

mtbsteve:
Yes I managed to get it working- thanks!

tmendoza271:
> On Wed, 14 Jul 2021, mtbsteve wrote: AttributeError: module 'pymavlink.dialects.v20.ardupilotmega' has no attribute 'MAV_TYPE_DECAROTOR'
> Update your mavlink definitions. If that's not a possiblity, we can potentially work around this.

What do you mean by "update your mavlink definitions"? I'm having the same issue

marko-castral:
> > On Wed, 14 Jul 2021, mtbsteve wrote: AttributeError: module 'pymavlink.dialects.v20.ardupilotmega' has no attribute 'MAV_TYPE_DECAROTOR'
> > Update your mavlink definitions. If that's not a possiblity, we can potentially work around this.
> 
> @peterbarker What do you mean by "update your mavlink definitions"? I'm having the same issue

[...]

PR #889, opened 2023-12-15
Update deprecated MAV_TYPE enum for tailsitter

Dennis-Johnson:
Hi,

This fixes https://github.com/ArduPilot/pymavlink/issues/571.

With latest Pymavlink (master), the mavgen step works, but on trying to import mavutil, I get this error. 
Screenshot 2023-12-15 at 1 46 44 PM

It seems that these two enums were changed, but weren't reflected in mavutil.py
- MAV_TYPE_VTOL_TAILSITTER_QUADROTOR was previously MAV_TYPE_VTOL_QUADROTOR
- MAV_TYPE_VTOL_TAILSITTER_DUOROTOR was previously MAV_TYPE_VTOL_DUOROTOR

I believe there is still a reference to the old enum name in Swift generator tests ```pymavlink/generator/swift/Tests/MAVLinkTests/Testdata/common.xml```, but I haven't run that locally.

### DIFF
--- a/mavutil.py
+++ b/mavutil.py
@@ -2113,8 +2113,8 @@ AP_MAV_TYPE_MODE_MAP_DEFAULT = {
     mavlink.MAV_TYPE_COAXIAL:     mode_mapping_acm,
     # plane
     mavlink.MAV_TYPE_FIXED_WING: mode_mapping_apm,
-    mavlink.MAV_TYPE_VTOL_DUOROTOR: mode_mapping_apm,
-    mavlink.MAV_TYPE_VTOL_QUADROTOR: mode_mapping_apm,
+    mavlink.MAV_TYPE_VTOL_TAILSITTER_DUOROTOR: mode_mapping_apm,
+    mavlink.MAV_TYPE_VTOL_TAILSITTER_QUADROTOR: mode_mapping_apm,
     mavlink.MAV_TYPE_VTOL_TILTROTOR: mode_mapping_apm,
     # rover
     mavlink.MAV_TYPE_GROUND_ROVER: mode_mapping_rover,


peterbarker (commented):
This would break ArduPilot's CI as we have not adopted the name-change into `ardupilot/mavlink`.

@hamishwillee what process was gone through to ensure renaming this enumeration entry would be smooth?  i.e. `MAV_TYPE_VTOL_QUADROTOR` -> `MAV_TYPE_VTOL_TAILSITTER_QUADROTOR` ?


I'm not actually sure the rename is actually valid - the description does not precluse the use of `MAV_TYPE_VTOL_DUOROTOR` for belly-sitters as well as tailsitters.

hamishwillee:
So it was discussed with @auturgy @DonLakeFlyer @meee1 both in https://github.com/mavlink/mavlink/pull/1818 and also offline. The reasoning for the change is mostly here: https://github.com/mavlink/mavlink/pull/1808#issue-1157921604

> - the description does not precluse the use of MAV_TYPE_VTOL_DUOROTOR for belly-sitters as well as tailsitters.

W.r.t.  there is no such thing. Though note that I see https://mavlink.io/en/messages/minimal.html#MAV_TYPE_VTOL_TAILSITTER is misdocumented in the description to mention that removed concept. It should say [MAV_TYPE_VTOL_TAILSITTER_DUOROTOR](https://mavlink.io/en/messages/minimal.html#MAV_TYPE_VTOL_TAILSITTER_DUOROTOR). 

For a two rotor bellysitter you'd have to use MAV_TYPE_VTOL_FIXEDROTOR or a more specific MAV_TYPE_VTOL_FIXEDROTOR_DUO_ROTOR or similar. If that is invalid then yes, more thinking needed.

peterbarker:
I've created a PR with an alternative approach here: https://github.com/ArduPilot/pymavlink/pull/898

That will let mavutil.py work with both old and new message definitions.

Why has this only just now become a problem?  It looks like these renames happened quite some time ago, which should have led to the mavutil.py AttributeErrors?!?

hamishwillee:
So can this now be closed @peterbarker

PR #891, opened 2023-12-18
mavutil: Do not use unknown messages to calculate number of lost packets

joshanne:
Found this one when adding a custom mavlink message.
The decoding of the packet affects the mav_loss calculation because the unknown message does not have its header populated. 

To a user, this could appear as though the vehicle radio link is intermittent/bad.


tridge (approved):

tridge:
good find, thanks!

PR #856, opened 2023-09-22
Mavgen WLua: Add param units to display

shancock884:
The idea of this pull request is to add the field unit to the Wireshark display, whenever units are specified in the Mavlink dialect XML file. For example, output is as shown below, where the text such as "[us]", "[degE7]", etc is new:
```
Payload: GPS_RAW_INT (24)
    time_usec (uint64_t) [us]: 3625540000
    fix_type (GPS_FIX_TYPE): GPS_FIX_TYPE_3D_FIX (3)
    lat (int32_t) [degE7]: 514783466
    lon (int32_t) [degE7]: -25922214
    alt (int32_t) [mm]: 56900
    eph (uint16_t): 100
    epv (uint16_t): 100
    vel (uint16_t) [cm/s]: 0
    cog (uint16_t) [cdeg]: 0
    satellites_visible (uint8_t): 16
    alt_ellipsoid (int32_t) [mm]: 0
    h_acc (uint32_t) [mm]: 1000
    v_acc (uint32_t) [mm]: 1000
    vel_acc (uint32_t) [mm]: 200
    hdg_acc (uint32_t) [degE5]: 0
    yaw (uint16_t) [cdeg]: 0
``` 
Code lines changed is quite small, and I have done a number of manual checks on the output, but let me know if I need to do more...

PR #896, opened 2023-12-21
Mavgen WLua: Updates to displayed field values in Wireshark

shancock884:
As a follow up to #856, this PR proposes some updates to the decoding and displaying of Mavlink fields in Wireshark:

_- Convert some units into more readable values (appended to the original, so raw value is still shown), e.g.:_
```
lon (int32_t) [degE7]: -25922213 (-2.5922213 deg)
``` 
More converted units can be added, by adding to the unit_decoder_mapping lookup table.

_- Show value of bitmask fields in HEX_DEC format, and pad bitfields to a multiple of 4 to match hex, e.g.:_
```
flags (EKF_STATUS_FLAGS): 0x033f (831)
  .... .... ...1 = EKF_ATTITUDE: True
... (etc) ...
  .0.. .... .... = EKF_UNINITIALIZED: False
```

_- Append local date/time to μs fields if it represents a unix time, or display time in seconds, unless < 1s, e.g.:_
```
time_unix_usec (uint64_t) [us]: 1694287362375871 (2023-09-09 20:22:42.375871 BST)
time_usec (uint32_t) [us]: 937785569 (937.785569 s)
```

_- In header, show system id as DEC, and MAV_COMPONENT as enum, e.g.:_
```
System id: 1
Component id: MAV_COMP_ID_AUTOPILOT1 (1)
```

_- Decode command parameters on MISSION_ITEM messages, e.g.:_
```
param1: Land Options (NAV_VTOL_LAND_OPTIONS): NAV_VTOL_LAND_OPTIONS_DEFAULT (0)
param2 (float): 0
param3: Approach Altitude (float) m: 0
param4: Yaw (float) deg: 0
x: Latitude (int32_t): 514775152
y: Longitude (int32_t): -25939161
z: Ground Altitude (float) m: 0
```
Previously this was only done for COMMAND_LONG/INT messages, but not for MISSION_ITEM(_INT) messages.

_- Added 2 new automated test cases to demonstrate/check the unit decoding and MISSION_ITEM_INT fields._


peterbarker:
Merged, thanks!

PR #900, opened 2024-01-09
Mavdumplog: Handle TLog files saved with .log extension

### LINKED ISSUE #710: Mavlogdump fails on tlog with .log suffix
Mavlogdump fails on tlogs that have a `.log` suffix.

```
  File "/home/borgers/projects/MAVProxy/env/bin/mavlogdump.py", line 4, in <module>
    __import__('pkg_resources').run_script('pymavlink==2.4.31', 'mavlogdump.py')
  File "/home/borgers/projects/MAVProxy/env/lib/python3.10/site-packages/pkg_resources/__init__.py", line 651, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/home/borgers/projects/MAVProxy/env/lib/python3.10/site-packages/pkg_resources/__init__.py", line 1448, in run_script
    exec(code, namespace, namespace)
  File "/home/borgers/projects/MAVProxy/env/lib/python3.10/site-packages/pymavlink-2.4.31-py3.10-linux-x86_64.egg/EGG-INFO/scripts/mavlogdump.py", line 83, in <module>
    mlog = mavutil.mavlink_connection(filename, planner_format=args.planner,
  File "/home/borgers/projects/MAVProxy/env/lib/python3.10/site-packages/pymavlink-2.4.31-py3.10-linux-x86_64.egg/pymavlink/mavutil.py", line 1776, in mavlink_connection
    if DFReader.DFReader_is_text_log(device):
  File "/home/borgers/projects/MAVProxy/env/lib/python3.10/site-packages/pymavlink-2.4.31-py3.10-linux-x86_64.egg/pymavlink/DFReader.py", line 1045, in DFReader_is_text_log
    ret = (f.read(8000).find('FMT,') != -1)
  File "/usr/lib/python3.10/codecs.py", line 322, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe4 in position 2: invalid continuation byte
```

The `DFReader_is_text_log` [function](https://github.com/ArduPilot/pymavlink/blob/master/DFReader.py#L1041) opens the binary file as text file and tries to do UTF-8 decoding which does not work in case of a binary tlog file. This happens because we assume that files with suffix `.log` can be opened as text: https://github.com/ArduPilot/pymavlink/blob/master/mavutil.py#L1773

We should not make such assumptions.

One solution could be to wrap the opening of the file in  `DFReader_is_text_log` into a try/catch and return False in case of a exception.

shancock884:
If a binary file is passed to the DFReader_is_text_log function, there is a good chance that it will throw an exception, as it attempts to interpret the binary data as a UTF8 string. It should simply return false to say "No, its not a DF Text log". This then allows the TLog reader to handle the file instead.

I was bitten by this when I did a file-naming mistake of my own, but then spotted that issue #710 mentioned the same thing, so thought I would put the proposed fix forward.

The main part of the fix is to simply catch the UnicodeDecodeError exception in DFReader_is_text_log, and return false, as proposed by @booo in the issue description.

A secondary part is that mavlogdump makes some assumptions about the file type based on the file extension, which may not hold true, if (for example) .log extension is used for a Mavlink/Tlog file.
So, I am proposing to use the class type returned by mavutil.mavlink_connection instead to determine the type of the file being dumped.

I also thought that the summary text you get when typing 'mavlink.py -h' was not very clear or complete, so have updated it.

fix #710


tools/mavlogdump.py
```diff
@@ -99,10 +100,10 @@
 if nottypes is not None:
     nottypes = nottypes.split(',')
 
-ext = os.path.splitext(filename)[1]
-isbin = ext in ['.bin', '.BIN', '.px4log']
-islog = ext in ['.log', '.LOG'] # NOTE: "islog" does not mean a tlog
-istlog = ext in ['.tlog', '.TLOG']
+mlog_type = str(type(mlog))
```
peterbarker, line 103: Shouldn't this use `isinstance`?

shancock884: Yeah, I thought of using isinstance first. It does feel like nicer coding and more precise. The only thing that made me use str(type(x)) instead was that I would need to add an import (DFReader).
Happy to change it if that is the preferred etiquette.

tools/mavlogdump.py
```diff
@@ -99,10 +100,10 @@
...
+isbin = "DFReader_binary" in mlog_type
+islog = "DFReader_text" in mlog_type # NOTE: "islog" does not mean a tlog
+istlog = "mavmmaplog" in mlog_type
```
peterbarker, line 106: This should only ever be a dataflash log - if you look at the class you can see it's a DFLOG parser.

So something *very* weird is going on here :-)

tools/mavlogdump.py
shancock884: Code updated to use "isinstance".

tools/mavlogdump.py
shancock884: mavlogdump handles either DF binary, DF text or Mavlink/Tlog format files, and the class of the variable returned by `mlog = mavutil.mavlink_connection(...)` will reflect this.
I added a print on these flags during testing with all 3 file types (including the TLog named both as .log and .tlog), and the right flag gets set.

Issue #913, opened 2024-02-01
Circular import import x25crc in generator/mavparse.py


ScottHo:
https://github.com/ArduPilot/pymavlink/blob/e39879aadc5cb98be10752e2843b55f92c972e07/generator/mavparse.py#L459

The dot import is causing a circular import error on my end.

Changing the import to `from pymavlink.generator.mavcrc import x25crc` fixed it.

`   File "...\lib\site-packages\pymavlink\generator\mavgen.py", line 243, in mavgen

     if not expand_includes():

   File "...\lib\site-packages\pymavlink\generator\mavgen.py", line 116, in expand_includes

    if not expand_oneiteration():

   File "...\lib\site-packages\pymavlink\generator\mavgen.py", line 110, in expand_oneiteration
     xml.append(mavparse.MAVXML(fname, opts.wire_protocol))

   File "...\lib\site-packages\pymavlink\generator\mavparse.py", line 437, in __init__

     m.crc_extra = message_checksum(m)

   File "...\lib\site-packages\pymavlink\generator\mavparse.py", line 459, in message_checksum

    from .mavcrc import x25crc
ImportError: cannot import name 'x25crc' from partially initialized module 'pymavlink.generator.mavcrc' (most likely due to a circular import) (...\lib\site-packages\pymavlink\generator\mavcrc.py)`

PR #914, opened 2024-02-01
Use absolute import to avoid circular imports.

ScottHo:
The PR is to fix https://github.com/ArduPilot/pymavlink/issues/913.

Fixing a circular import on my end.

### DIFF
--- a/generator/mavparse.py
+++ b/generator/mavparse.py
@@ -456,7 +456,7 @@ class MAVXML(object):
 def message_checksum(msg):
     '''calculate CRC-16/MCRF4XX checksum of the key fields of a message, so we
        can detect incompatible XML changes'''
-    from .mavcrc import x25crc
+    from pymavlink.generator.mavcrc import x25crc
     crc = x25crc()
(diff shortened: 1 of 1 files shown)


generator/mavparse.py
```diff
@@ -456,7 +456,7 @@ def __str__(self):
 def message_checksum(msg):
     '''calculate CRC-16/MCRF4XX checksum of the key fields of a message, so we
        can detect incompatible XML changes'''
-    from .mavcrc import x25crc
+    from pymavlink.generator.mavcrc import x25crc
```
tridge, line 459: how is this circular? mavcrc.py doesn't include this module

tridge (changes requested):

generator/mavparse.py
ScottHo: I believe the cause of the issue is I had a partially initialized module due to another python module of the same name.

This fixes it though and is more robust anyways in my opinion 

`AttributeError: partially initialized module 'mavcrc' has no attribute 'x25crc' (most likely due to a circular import)`

Issue #895, opened 2023-12-21
Wireshark Wlua dissection 64-bit flag fields will error on higher bits


shancock884:
The issue is a spin-off from: #852.

As and when Mavlink uses the high 32-bits of any 64-bit bitfield, the generated Wireshark LUA script will error when it is loaded in Wireshark/TShark. This is due to 2 things:

_1/ Creating the protofield bitfield with high bits set:_
For bit 32, current code will generate this, which produces an error when starting Wireshark as the value is too big for a 32-bit integer, so it thinks we are setting a bitmask of zero:
```
ProtoField.bool("mavlink_proto.AUTOPILOT_VERSION_capabilities.HIGH_BIT", "HIGH_BIT", 64, nil, 4294967296)
```
For values higher than maximum uint32, this should be replaced with this:
```
ProtoField.bool("mavlink_proto.AUTOPILOT_VERSION_capabilities.HIGH_BIT", "HIGH_BIT", 64, nil, UInt64(0,1))
```
(it should be noted that this syntax only works from Wireshark v4.2 onwards, released Nov 2023)

_2/ Attaching the bitfield value to the tree:_
Calling add_le with a uint64 value like this, generates an error (only when a relevant message needs to be decoded).
```
value = tvbrange:le_uint64()
tree:add_le(f.AUTOPILOT_VERSION_capabilities_flagHIGH_BIT, tvbrange, value)
```
A fix for now, implemented in #886, replaces value with value:tonumber() for 64-bit bitfields, but this basically chops off any data in the high bits.
This is believed to be an issue with Wireshark itself, which as been raised here: https://gitlab.com/wireshark/wireshark/-/issues/19552

Once fixed, "value:tonumber()" can be replaced back to "value" (or something different, if an alternative way to handle this is proposed by Wireshark developers).

For info: @hamishwillee .

hamishwillee:
Thanks very much for posting. Let's hope Wireshark fix that issue soon.

shancock884:
Update: Wireshark have fixed the issue in their master branch, so I suppose it will arrive in a release at some point this year.

They were also able to propose an alternative way to call the `tree:add_le` function, which works and does not rely on a future version, so I will put forward this as a PR to fix this in due course.
This is to call `tree:add_le` _without_ the value argument. In this case, Wireshark will directly decode the value using the raw bytes in the `tvbrange` argument.
[...]

PR #915, opened 2024-02-02
MavSchema: Allow 64-bit enum values to be specified

shancock884:
Within the Mavlink schema, enum values are currently limited to 10 decimal, 8 hex, or 32 binary digits.
However, we already have two fields which are bitmasks and use uint64_t type (AUTOPILOT_VERSION.capabilities and GENERATOR_STATUS.status). This character limit means that you can't currently specify a value using (most of) the higher 32-bits.
I came across this while trying to build a test case for #895.

My proposal is to extend the character limit to 20 decimal, 16 hex or 64 binary digits to allow for this.
@hamishwillee - does this make sense?

I did wonder if it might be useful to be allowed to specify values as powers of 2 (e.g. "2^0", "2^4", "2^55") to be more readable as the numbers get bigger, but as this would require work on the generators, so have not proposed it.
I also wondered if the test.xml file in Mavlink repo should include at least one enum and bitmask, if it is supposed to be used a nice little test file cover major functionality?

### DIFF
--- a/generator/mavschema.xsd
+++ b/generator/mavschema.xsd
@@ -27,18 +27,18 @@
 <xs:attribute name="value"> <!-- used in entry elements -->
   <xs:simpleType>
     <xs:restriction base="xs:string">
-        <xs:pattern value="\d{1,10}"/> <!-- base 10 int -->
-        <xs:pattern value="0[xX][0-9a-fA-F]{1,8}"/> <!-- base 16 -->
-        <xs:pattern value="0[bB][0-1]{1,32}"/> <!-- base 1 -->
+        <xs:pattern value="\d{1,20}"/> <!-- base 10 int -->
+        <xs:pattern value="0[xX][0-9a-fA-F]{1,16}"/> <!-- base 16 -->
+        <xs:pattern value="0[bB][0-1]{1,64}"/> <!-- base 1 -->
     </xs:restriction>
   </xs:simpleType>
 </xs:attribute>
 <xs:attribute name="default"> <!-- used in enum field elements -->
   <xs:simpleType>
     <xs:restriction base="xs:string">
-        <xs:pattern value="\d{1,10}"/> <!-- base 10 int -->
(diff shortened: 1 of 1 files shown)


hamishwillee:
@shancock884 Thanks very much for this. I think it makes sense but I don't have merge rights in this repo.

@peterbarker @amilcarlucas Can you please review this.

> I did wonder if it might be useful to be allowed to specify values as powers of 2 (e.g. "2^0", "2^4", "2^55") to be more readable as the numbers get bigger, but as this would require work on the generators, so have not proposed it.
  
  It would be useful for bitmask readability. I'd like it, but IMO this would be hard to get in. I suspect you could do this easily enough for mavgen by converting the values in the parser phase so that all the generators get a converted value. But there are more than just the generators in this repo to think about. @peterbarker might have other thoughts.

> I also wondered if the test.xml file in Mavlink repo should include at least one enum and bitmask, if it is supposed to be used a nice little test file cover major functionality?

Absolutely, or a new test file and test. Again, discussion with Peter about what test makes sense would be good.

peterbarker:
I like all of those ideas.

shancock884:
Regarding the "2^n" idea, mavparse.py can _almost_ cope with it already, as it does an "eval" on the value string.
The issue though is that "^" means bitwise-not, and not raise-to-power in Python expressions.
So either, mavparse would need to string replace "^" => "\*\*" first, or the schema get defined such that "\*\*" is used (i.e. directly use Python expression syntax).

amilcarlucas:
I vote for the simpler ** Syntax

hamishwillee:
FWIW More than happy for you to try it but:
- consider restricting to bitmasks
- allow either format
- we will need to have a period where other generators have a chance to catch up before we modify the XML.

shancock884:
> FWIW More than happy for you to try it but:

As there seemed to be some enthusiasm for the power-of-2 idea, I've created PR #920....

PR #917, opened 2024-02-04
Tools: fix mlog attribute in mavplayback

Szymon-SR:
Currently, rewind & forward buttons as well as dragging across the timeline don't work and produce an error:

`Exception in Tkinter callback
  File "/usr/lib64/python3.11/tkinter/__init__.py", line 1948, in __call__
    return self.func(*args)
           ^^^^^^^^^^^^^^^^
  File "/home/szymon/development/opensource/forks/pymavlink/tools/mavplayback.py", line 136, in forward
    self.mlog.filehandle.seek(pos)
    ^^^^^^^^^^^^^^^^^^^^
AttributeError: 'mavmmaplog' object has no attribute 'filehandle'`

Some of the calls were changed from mlog.filehandle to mlog.f by @tridge in #890, it seems to me that the remaining have to be changed as well.

### DIFF
--- a/tools/mavplayback.py
+++ b/tools/mavplayback.py
@@ -122,18 +122,18 @@ class App(object):
 
     def rewind(self):
         '''rewind 10%'''
-        pos = int(self.mlog.f.tell() - 0.1*self.filesize)
+        pos = int(self.mlog.f.tell() - 0.1 * self.filesize)
         if pos < 0:
             pos = 0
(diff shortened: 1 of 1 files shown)


peterbarker (approved):
Usually we'd prefix the commit message (as well as the PR title) with the directory, but we're not really strict in pymavlink.

PR #921, opened 2024-02-12
generator: add javascript "lite"

IamPete1:
This adds a very basic JavaScript generator. This outputs message name and CRC for a given index. Also outputs "MAV_COMPONENT" enum. 

This is not a full parser, and does not consider fields at all. Used for https://github.com/ArduPilot/WebTools/pull/124

Example output https://github.com/IamPete1/WebTools/blob/updates/StreamStats/mavlink_msgs.js


tridge:
why not use the full javascript generator?

IamPete1:
> why not use the full javascript generator?

The full parser is a huge amount of code, it would be much slower to run in my use-case. I do not care about the message payload, just what message it is and the size.

PR #922, opened 2024-02-14
Mavgen C: Use strncpy instead of memcpy to pack char[n] fields

### LINKED ISSUE #725: String serialization
Seems like mavlink is handling string serialization in non common way.

Let's look, for example, at
`uint16_t mavlink_msg_camera_information_pack_chan(... , const char *cam_definition_uri);`

As one would expect cam_definition_uri is variable-length zero-terminated string because it is natural on the 'host' side (C/C++ app). But if you look inside `mavlink_msg_camera_information_pack_chan()` you'll see that it is handled with
`_mav_put_char_array(buf, 95, cam_definition_uri, 140);`
which expands into
`mav_array_memcpy(&buf[wire_offset], b, array_length);`
which expands into
`memcpy(dest, src, n);`
which basically is not either safe nor secure because it can go beyond buffer boundaries and
which was generated by pymavlink and
which is the reason why am I here.

So the workaround to handle that safely is to pre-allocate big enough buffer, copy string there, then pass that buffer to one of those serialization calls. Which sounds odd - if you get your url as a std::string which you use to save time on your dev cycle, because most of the routine is handled by compilier and/or run time library, you have to return back to the roots of development.

Since mavlink serialization code is a bridge between host (C/C++) on one side and mavlink bus on the other side I'd expect that it follows hosts assumptions and translates them correctly into mavlink messages.  What is confusing there on the host side is definition of `const char* cam_definition_uri` - URI and `const char*` assumes zero terminated string. But what current code expects is `char uri[sizeof(mavlink_camera_information_t::cam_definition_uri)]` and therefore `mavlink_msg_camera_information_pack_chan` should look like as
`uint16_t mavlink_msg_camera_information_pack_chan(... , const char uri[sizeof(mavlink_camera_information_t::cam_definition_uri)]);`
which technically is the same pointer as before but at least gives idea to developers what kinds of stuff it expect as input which is already much better. What would make it even better is something like
`uint16_t mavlink_msg_camera_information_pack_chan(... , const char* uri, size_t uri_length);`
And my favorite is commond sense
`uint16_t mavlink_msg_camera_information_pack_chan(... , const char *cam_definition_uri);`
using `srtncpy()` to copy the URI.

So my question is - is there any chance to get the fix merged soon or it is something that can be only evaluated for one of the next major updates? (on my side it looks like - I'll either review all my code and add workarounds where necessary or fix mavlink code generator to work in a way expected by humans).

shancock884:
The aim of this PR is to fix #725 and fix mavlink/mavlink#1946

### Current situation:
At present when a user calls the `mavlink_msg_<message>_pack` function on a message containing char[n] fields, these fields will be copied into the destination buffer using memcpy. This can go wrong in a couple of ways, if end users are not careful, for example....

1. If specifying an argument as a string literal, the read will go beyond the input variable's scope, which could have some bad consequences:
```C
mavlink_msg_param_request_read_pack(mysys, mycomp, &msg, tgtsys, tgtcomp, "PARAM", -1);
```
2. Even if using a large enough buffer, it is possible that the bits of memory beyond the null may contain leftover or random values, which would end up in the message. In this example, bytes after null could be anything when requesting PARAM, and will include the last bit of LONGER_PARAM when reading SHORT:
```C
const char* list_of_params = {"PARAM", "LONGER_PARAM", "SHORT"};
char param[17]; // enough for null
for (int i=0; i<3;i++) {
    strcpy(param,list_of_params_i_want[i]);
    mavlink_msg_param_request_read_pack(mysys, mycomp, &msg, tgtsys, tgtcomp, param, -1);
}
```
These extra bytes will stop the Mavlink 2 message-truncation from being as efficient as it can be, and may confuse some receivers.

### Proposal:
This PR proposes to use the strncpy function when packing char arrays, as I believe that this aligns with the way in which Mavlink char array fields are specified and used:
* destination array will be filled up to a maximum of n chars, and therefore may not be null-terminated.
* if null is found within the n characters, reading stops, and the remainder of the destination up to n is filled with zeros.

This should both protect against reading bad memory, and also ensure bytes beyond the length of short strings are zero, allowing the Mavlink 2 message truncation to work properly.

For little-endian and struct-aligned cases:
This is done by replacing the calls to `mav_array_memcpy` in mavgen_c.py with type-specific array assignment macros `mavlink_array_assign_[type]`.
For char arrays, this uses `strncpy` (after check that input is not null pointer).
For other arrays, this is a thin-macro which calls `mav_array_memcpy`, as per current behaviour.

Big big-endian or non-struct-aligned cases:
This is done by an update to the existing `_mav_put_char_array` function to use `strncpy` (after check that input is not null pointer).

### Testing
I've tested this update (in particular using the PARAM_REQUEST_READ and PLAY_TUNE messages) on my system.
I also checked compilation with MAVLINK_ALIGNED_FIELDS=0.

### DIFF
--- a/generator/C/include_v1.0/protocol.h
+++ b/generator/C/include_v1.0/protocol.h
@@ -187,12 +187,41 @@ static inline void mav_array_memcpy(void *dest, const void *src, size_t n)
 }
 
 /*
+ * Array direct assignment, for use when fields align and no byte swapping
+ * is required. Most are directly #defined to mav_array_memcpy, except for 
+ * mav_array_assign_char, which uses strncpy instead.
+ */
+#if !MAVLINK_NEED_BYTE_SWAP && MAVLINK_ALIGNED_FIELDS
+static inline void mav_array_assign_char(char *dest, const char *src, size_t n)
+{
+	if (src == NULL) {
+		memset(dest, 0, n);
+	} else {
+		strncpy(dest, src, n);
+	}
+}
+#define mav_array_assign_uint8_t(DEST,SRC,N) mav_array_memcpy(DEST,SRC,N*sizeof(uint8_t))
+#define mav_array_assign_int8_t(DEST,SRC,N) mav_array_memcpy(DEST,SRC,N*sizeof(int8_t))
+#define mav_array_assign_uint16_t(DEST,SRC,N) mav_array_memcpy(DEST,SRC,N*sizeof(uint16_t))
+#define mav_array_assign_int16_t(DEST,SRC,N) mav_array_memcpy(DEST,SRC,N*sizeof(int16_t))
+#define mav_array_assign_uint32_t(DEST,SRC,N) mav_array_memcpy(DEST,SRC,N*sizeof(uint32_t))
+#define mav_array_assign_int32_t(DEST,SRC,N) mav_array_memcpy(DEST,SRC,N*sizeof(int32_t))
+#define mav_array_assign_uint64_t(DEST,SRC,N) mav_array_memcpy(DEST,SRC,N*sizeof(uint64_t))
+#define mav_array_assign_int64_t(DEST,SRC,N) mav_array_memcpy(DEST,SRC,N*sizeof(int64_t))
+#define mav_array_assign_float(DEST,SRC,N) mav_array_memcpy(DEST,SRC,N*sizeof(float))
+#define mav_array_assign_double(DEST,SRC,N) mav_array_memcpy(DEST,SRC,N*sizeof(double))
+#endif
+
+/*
  * Place a char array into a buffer
  */
 static inline void _mav_put_char_array(char *buf, uint8_t wire_offset, const char *b, uint8_t array_length)
 {
-	mav_array_memcpy(&buf[wire_offset], b, array_length);
-
+	if (b == NULL) {
+		memset(&buf[wire_offset], 0, array_length);
+	} else {
+		strncpy(&buf[wire_offset], b, array_length);
+	}
 }
 
 /*
--- a/generator/C/include_v0.9/protocol.h
+++ b/generator/C/include_v0.9/protocol.h
@@ -169,12 +169,41 @@ static void mav_array_memcpy(void *dest, const void *src, size_t n)
 }
 
 /*
+ * Array direct assignment, for use when fields align and no byte swapping
+ * is required. Most are directly #defined to mav_array_memcpy, except for 
+ * mav_array_assign_char, which uses strncpy instead.
+ */
+#if !MAVLINK_NEED_BYTE_SWAP && MAVLINK_ALIGNED_FIELDS
+static inline void mav_array_assign_char(char *dest, const char *src, size_t n)
+{
+	if (src == NULL) {
+		memset(dest, 0, n);
+	} else {
+		strncpy(dest, src, n);
+	}
+}
+#define mav_array_assign_uint8_t(DEST,SRC,N) mav_array_memcpy(DEST,SRC,N*sizeof(uint8_t))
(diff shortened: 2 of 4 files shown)


hamishwillee:
@peterbarker Can you please look at this. Fixes https://github.com/ArduPilot/pymavlink/issues/725 which has come up a few times as a problem, in particular for new users after they have discovered this the hard way.

generator/C/include_v1.0/protocol.h
```diff
@@ -186,13 +186,42 @@ static inline void mav_array_memcpy(void *dest, const void *src, size_t n)
...
+ * is required. Most are directly #defined to mav_array_memcpy, except for 
+ * mav_array_assign_char, which uses strncpy instead.
+ */
+#if !MAVLINK_NEED_BYTE_SWAP && MAVLINK_ALIGNED_FIELDS
+static inline void mav_array_assign_char(char *dest, const char *src, size_t n)
+{
+	if (src == NULL) {
+		memset(dest, 0, n);
+	} else {
+		strncpy(dest, src, n);
```
tridge, line 200: the problem is that this will leave the bytes after the string uninitialised, which leaks information and will change the CRC. I think we need to have a mav_strcpy() which zero fills

tridge (changes requested):

generator/C/include_v1.0/protocol.h
shancock884: Hi. My understanding of strncpy is that it already will zero-fill the destination buffer up to n, in the case where the src buffer is shorter.
e.g. [this page](https://pubs.opengroup.org/onlinepubs/9699919799/functions/strncpy.html#:~:text=If%20the%20array%20pointed%20to,are%20written.) states "If the array pointed to by s2 is a string that is shorter than n bytes, NUL characters shall be appended to the copy in the array pointed to by s1, until n bytes in all are written."

julianoes: I didn't know that `strncpy` fills with zeros. I also always assumed it just wrote only one zero. In that case, I think this should work.

@tridge do you agree?

julianoes: I'm bumping this as it would be nice to have!

tridge: hmm, it does seem to be guaranteed by the strncpy definition
```
       If the length of src is less than n, strncpy() writes additional null bytes to dest to ensure that a total of n bytes are written.
```

shancock884: I've added a code comment to each strncpy call to say this, as it is indeed not a very known feature:
`/* strncpy will zero pad dest array up to n bytes */`
Is this then ok to resolve the review comment?
Also rebased to latest master.

peterbarker:
On Tue, 17 Sep 2024, Simon Hancock wrote:

> /* strncpy will zero pad dest array up to n bytes */

Not a blocker at all, but I will note that ArduPilot's "strncpy_noterm" 
does not obey this rule.

hamishwillee:
So there is no reason this can't go in @peterbarker? Can you merge?

peterbarker:
I've marked for discussion at DevCallEU

julianoes:
Someone ran into this again and asked on Discord. So, it would still be worthwhile to fix it.

DonLakeFlyer:
Yeah, I hit this when doing address sanitizer testing in QGC. Would be good to get this in so memory testing could be cleaned up.

hamishwillee:
@peterbarker Can we get this in please. I presume it has already been discussed :-) https://github.com/ArduPilot/pymavlink/pull/922#issuecomment-2458507840

tridge (approved):

PR #923, opened 2024-02-18
mavparm: fixed mavset with float

tridge:
this fixes param revert in MAVProxy


peterbarker:
Merged.  Thought I'd fixed that one already :-/

PR #924, opened 2024-02-28
Replace print statements with python standard library logging

ThomasDavid0:
This is so the std output can be controlled by code that depends on pymavlink.


MauroPfister:
I'd love to see this changed at some point. Thanks for taking the initiative @ThomasDavid0!

One small remark: AFAIK the recommended way to write log messages is `logger.info("timeout setting %s to %f", name, numeric_value)` instead of `logger.info("timeout setting %s to %f" % (name, numeric_value))`.

amilcarlucas (approved):

PR #925, opened 2024-03-03
PyMavlink : change shebangs to python3

kwikius:
See https://discuss.ardupilot.org/t/waf-update-waf-submodule-to-ardupilot-waf-master-user-setup-errror/113898

### DIFF
--- a/CSVReader.py
+++ b/CSVReader.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 '''
 CSV log file reader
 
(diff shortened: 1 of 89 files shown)


davidbuzz:
why? anywhere that ardupilot supports running hthis, its already the case that python is python3 .?

HTRamsey:
@davidbuzz at least then you don't need to create a symbolic link

davidbuzz:
ubuntu/debian has: 
sudo update-alternatives --config python

HTRamsey:
That's still creating a symbolic link. But regardless I was under the assumption that Python2 had been deprecated, which would mean there would be zero reason not to do this. Looking closer I see that Python2 is still supported and you're right this shouldn't be changed until after that is dropped

kwikius:
Close stale PR

PR #926, opened 2024-03-03
Mavgen JS: Heartbeat message mavlink_version should be automaticlly completed

shancock884:
This PR updates the Javascript generator to automate the filling of the HEARTBEAT.mavlink_version parameter.
As per the C Mavgen, it uses 'omit_arg' parameter of the field to determine that the parameter should be omitted from the input parameter list, and 'const_value' parameter to determine the value to set.

When building a HEARTBEAT message via the constructor the parameter should no longer be specified, and the test cases have been updated to do this.
When the message is being read from a data stream, the value is correctly populated to whatever its actual value is, as this uses direct field assignment and not the constructor to build the message (I have tested this).

The change in output ends up looking like this:
from:
```javascript
/*              ...
                system_status             : System status flag. (uint8_t)
                mavlink_version           : MAVLink version, not writable by user, gets added by protocol because of magic data type: uint8_t_mavlink_version (uint8_t)
*/
mavlink20.messages.heartbeat = function( ...moreargs ) {
[ this.type , this.autopilot , this.base_mode , this.custom_mode , this.system_status , this.mavlink_version ] = moreargs;
```
to:
```javascript
/*              ...
                system_status             : System status flag. (uint8_t)
*/
mavlink20.messages.heartbeat = function( ...moreargs ) {
[ this.type , this.autopilot , this.base_mode , this.custom_mode , this.system_status ] = moreargs;
this.mavlink_version = 3;
```

Testing:
I have tested this on my React Native application, including checking that a received message correctly populates the field - even if its value differs from the default.
This is also tested in CI on a variety of NodeJS versions.

### DIFF
--- a/generator/javascript/test/message10.js
+++ b/generator/javascript/test/message10.js
@@ -93,13 +93,12 @@ describe('Complete MAVLink 1.0 packet', function() {
             , base_mode=45
             , custom_mode=68
             , system_status=13
-            , mavlink_version=1
         );
         
         this.mav.seq = 7;
 
         // Create a buffer that matches what the Python version of MAVLink creates
-        var reference = new Buffer.from([0xfe, 0x09, 0x07, 0x2a, 0x96, 0x00, 0x44, 0x00, 0x00, 0x00, 0x05, 0x03, 0x2d, 0x0d, 0x01, 0xac, 0x9d]);
+        var reference = new Buffer.from([0xfe, 0x09, 0x07, 0x2a, 0x96, 0x00, 0x44, 0x00, 0x00, 0x00, 0x05, 0x03, 0x2d, 0x0d, 0x03, 0x1c, 0xae]);
         new Buffer.from(heartbeat.pack(this.mav)).should.eql(reference);
 
     });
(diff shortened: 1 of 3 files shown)


peterbarker:
Looks fine to me, but @davidbuzz will know better than I.

davidbuzz (approved):
looks good to me, thankyou.!

davidbuzz:
@shancock884  -if your're using the Javascript generator in a project, I'd live to hear about it..  eg, I've used it in a few projects... for example, I implemeneted this tool... which is the only one i know of that supports packet-signing with these bindings...  https://github.com/davidbuzz/MAVAgent

peterbarker:
Merged, thanks!

shancock884:
> @shancock884 -if your're using the Javascript generator in a project, I'd love to hear about it..

Hi @davidbuzz. The short version is that I have been working on a mobile phone app for planning waypoint routes - initially just for the drone that our startup (Linking Drones) is working on, using bluetooth. Then to expand its scope of usefulness I was looking into using the phone's USB port as a wired connection to any Mavlink drone. As the app is using React Native I needed some JS!
If you have any follow-ups, feel free to message or email!

PR #927, opened 2024-03-09
WIP: Add debian package files

silentjet:


debian/control
```diff
@@ -0,0 +1,25 @@
...
+Priority: optional
+Maintainer: Andrii Beregovenko <jet@jet.kiev.ua>
+Build-Depends: debhelper (>= 10),
+  dh-python,
+  python3-all-dev (>= 3.3.3),
+  libxml2-dev, libxslt1-dev, zlib1g-dev,
+  python3-setuptools (>= 0.6.29),
+  python3-numpy,
+  python3-pytest,
+X-Python-Version: all
```
peterbarker, line 12: I'm not familiar with this tag set, but this one doesn't look right....

debian/control
```diff
@@ -0,0 +1,25 @@
...
+Section: python
+Priority: optional
+Maintainer: Andrii Beregovenko <jet@jet.kiev.ua>
+Build-Depends: debhelper (>= 10),
+  dh-python,
+  python3-all-dev (>= 3.3.3),
+  libxml2-dev, libxslt1-dev, zlib1g-dev,
+  python3-setuptools (>= 0.6.29),
+  python3-numpy,
+  python3-pytest,
```
peterbarker, line 11: I'm not sure we have a hard dep for this.

debian/control
```diff
@@ -0,0 +1,25 @@
...
+X-Python-Version: all
+Standards-Version: 4.5.1
+Homepage: https://github.com/ArduPilot/pymavlink
+
+Package: python3-pymavlink
+Architecture: any
+Multi-Arch: same
+Depends: ${python3:Depends}, ${shlibs:Depends}, ${misc:Depends}, python3-future, python3-lxml
+Provides: ${python3:Provides}
+Description: Pymavlink is a Python implementation of the MAVLink protocol
```
peterbarker, line 21: It really is three *distinct* things.

The most commonly used mavlink bindings generator.  Which also generates the Python mavlink bindings.

A set of utilities that make working with the Python bindings easier, most notably an object to handle opening data streams.

The command-line utilities that you've mentioned.

I don't think you really need to fix this in your stuff here - it's no different to many other places....

debian/changelog
```diff
@@ -0,0 +1,5 @@
+pymavlink (2.4.41) UNRELEASED; urgency=medium
```
peterbarker, line 1: Is this the changelog for the package or the packaged software?

silentjet:
Hi @peterbarker 
Thanks for your review. Was busy for quite some time. I'll update it and will upload a clean version. This one is quite made-on-knees, thus dirty (and that's why WIP is here).

debian/control
silentjet: Oh well, I just have to specify something here, thus stating what I can at my best :D
I would appreciate a more accurate Summary and Longer description.

debian/changelog
silentjet: hi,
this is a changelog for the package itself... In most cases should be aligned with the actual software version

silentjet:
Hi @peterbarker 
Sorry for such a delay, I worked through the actual build and runtime dependencies, and now they are real and validated.
please review.
building the package can be performed using a command:
```
dpkg-buildpackage -b -us -uc
```
resulting debian package can be installed using `dpkg` tool:
```
dpkg -i ../python3-pymavlink_2.4.41_amd64.deb
```

debian/control
silentjet: ok, fixed

tpwrules:
Could you help me understand why this would be maintained here instead of by Debian?

debian/control
silentjet: Since it is for the localhost builds only (at least as of now), I do not think it is that much important. However if in a future there would be an intention to add the package to official deb/ubu repo, that text need to be refined by the package maintainer (whoever would be adding it)

PR #928, opened 2024-03-10
mavlogdump: show clear error message when mat_file arg is missing

shancock884:
I was trying to use mavlogdump to generate a .mat file from a .bin file, and forgot to specify the --mat_file argument. I found the exception thrown by the call to scipy.io.savemat was not very clear as to my mistake.
This PR adds a check before processing the input file that the required argument is specified, and gives the user a clearer message as follows:
`mat_file argument must be specified when mat format is selected`

### DIFF
--- a/tools/mavlogdump.py
+++ b/tools/mavlogdump.py
@@ -76,6 +76,10 @@ if args.profile:
     yappi.start()
 
 if args.format == 'mat':
+    # Check that the mat_file argument has been specified
+    if args.mat_file is None:
+        print("mat_file argument must be specified when mat format is selected")
+        quit()
     # Load these modules here, as they're only needed for MAT file creation
     import scipy.io
     import numpy as np


tools/mavlogdump.py
```diff
@@ -76,6 +76,10 @@
     yappi.start()
 
 if args.format == 'mat':
+    # Check that the mat_file argument has been specified
+    if not args.mat_file:
+        print("mat_file argument must be specified when mat format is selected")
+        quit()
```
peterbarker, line 82: I'm not familiar with this, but a quick bit of research shows use of this shouldn't be used in scripts, only in REPL.

tools/mavlogdump.py
```diff
@@ -76,6 +76,10 @@
     yappi.start()
 
 if args.format == 'mat':
+    # Check that the mat_file argument has been specified
+    if not args.mat_file:
```
peterbarker, line 80: Should this be
```suggestion
    if not args.mat_file is not None:
```

... i.e. we don't care about the truthiness of the value, just that it was supplied.

tools/mavlogdump.py
shancock884: Changed to use "if args.mat_file is None"

tools/mavlogdump.py
shancock884: Changed to "if args.mat_file is None" (rather than double negative!)

tools/mavlogdump.py
peterbarker: That's not the right resolution for my query :-)

peterbarker: > Ah! In that case I did not understand the query at all....! Are you talking about the use of 'quit()' command, instead of something like exit() or exit(1)? If so, I used it because similar error checking in the same script was already using it. If that's not what you were talking about, then I'm lost!

Yes, quit() shouldn't be used in scripts, FWIU.

But hey, this is no worse than the script already is, so merging.

shancock884: OK thanks! Cleaning all the quits up would be a nice tidy-up, so I can do that too when I get another mo.

PR #929, opened 2024-03-10
DFReader: Create DFMetaData class to handle logger metadata

shancock884:
This PR adds a DFMetaData class to the DFReader.m file, which is used to handle LogMessage XML meta-data for use by MAVExplorer and potentially other tools. It links to PR https://github.com/ArduPilot/MAVProxy/pull/1337.

The class is set as the metadata attribute of DFReader class, and at present includes methods:
* reset - clear cached data, in case of new download, or change of vehicle type
* dot_mavproxy - copy of function in mp_util, to avoid dependency on MavProxy
* metadata_tree - return metadata map for use by other function, including cached version once loaded.
* print_help - print the help for specified log message
* get_description - return the description of the specified log message

Tested using the "logmessage help" command in MAVExplorer (with PR https://github.com/ArduPilot/MAVProxy/pull/1337). For example:
```
MAV> logmessage help ORGN
Log Message: ORGN
Vehicle navigation origin or other notable position

TimeUS [μs] : Time since system startup
Type [instance] : Position type
Lat [deglatitude] : Position latitude
Lng [deglongitude] : Position longitude
Alt [m] : Position altitude
```

### DIFF
--- a/DFReader.py
+++ b/DFReader.py
@@ -20,7 +20,9 @@ import platform
 import time
 
 import struct
-import sys
+import gzip
+import io
+
 from . import mavutil
 
 try:
@@ -555,6 +557,154 @@ class DFReaderClock_gps_interpolated(DFReaderClock):
         m._timestamp = self.timebase + count/rate
 
 
+class DFMetaData(object):
+    '''handle dataflash messages metadata'''
+    def __init__(self, parent):
+        self.parent = parent
+        self.data = None
+        self.metadata_load_attempted = False
+
+    def reset(self):
+        '''clear cached data'''
+        self.data = None
+        self.metadata_load_attempted = False
+
+    @staticmethod
+    def dot_pymavlink(*args):
+        '''return a path to store pymavlink data'''
+        if 'HOME' not in os.environ:
+            dir = os.path.join(os.environ['LOCALAPPDATA'], '.pymavlink')
+        else:
+            dir = os.path.join(os.environ['HOME'], '.pymavlink')
+        if len(args) == 0:
+            return dir
+        return os.path.join(dir, *args)
+
+    @staticmethod
+    def download_url(url):
+        '''download a URL and return the content'''
+        if sys.version_info.major < 3:
+            from urllib2 import urlopen as url_open
+            from urllib2 import URLError as url_error
+        else:
+            from urllib.request import urlopen as url_open
+            from urllib.error import URLError as url_error
+        try:
+            resp = url_open(url)
+        except url_error as e:
+            print('Error downloading %s : %s' % (url, e))
+            return None
+        return resp.read()
+
+    @staticmethod
+    def download():
+        # Make sure the folder to store XML in has been created
+        os.makedirs(DFMetaData.dot_pymavlink('LogMessages'), exist_ok=True)
+        # Loop through vehicles to download
+        for vehicle in ['Rover', 'Copter', 'Plane', 'Tracker', 'Blimp', 'Sub']:
+            url = 'http://autotest.ardupilot.org/LogMessages/%s/LogMessages.xml.gz' % vehicle
+            file = DFMetaData.dot_pymavlink('LogMessages', "%s.xml" % vehicle)
+            print("Downloading %s as %s" % (url, file))
+            data = DFMetaData.download_url(url)
+            if data is None:
+                continue
+            # decompress it...
+            with gzip.GzipFile(fileobj=io.BytesIO(data)) as gz:
+                data = gz.read()
+            try:
+                open(file, mode='wb').write(data)
+            except Exception as e:
+                print("Failed to save to %s : %s" % (file, e))
+
+    def metadata_tree(self, verbose=False):
+        ''' return a map between a log message and its metadata. May return
+        None if data is not available '''
+        # If we've already tried loading data, use it if we have it
+        # This avoid repeated attempts, when the file is not there
+        if self.metadata_load_attempted:
+            return self.data
+        self.metadata_load_attempted = True
+        # Get file name, based on vehicle type
+        mapping = {mavutil.mavlink.MAV_TYPE_GROUND_ROVER : "Rover",
+                   mavutil.mavlink.MAV_TYPE_FIXED_WING : "Plane",
+                   mavutil.mavlink.MAV_TYPE_QUADROTOR : "Copter",
+                   mavutil.mavlink.MAV_TYPE_HELICOPTER : "Copter",
+                   mavutil.mavlink.MAV_TYPE_ANTENNA_TRACKER : "Tracker",
+                   mavutil.mavlink.MAV_TYPE_SUBMARINE : "Sub",
+                   mavutil.mavlink.MAV_TYPE_AIRSHIP : "Blimp",
+                   }
+        if self.parent.mav_type not in mapping:
+            return None
+        path = DFMetaData.dot_pymavlink("LogMessages", "%s.xml" % mapping[self.parent.mav_type])
+        # Does the file exist?
+        if not os.path.exists(path):
+            if verbose:
+                print("Can't find '%s'" % path)
+                print("Please run 'logmessage download' from MAVExplorer, or call")
+                print("DFMetaData.download() from Python.")
+            return None
+        # Read in the XML
+        xml = open(path, 'rb').read()
+        from lxml import objectify
+        objectify.enable_recursive_str()
+        tree = objectify.fromstring(xml)
+        data = {}
+        for p in tree.logformat:
+            n = p.get('name')
+            data[n] = p
+        # Cache and return data
+        self.data = data
+        return self.data
+
+    def print_help(self, msg):
+        '''print help for a log message'''
+        data = self.metadata_tree(verbose=True)
+        if data is None:
+            return
+        if msg not in data:
+            print("No help found for message: %s" % msg)
+            return
+        node = data[msg]
+        # Message name and description
+        print("Log Message: %s\n%s\n" % (msg, node.description.text))
+        # Protect against replay messages which dont list their fields
+        if not hasattr(node.fields, 'field'):
+            return
+        namelist = []
+        unitlist = []
+        # Loop through fields to build list of name/units
+        for f in node.fields.field:
+            namelist.append(f.get('name'))
+            units = f.get('units')
+            dtype = f.get('type')
+            if units:
+                unitlist.append("[%s] " % units)
+            elif 'char' in dtype:
+                unitlist.append("[%s] " % dtype)
+            elif hasattr(f, 'enum'):
+                unitlist.append("[enum] ")
+            elif hasattr(f, 'bitmask'):
+                unitlist.append("[bitmask] ")
+            else:
+                unitlist.append("")
+        # Now get the max string length from each list
+        namelen = len(max(namelist, key=len))
+        unitlen = len(max(unitlist, key=len))
+        # Loop through fields again to do the actual printing
+        for i in range(0, len(namelist)):
+            desc = node.fields.field[i].description.text
+            print("%-*s %-*s: %s" % (namelen, namelist[i], unitlen, unitlist[i], desc))
+
+    def get_description(self, msg):
+        '''get the description of a log message'''
+        data = self.metadata_tree()
+        if data is None:
+            return None
+        if msg in data:
+            return data[msg].description.text
+        return ""
+
+
 class DFReader(object):
     '''parse a generic dataflash file'''
     def __init__(self):
@@ -572,6 +722,7 @@ class DFReader(object):
         self.percent = 0
         self.unit_lookup = {}  # lookup table of units defined by UNIT messages
         self.mult_lookup = {}  # lookup table of multipliers defined by MULT messages
+        self.metadata = DFMetaData(self)
 
     def _rewind(self):
         '''reset state on rewind'''


DFReader.py
```diff
@@ -555,6 +555,103 @@ def set_message_timestamp(self, m):
         m._timestamp = self.timebase + count/rate
 
 
+class DFMetaData(object):
+    '''handle dataflash messages metadata'''
+    def __init__(self,parent):
```
peterbarker, line 560:
```suggestion
    def __init__(self, parent):
```

there are similar formatting things throughout this Python code.  At some stage it would be nice to have this stuff `flake8`-clean, but if we could generally avoid introducing more bad stuff that would be nice.  No need to go back and do so on this PR.

DFReader.py
```diff
@@ -555,6 +555,103 @@ def set_message_timestamp(self, m):
...
+            dir = os.path.join(os.environ['HOME'], '.mavproxy')
+        if name is None:
+            return dir
+        return os.path.join(dir, name)
+
+    def metadata_tree(self,verbose=False):
+        '''return a map between a log message and its metadata. May return None if data is not available'''
+        # If we've already tried loading data, use it if we have it
+        # This avoid repeated attempts, when the file is not there
+        if self.tried:
```
peterbarker, line 585:
```suggestion
        if self.metadata_load_attempted:
```

DFReader.py
```diff
@@ -555,6 +555,103 @@ def set_message_timestamp(self, m):
...
+        mapping = { mavutil.mavlink.MAV_TYPE_GROUND_ROVER : "Rover",
+                    mavutil.mavlink.MAV_TYPE_FIXED_WING : "Plane",
+                    mavutil.mavlink.MAV_TYPE_QUADROTOR : "Copter",
+                    mavutil.mavlink.MAV_TYPE_HELICOPTER : "Copter",
+                    mavutil.mavlink.MAV_TYPE_ANTENNA_TRACKER : "Tracker",
+                    mavutil.mavlink.MAV_TYPE_SUBMARINE : "Sub",
+                    mavutil.mavlink.MAV_TYPE_AIRSHIP : "Blimp",
+                    }
+        if self.parent.mav_type in mapping:
+            path = self.dot_mavproxy(os.path.join("LogMessages","%s.xml" % mapping[self.parent.mav_type]))
```
peterbarker, line 598: I'll bring this up to tridge - perhaps we want a separate directory for pymavlink to put its metadata in, rather than re-using MAVProxy's.  That would entail some calls onto pymavlink to tell it to go and freshen bits of metadata, but would also mean that we're not leaning on the MAVProxy install to get pymavlink functionality.

DFReader.py
```diff
@@ -555,6 +555,103 @@ def set_message_timestamp(self, m):
...
+        # Get file name, based on vehicle type
+        mapping = { mavutil.mavlink.MAV_TYPE_GROUND_ROVER : "Rover",
+                    mavutil.mavlink.MAV_TYPE_FIXED_WING : "Plane",
+                    mavutil.mavlink.MAV_TYPE_QUADROTOR : "Copter",
+                    mavutil.mavlink.MAV_TYPE_HELICOPTER : "Copter",
+                    mavutil.mavlink.MAV_TYPE_ANTENNA_TRACKER : "Tracker",
+                    mavutil.mavlink.MAV_TYPE_SUBMARINE : "Sub",
+                    mavutil.mavlink.MAV_TYPE_AIRSHIP : "Blimp",
+                    }
+        if self.parent.mav_type in mapping:
```
peterbarker, line 597:
```suggestion
        if self.parent.mav_type not in mapping:
            return None
```

pointless `else` here...

DFReader.py
```diff
@@ -555,6 +555,103 @@ def set_message_timestamp(self, m):
...
+            path = self.dot_mavproxy(os.path.join("LogMessages","%s.xml" % mapping[self.parent.mav_type]))
+        else:
+            return None
+        # Does the file exist?
+        if not os.path.exists(path):
+            if verbose:
+                print("Can't find '%s'\nPlease run 'logmessage download' first" % path)
```
peterbarker, line 604: This is pre-supposing `MAVExplorer.py`, of course :-)

DFReader.py
```diff
@@ -555,6 +555,103 @@ def set_message_timestamp(self, m):
...
+            n = p.get('name')
+            data[n] = p
+        # Cache and return data
+        self.data = data
+        return self.data
+
+    def print_help(self,msg):
+        '''print help for a log message'''
+        data = self.metadata_tree(True)
+        if data is None or not msg in data:
```
peterbarker, line 622:
```suggestion
        if data is None or msg not in data:
```

DFReader.py
```diff
@@ -555,6 +555,103 @@ def set_message_timestamp(self, m):
...
+            return
+        node = data[msg]
+        # Message name and description
+        print("Log Message: %s\n%s\n" % (msg,node.description.text))
+        # Protect against replay messages which dont list their fields
+        if hasattr(node.fields,'field'):
```
peterbarker, line 628:
```suggestion
        if not hasattr(node.fields,'field'):
            return
```

early return to simplify

DFReader.py
```diff
@@ -555,6 +555,103 @@ def set_message_timestamp(self, m):
...
+                addstr = ''
+                if units:
+                    addstr = "[%s] " % units
+                elif 'char' in dtype:
+                    addstr = "[%s] " % dtype
+                elif hasattr(f,'enum'):
+                    addstr = "[enum] "
+                elif hasattr(f,'bitmask'):
+                    addstr = "[bitmask] "
+                print("%s %s: %s" % (f.get('name'),addstr,f.description.text))
```
peterbarker, line 642: I imagine we might argue about this format a bit, but I'm not going to here.

... except, perhaps, to mention that fixed-width might be a reasonable option... but `degheading` is way too-long a string, really :-)

DFReader.py
```diff
@@ -555,6 +555,103 @@ def set_message_timestamp(self, m):
...
+
+    def get_description(self,msg):
+        '''get the description of a log message'''
+        data = self.metadata_tree()
+        if data is None:
+            return None
+        if msg in data:
+            return data[msg].description.text
+        else:
+            return ""
```
peterbarker, line 652:
```suggestion
        return ""
```

peterbarker (changes requested):
Have you checked how much extra RAM this consumes when you load an ArduPlane logfile?  Is it simply roughly equal to the size of the metadata file, or does it blow out?

Also, what's the parse-time like on the file?  "not noticable" is probably the answer, but a quick check with a pair of `time.time()` calls would be nice.

DFReader.py
shancock884: ok! I await direction!

shancock884: I can certainly see some advantages of having the download code on the pymavlink side....
- any script wanting to use could do so without it being pre-downloaded (`mylog.metadata.download()`).
- with `mavlogdump --verbose` there could be a `--download` or similar option to make sure that any enums are sure to be nicely populated.

DFReader.py
```diff
@@ -732,6 +833,8 @@ def _add_msg(self, m):
                 self.mav_type = mavutil.mavlink.MAV_TYPE_SUBMARINE
             elif m.Message.find("Blimp") != -1:
                 self.mav_type = mavutil.mavlink.MAV_TYPE_AIRSHIP
+            # Reset the meta data, when mav type is updated
+            self.metadata.reset()
```
tridge, line 837: should only do it when mav_type changes

DFReader.py
```diff
@@ -744,6 +847,9 @@ def _add_msg(self, m):
             mavtype = build_types.get(m.BU,None)
             if mavtype is not None:
                 self.mav_type = mavtype
+                # Reset the meta data, when mav type is updated
+                self.metadata.reset()
```
tridge, line 851: should only reset when it actually changes

tridge (changes requested):

DFReader.py
shancock884: I did some checks, and as speculated, this doesn't appear to be needed at all..... init_arrays, called by constructor will have set the mav_type for the file before anyone can try accessing any metadata. Therefore I propose to remove the lines.

DFReader.py
shancock884: I did some checks, and as speculated, this doesn't appear to be needed at all..... init_arrays, called by constructor will have set the mav_type for the file before anyone can try accessing any metadata. Therefore I propose to remove the lines.

DFReader.py
```diff
@@ -555,6 +555,105 @@ def set_message_timestamp(self, m):
...
+
+    def dot_pymavlink(self, *args):
+        '''return a path to store mavproxy data'''
+        if 'HOME' not in os.environ:
+            dir = os.path.join(os.environ['LOCALAPPDATA'], '.pymavlink')
+        else:
+            dir = os.path.join(os.environ['HOME'], '.pymavlink')
+        if len(args) == 0:
+            return dir
+        return os.path.join(dir, *args)
```
shancock884, line 578: @peterbarker - It feels like this should be in mavutil or something for consistent use by all pymavlink functions.
(for example one place that accesses .mavproxy specifies "~" - which looks very Linux specific!)
Would you agree, and if so, shall I do so on this PR, or on a pre-PR, or a later PR?

DFReader.py
shancock884: Something like this....?
```
TimeUS   [μs]         : Time since system startup
DesRoll  [deg]        : vehicle desired roll
Roll     [deg]        : achieved vehicle roll
DesPitch [deg]        : vehicle desired pitch
Pitch    [deg]        : achieved vehicle pitch
DesYaw   [degheading] : vehicle desired yaw
Yaw      [degheading] : achieved vehicle yaw
ErrRP    [deg]        : lowest estimated gyro drift error
ErrYaw   [degheading] : difference between measured yaw and DCM yaw estimate
AEKF                  : active EKF type
```

peterbarker:
On Wed, 20 Mar 2024, Simon Hancock wrote:

> 
> @shancock884 commented on this pull request.
> 
> _______________________________________________________________________________________________________________________________________________________________________________________________________________
> 
> In DFReader.py:
> 
> > +    def dot_pymavlink(self, *args):
> +        '''return a path to store mavproxy data'''
> +        if 'HOME' not in os.environ:
> +            dir = os.path.join(os.environ['LOCALAPPDATA'], '.pymavlink')
> +        else:
> +            dir = os.path.join(os.environ['HOME'], '.pymavlink')
> +        if len(args) == 0:
> +            return dir
> +        return os.path.join(dir, *args)
> 
> @peterbarker - It feels like this should be in mavutil or something for consistent use by all pymavlink functions.
> (for example one place that accesses .mavproxy specifies "~" - which looks very Linux specific!)
> Would you agree, and if so, shall I do so on this PR, or on a pre-PR, or a later PR?

I'd suggest keeping this self-contained initially and we can look at 
factorying out on a future PR.

DFReader.py
shancock884: Message updated to be useful for either case:
```
Can't find '%s'
Please run 'logmessage download' from MAVExplorer, or call
DFMetaData.download() from Python.
```

shancock884:
> Have you checked how much extra RAM this consumes when you load an ArduPlane logfile? Is it simply roughly equal to the size of the metadata file, or does it blow out?
> 
> Also, what's the parse-time like on the file? "not noticable" is probably the answer, but a quick check with a pair of `time.time()` calls would be nice.

I still need to do the RAM/CPU time check, but I think all other comments are addressed.

DFReader.py
```diff
@@ -555,6 +557,154 @@ def set_message_timestamp(self, m):
...
+
+    @staticmethod
+    def download_url(url):
+        '''download a URL and return the content'''
+        if sys.version_info.major < 3:
```
peterbarker, line 586: Probably don't need this compatability code.  IIRC the "exists_ok" used below makes this stuff Py3 only anyway...

DFReader.py
```diff
@@ -555,6 +557,154 @@ def set_message_timestamp(self, m):
...
+            file = DFMetaData.dot_pymavlink('LogMessages', "%s.xml" % vehicle)
+            print("Downloading %s as %s" % (url, file))
+            data = DFMetaData.download_url(url)
+            if data is None:
+                continue
+            # decompress it...
+            with gzip.GzipFile(fileobj=io.BytesIO(data)) as gz:
+                data = gz.read()
+            try:
+                open(file, mode='wb').write(data)
```
peterbarker, line 615: I've become rather partial to `pathlib` myself, but I do understand this is just swiped from MAVProxy.

DFReader.py
```diff
@@ -555,6 +557,154 @@ def set_message_timestamp(self, m):
...
+            from urllib2 import urlopen as url_open
+            from urllib2 import URLError as url_error
+        else:
+            from urllib.request import urlopen as url_open
+            from urllib.error import URLError as url_error
+        try:
+            resp = url_open(url)
+        except url_error as e:
+            print('Error downloading %s : %s' % (url, e))
+            return None
```
peterbarker, line 596: I wonder if we really should be doing this rather than making the caller catch the exceptions...

DFReader.py
```diff
@@ -555,6 +557,154 @@ def set_message_timestamp(self, m):
...
+        # Loop through vehicles to download
+        for vehicle in ['Rover', 'Copter', 'Plane', 'Tracker', 'Blimp', 'Sub']:
+            url = 'http://autotest.ardupilot.org/LogMessages/%s/LogMessages.xml.gz' % vehicle
+            file = DFMetaData.dot_pymavlink('LogMessages', "%s.xml" % vehicle)
+            print("Downloading %s as %s" % (url, file))
+            data = DFMetaData.download_url(url)
+            if data is None:
+                continue
+            # decompress it...
+            with gzip.GzipFile(fileobj=io.BytesIO(data)) as gz:
```
peterbarker, line 612: We actually also produce .xz for these... they're about 75% the size of the .gz

peterbarker (approved):
LGTM

peterbarker:
Tested it.  Merged,thanks!

PR #930, opened 2024-03-17
mavlogdump: Use sys.exit instead of quit on error

shancock884:
As @peterbarker pointed out in comments on #928, quit() should not be normally be used in scripts, but instead sys.exit() is preferred.
This PR replaces the uses of quit() and exit() in mavlogdump with sys.exit(1) - all uses are error conditions, hence using return code 1.
I tested a few of the error conditions out, and checked they return neatly with the correct error code set, for example:
```
$ mavlogdump.py --format=csv file.BIN 
Need exactly one type when dumping CSV from bin file
$ echo $?
1
```


peterbarker (approved):
LGTM

peterbarker:
Merged, thanks!

PR #931, opened 2024-03-20
mavutil: Raise ValueError instead of exiting

Newbytee:
Users of this library may want to handle these values being wrong. While it is possible to handle sys.exit, it is more conventional to raise a ValueError on incorrect data passed as an argument.

### DIFF
--- a/mavutil.py
+++ b/mavutil.py
@@ -1053,8 +1053,7 @@ class mavudp(mavfile):
     def __init__(self, device, input=True, broadcast=False, source_system=255, source_component=0, use_native=default_native, timeout=0):
         a = device.split(':')
         if len(a) != 2:
-            print("UDP ports must be specified as host:port")
-            sys.exit(1)
+            raise ValueError("UDP ports must be specified as host:port")
         self.port = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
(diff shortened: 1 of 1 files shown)


peterbarker (approved):
Tested MAVProxy's error output and it doesn't make much difference as MAVProxy caches any exception and handles it:

```
pbarker@fx:~/rc/pymavlink((HEAD detached at Newbytee/raise-value-error-instead-of-exit))$ mavproxy.py --master udp:bob
Connect udp:bob source_system=255
Failed to connect to udp:bob : UDP ports must be specified as host:port
pbarker@fx:~/rc/pymavlink((HEAD detached at Newbytee/raise-value-error-instead-of-exit))$ 
```

... it actually fixes a bug in MAVProxy if you screw up:

```
STABILIZE> Adding link udp:bob
Connect udp:bob source_system=255
UDP ports must be specified as host:port

STABILIZE> 
STABILIZE> 
STABILIZE> link add udp:bob
STABILIZE> 
STABILIZE> 
STABILIZE> 
STABILIZE> link add udp:bob
STABILIZE> status
STABILIZE> status
STABILIZE> 
```

peterbarker:
Merged, thanks!

Issue #864, opened 2023-10-10
Switching between versions of dialects


MauroPfister:
I was wondering if there is an easy way to switch between different versions of dialects in code. 

We are using a custom dialect to communicate with various devices over mavlink. Not all these devices necessarily use the same version of the dialect and we would like to be able to switch which dialect we're using depending on what device we are connected to.

Is there some way to achieve this?

I assume that I would have to generate the python modules for each version of the dialect according to [these instructions](https://mavlink.io/en/mavgen_python/#generate-a-custom-mavlink-dialect) and then make them available to mavlink by setting the `MAVLINK_DIALECT` and `MDEF` environment variables as mentioned [here](https://mavlink.io/en/mavgen_python/#dialect_file)?

peterbarker:
No easy way to do that.  You can include the new xml in `all.xml` and use the `all` dialect, however.

MauroPfister:
The problem here is that these dialects contain mostly the same messages but might be incompatible because some message definitions are different. I guess in this case I couldn't just include them in `all.xml` because the same message would be defined multiple times. Besides that, there would be no way of versionning the different dialects.

I recently discovered, that I can simply add the generated dialect `.py` files to the `dialects/v20/` folder of the pymavlink installation and then select that dialect with `mavutil.set_dialect("my_dialect.py")`. However, it doesn't seem possible to load these dialect files from another location (which would solve my problem). Is there any reason for this?

PR #934, opened 2024-03-28
mavutil: Add option to set dialect directly from file

MauroPfister:
Add option to set a dialect directly from a file anywhere on the system. The dialect file has to be generated beforehand from message definitions using mavgen.

This is useful in situations where different versions of the same dialect might be needed to talk to different devices (see https://github.com/ArduPilot/pymavlink/issues/864).

Example:
```python
from pymavlink import mavutil
import os

os.environ['MAVLINK20'] = "1"

mavutil.set_dialect_from_file("/home/user/my_dialects/v20/my_special_dialect.py")
```

I'm not sure if this is something that pymavlink intends to support or if there are better solutions to this problem. Happy to be educated.

### DIFF
--- a/mavutil.py
+++ b/mavutil.py
@@ -144,6 +144,28 @@ def set_dialect(dialect, with_type_annotations=None):
     current_dialect = dialect
     mavlink = mod
 
+def set_dialect_from_file(file):
+    '''set the MAVLink dialect to work with from a file.
+    For example, set_dialect_from_file("/home/user/my_dialects/v20/my_special_dialect.py")
+    '''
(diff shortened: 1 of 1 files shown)


khancyr:
probably using venv for each dialect type could work.

MauroPfister:
That only works if you don't need to switch dialect at runtime. Imagine a test application that can connect (at runtime) to various devices that might not use the same dialect. In this case using different vens won't be an option.

MauroPfister:
Any chance to get a review on this?

PR #937, opened 2024-04-14
DFReader.py: add bitmask data to verbose print output

peterbarker:
.. taking the conservative approach to this new feature failing....

@shancock884

### DIFF
--- a/DFReader.py
+++ b/DFReader.py
@@ -293,12 +293,85 @@ class DFMessage(object):
             ret = ret[:-2]
         return ret + '}'
 
+    def dump_verbose_bitmask(self, f, c, val, field_metadata):
+        try:
+            try:
+                bitmask = field_metadata["bitmask"]
+            except Exception:
+                return
+
+            # work out how many bits to show:
+            t = field_metadata.get("type")
+            bit_count = None
+            if t == "uint8_t":
+                bit_count = 8
+            elif t == "uint16_t":
+                bit_count = 16
+            elif t == "uint32_t":
+                bit_count = 32
+
+            if bit_count is None:
+                return
+
+            highest = -1
+
+            # we show bit values at least up to the highest bit set:
+            for i in range(bit_count):
+                if val & (1<<i):
+                    highest = i
+
+            # we show bit values at least up until the highest
+            # bit we have a description for:
+            for bit in bitmask.bit:
+                bit_offset = int(math.log(bit["value"], 2))
+                if bit_offset > highest:
+                    highest = bit_offset
+
+            for i in range(bit_offset):
+                bit_value = 1 << i
+                done = False
+                for bit in bitmask.bit:
+                    if bit["value"] != bit_value:
+                        continue
+                    if val & bit_value:
+                        bang = ""
+                    else:
+                        bang = "!"
+                    bit_name = bit.get('name')
+                    bit_desc = None
+                    try:
+                        bit_desc = bit["description"]
+                    except KeyError:
+                        pass
+                    if bit_desc is None:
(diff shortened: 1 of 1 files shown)


peterbarker:
```
2024-03-28 15:13:19.699: POWR
    TimeUS: 211395439 µs
    Vcc: qnan V
    VServo: qnan V
    Flags: 0
        !BRICK_VALID (main brick power supply valid)
        !SERVO_VALID (main servo power supply valid for FMU)
        !USB_CONNECTED (USB power is connected)
        !PERIPH_OVERCURRENT (peripheral supply is in over-current state)
        !PERIPH_HIPOWER_OVERCURRENT (hi-power peripheral supply is in over-current state)
    AccFlags: 0
        !BRICK_VALID (main brick power supply valid)
        !SERVO_VALID (main servo power supply valid for FMU)
        !USB_CONNECTED (USB power is connected)
        !PERIPH_OVERCURRENT (peripheral supply is in over-current state)
        !PERIPH_HIPOWER_OVERCURRENT (hi-power peripheral supply is in over-current state)
    Safety: 6

2024-03-28 15:13:19.700: RCI2
    TimeUS: 211396039 µs
    C15: 1501 us
    C16: 1501 us
    OMask: 0
    Flags: 1
        HAS_VALID_INPUT (true if the system is receiving good RC values)
```

shancock884:
Looks nice! A couple of thoughts you are free to consider or ignore :-) ....

I know it is the same as dump_message_verbose in mavutil, but I don't really like the use of "!" for not.... when I first came across it my eyes read it as a check-mark meaning true (i.e. opposite of what it means)! But maybe its best to remain consistent until someone has a strong enough argument/opinion to change both.

I was thinking it could be good to split the dump_verbose_bitmask function into two: get_field_metadata and dump_verbose_bitmask, as it should make it easier to implement showing enums too in a later PR without looking up metadata twice (and probably needed to do following suggestion...).

It would be nice to show the hex value for bitmasks, alongside the decimal (as dump_message_verbose does). If you do create get_field_metadata, then you can get the field metadata first and do something like `if hasattr(field_metadata, 'bitmask'): f.write(" 0x%X (%d)\n" % (val, val))`.

peterbarker:
> I know it is the same as dump_message_verbose in mavutil, but I don't really like the use of "!" for False values.... when I first came across it my eyes read it as a check-mark meaning true (i.e. opposite of what it means)! But maybe its best to remain consistent until someone has a strong enough argument/opinion to change both.

You wouldn't be surprised to hear I find the ! thing readable, but definitely open to alternatives.

> I was thinking it could be good to split the dump_verbose_bitmask function into two: get_field_metadata and dump_verbose_bitmask, as it should make it easier to implement showing enums too in a later PR without looking up metadata twice (and probably needed to do following suggestion...).

> It would be nice to show the hex value for bitmasks, alongside the decimal (as dump_message_verbose does). If you do create get_field_metadata, then you can get the field metadata first and do something like `if hasattr(field_metadata, 'bitmask'): f.write(" 0x%X (%d)\n" % (val, val))`.

Seems reasonable.  If you can put those suggestions in patch form I can apply them... otherwise I'll come back to this in a while.

Or I could merge and accept patches later?

peterbarker:
@shancock884 I had another look at your suggestions.  I've re-arranged things a little to be somewhat like what you're suggesting, but haven't modified the printing of the value.  Too many cases to test for this PR ;-)

Further patches would probably be something like `field_metadata = field_metadata_by_name.get(c, None)`, test for `None` instead of presence-in-field-metadata-by-name in the call to dump_verbose_bitmask, and then some weird stuff when printing the values.  It's complicated by the py2 compat code.

PR #943, opened 2024-04-26
use fastcrc if available

tridge:
fastcrc runs about 3x faster than existing x25 crc

### DIFF
--- a/generator/mavcrc.py
+++ b/generator/mavcrc.py
@@ -7,8 +7,13 @@ Released under GNU LGPL version 3 or later
 import sys
 from builtins import object
 
+try:
+    import fastcrc
+    mcrf4xx = fastcrc.crc16.mcrf4xx
+except Exception:
+    mcrf4xx = None
 
-class x25crc(object):
+class x25crc_slow(object):
     """CRC-16/MCRF4XX - based on checksum.h from mavlink library"""
 
     def __init__(self, buf=None):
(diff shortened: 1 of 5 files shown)


generator/mavcrc.py
```diff
@@ -39,3 +44,57 @@ def accumulate_str(self, buf):
...
+class x25crc_fast(object):
+    """CRC-16/MCRF4XX - based on checksum.h from mavlink library"""
+    def __init__(self, buf=None):
+        self.crc = 0xFFFF
+        if buf is not None:
+            self.accumulate(buf)
+
+    def accumulate(self, buf):
+        """add in some more bytes (it also accepts strings)"""
+        if sys.version_info[0] == 2:
```
peterbarker, line 57: Should we really be adding Py2 compat here?

generator/mavcrc.py
```diff
@@ -39,3 +44,57 @@ def accumulate_str(self, buf):
...
+            if type(buf) is str:
+                buf = bytearray(buf)
+            elif type(buf).__name__ == 'unicode':
+                # we can't use the identifier unicode in python3
+                buf = bytearray(buf.encode())
+        elif type(buf) is str:
+            buf = bytes(buf.encode())
+        elif type(buf) is list:
```
peterbarker, line 65: Should all of these be `isinstance`?

peterbarker (approved):
A lot of ancient code being duplicated, so we're adding sub-optimal stuff in.... but at least it's well-tested sub-optimal stuff.

PR #944, opened 2024-04-30
Mavgen WLUA: Fix incorrect display of negative ints on mac

### LINKED ISSUE #701: Wireshark showing 0 for negative int32
With mission items with negative y (negative int32 value), the wireshark plugin incorrectly shows 0, even though the data is clearly not 0 (as seen in the hex dump)

This is cross posted from https://github.com/mavlink/mavlink/issues/1862

shancock884:
This PR is to fix https://github.com/mavlink/mavlink/issues/1862, fix https://github.com/mavlink/mavlink/issues/2073 and fix #701. (In particular https://github.com/mavlink/mavlink/issues/2073 includes the discussion about how it was figured out!).

There is a bug in wireshark, whereby a double is cast to an unsigned int when it should be a signed int, before being used as a signed int. On Linux this works, but on Mac negative values get cut to zero ([here is why](https://embeddeduse.com/2013/08/25/casting-a-negative-float-to-an-unsigned-int/)).
Thanks to @rotu, a Wireshark fix has been merged in [wireshark/wireshark#15438](https://gitlab.com/wireshark/wireshark/-/merge_requests/15438).

This PR provides a workaround/fix to avoid the issue completely, so that it will work even with older versions of Wireshark.

The fix is to _not_ include the value argument when calling add_le function, and let Wireshark do the value extraction internally from the raw bytes. This avoids the code which does the bad casting. It also makes the generated lua file have about 20-25% fewer lines of code.

There are some case in which the value still needs to separately extracted:
* The value argument is still required in the case when a float is used to represent an enum or bitmask, as on a number of command parameters.
* Value is used when adding unit conversions to the displayed value.
* Value is used for the bitmask extraction functions, as the input may have come from a command float parameter.

I've tested new LUA with some basic scenarios on Linux, and the CI also includes a decent selection of pytests for this.

@rotu

### DIFF
--- a/generator/mavgen_wlua.py
+++ b/generator/mavgen_wlua.py
@@ -308,19 +308,37 @@ def generate_field_dissector(outf, msg, field, offset, enums, cmd=None, param=No
         else:
             field_var = t.substitute("${fmsg}_${fname}${findex}", {'fmsg': msg.name, 'fname': field.name, 'findex': index_text})
 
-        t.write(outf,
+        # If there is an associated enum and the datatype is not uint, we need to extract
+        # and pass the value to add_le, as the raw and ProtoField types will not match.
+        # This occurs in the case of using a command field to represent an enum or bitmask
+        if enum_obj and tvb_func != "le_uint":
+            value_extracted = True
+            t.write(outf,
 """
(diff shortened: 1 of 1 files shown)


rotu:
lgtm!

PR #946, opened 2024-05-29
Mavgen WLua: Handle multiplier attribute in XML parser

shancock884:
This follows the addition of multiplier XML attribute (#885, #894), and its population on some message fields (https://github.com/mavlink/mavlink/pull/2120).

In mavparse.py: Read in the multiplier attribute for message fields and command parameters and store it in the internal classes for access by any generator wishing to use it.

In Mavgen WLua: Handle the multiplier when displaying values in Wireshark. Add a test case to demonstrate and test the behaviour. Also, a little code tidy-up by removing usused assignment of mavlink_type variable.

The change in output goes from something like this:
```
Payload: GPS_RAW_INT (24)
    ...
    eph (uint16_t): 233
    epv (uint16_t): 178
```
to something like this:
```
Payload: GPS_RAW_INT (24)
    ...
    eph (uint16_t*1E-2): 2.33
    epv (uint16_t*1E-2): 1.78
```

FYI: @hamishwillee , @peterbarker , @rotu


peterbarker:
@shancock884 always a good idea to say what testing you've done :-)

shancock884:
> @shancock884 always a good idea to say what testing you've done :-)

Yes you're right!
Using the common.xml changes proposed in https://github.com/mavlink/mavlink/pull/2120, I generated the new Wireshark lua file.
Checked the "1E-2" text ends up where I wanted it to - i.e. as part of the type in the field label, and to multiply by the extracted value for displaying.
Ran Wireshark with my test-drone outputting via UDP, and checked displayed values on GPS_RAW messages, and a few others for regression.
Used cut-down version of this Wireshark log to create new test case in CI.

peterbarker:
Merged, thanks!

PR #960, opened 2024-07-26
Mavutil: fix wrong sysid_state update from sysid mismatch

khancyr:
On a link with multiple mavlink device, we store the state of all device into the sysid holding the port which is tottally wrong.

The easiest way to see this bug in action is to run two SITL instance connect via mavlink. Then on either mavproxy instance, use "alllinks mode guided" the SITL instance directly connect to mavproxy will update its mode fine, the other will change to another mode 
image

GDB screen from mavproxy on the call of mavutil post_message() function

image

After fix : 
image


peterbarker:
Wow.,  Whoops.  Thanks @khancyr !

PR #962, opened 2024-08-03
A simple "Print flight controller banner statustext message contents"…

amilcarlucas:
… example

### DIFF
--- a/examples/status_msg.py
+++ b/examples/status_msg.py
@@ -1,0 +1,394 @@
+#!/usr/bin/env python
+
+'''
+Print flight controller banner statustext messages and AUTOPILOT_VERSION message information.
+
+SPDX-FileCopyrightText: 2024 Amilcar do Carmo Lucas <amilcar.lucas@iav.de>
+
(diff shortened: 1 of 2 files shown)


examples/status_msg.py
```diff
@@ -0,0 +1,68 @@
...
+
+import time
+
+from argparse import ArgumentParser
+
+from pymavlink import mavutil
+
+
+def print_banner_messages(master):
+    '''Decode and print banner information from the flight controller'''
```
peterbarker, line 19: I'm a bit concerned by this "decode" bit here.  Statustexts really should be opaque to software, they're really not a good way of transmitting data to automated systems.

An example decoding the `AUTOPILOT_VERSION` message would be good, 'though :-)  Particularly because of the way we encode the board ID in that message.
```suggestion
    '''print banner information from the flight controller'''
```

peterbarker (changes requested):

peterbarker:
Merged the commits, fixed the commit message and merged it.  Thanks!

PR #965, opened 2024-08-11
DFReader: fixed handling of multlipiers from FMTU and MULT

tridge:
fixes XKF4 scaled innovations with ArduPilot master

### DIFF
--- a/DFReader.py
+++ b/DFReader.py
@@ -166,7 +166,7 @@ class DFFormat(object):
         for i in range(len(self.units)):
             # If the format has its own multiplier, do not adjust the unit,
             # and if no unit is specified there is nothing to adjust
-            if self.msg_mults[i] is not None or self.units[i] == "":
+            if self.msg_mults[i] is not None or self.units[i] != "":
                 continue
             # Get the unit multiplier from the lookup table
             if mult_ids[i] in mult_lookup:
@@ -176,6 +176,9 @@ class DFFormat(object):
                     self.units[i] = MULT_TO_PREFIX[unitmult]+self.units[i]
                 else:
                     self.units[i] = "%.4g %s" % (unitmult, self.units[i])
+                if self.units[i] in FORMAT_TO_STRUCT:
+                    (label, mul, t) = FORMAT_TO_STRUCT[self.units[i]]
+                    self.msg_mults[i] = mul
 
     def get_unit(self, col):
         '''Return the unit for the specified field'''


peterbarker:
Ping @shancock884

DFReader.py
```diff
@@ -166,7 +166,7 @@ def set_mult_ids(self, mult_ids, mult_lookup):
         for i in range(len(self.units)):
             # If the format has its own multiplier, do not adjust the unit,
             # and if no unit is specified there is nothing to adjust
-            if self.msg_mults[i] is not None or self.units[i] == "":
+            if self.msg_mults[i] is not None or self.units[i] != "":
```
shancock884, line 169: I think you need just:
```suggestion
            if self.msg_mults[i] is not None:
```
Otherwise, things like *.TimeUS will say its in 's' instead of 'μs' and PM.Load will say its in '%' instead of 'd%', when running 'dump PM --verbose'.

DFReader.py
```diff
@@ -176,6 +176,9 @@ def set_mult_ids(self, mult_ids, mult_lookup):
                     self.units[i] = MULT_TO_PREFIX[unitmult]+self.units[i]
                 else:
                     self.units[i] = "%.4g %s" % (unitmult, self.units[i])
+                if self.units[i] in FORMAT_TO_STRUCT:
+                    (label, mul, t) = FORMAT_TO_STRUCT[self.units[i]]
+                    self.msg_mults[i] = mul
```
shancock884, line 182: I think "if self.units[i] in FORMAT_TO_STRUCT:" is doing what you want by a bit of chance, as 'c' is both prefix for 'centi' and also code for 'int16_t * 100'. Maybe something more explicit such as this for lines 175-181 (not tested!):
```suggestion
                # If no unit is specified, apply the multiplier to the value,
                # instead of prefixing the unit
                if self.units[i] == "":
                    self.msg_mults[i] = mult_lookup[mult_ids[i]]
                elseif unitmult in MULT_TO_PREFIX:
                    self.units[i] = MULT_TO_PREFIX[unitmult]+self.units[i]
                else:
                    self.units[i] = "%.4g %s" % (unitmult, self.units[i])
```
This would then work in cases which have any multiplier (not just *100) and no unit.

DFReader.py
```diff
@@ -176,6 +176,9 @@ def set_mult_ids(self, mult_ids, mult_lookup):
                     self.units[i] = MULT_TO_PREFIX[unitmult]+self.units[i]
                 else:
                     self.units[i] = "%.4g %s" % (unitmult, self.units[i])
+                if self.units[i] in FORMAT_TO_STRUCT:
```
shancock884, line 179: I think that the "if self.units[i] in FORMAT_TO_STRUCT" bit is working because the prefix for "centi" and the code for "int16_t * 100" are both 'c' - but in my mind they aren't really the same 'c'!

Assuming your aim is that any parameter which has no-unit but does have a multiplier specified, has that multiplier applied to the value, instead of becoming a unit prefix, then maybe something like this is cleaner:
```suggestion
                # 
                if self.units[i] == "":
                    self.msg_mults[i] = mult_lookup[mult_ids[i]]
                elif unitmult in MULT_TO_PREFIX:
                    self.units[i] = MULT_TO_PREFIX[unitmult]+self.units[i]
                else:
                    self.units[i] = "%.4g %s" % (unitmult, self.units[i])
```
(to replace all lines 175 -181)

Not checked in detail, but I think this should then work for any multiplier which is set on a unitless parameter, even if that multiplier is not 100.

amilcarlucas:
Any progress on this?

PR #967, opened 2024-08-15
mavutil: add WebSocket Server mavlink_connection type

IamPete1:
This adds support for a WebSocket server connection type. This is used with the prefix `wsserver`. For example `wsserver:localhost:5763`.

https://github.com/mavlink/mavlink2rest also does WebSockets, however, it sends a MAVLink JSON object in a text WebSocket message. This sends raw MAVLink in the binary WebSocket message. 

I have tested with a browser based client. It does work, but I wrote it too, so its not a independent test. I have also only tested sending from server to client, not from client to server. 

The more common python WebSocket [library ](https://websockets.readthedocs.io/en/stable/) is designed for a asyncio workflow which makes it hard to integrate here. I ended up using [wsproto](https://python-hyper.org/projects/wsproto/en/stable/) mostly because it had a synchronous server [example](https://github.com/python-hyper/wsproto/blob/main/example/synchronous_server.py).

It would be nice to get multi client working in the future. I'm not sure if we would want that to work under this same prefix or if we would want to add a new prefix to enable that. 

Packetisation to get more data into each packet might also be a good idea.

### DIFF
--- a/mavutil.py
+++ b/mavutil.py
@@ -1730,6 +1730,120 @@ class mavchildexec(mavfile):
     def write(self, buf):
         self.child.stdin.write(buf)
 
+class mavwebsocket(mavfile):
+    '''Mavlink WebSocket server, single client only'''
+    from wsproto import ConnectionType, WSConnection, utilities
+    from wsproto.events import (
+        AcceptConnection,
+        CloseConnection,
+        Request,
+        BytesMessage,
+    )
+
+    def __init__(self, device, source_system=255, source_component=0, use_native=default_native):
+        self.ws = None
+
+        # This is a duplicate of mavtcpin
+        a = device.split(':')
+        if len(a) != 2:
+            raise ValueError("WebSocket ports must be specified as host:port")
+        self.listen = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        self.listen_addr = (a[0], int(a[1]))
+        self.listen.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
+        self.listen.bind(self.listen_addr)
+        self.listen.listen(1)
+        self.listen.setblocking(0)
+        set_close_on_exec(self.listen.fileno())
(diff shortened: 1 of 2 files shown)


tridge (approved):

tridge:
nice!

iottrends:
Need some guidance on how to process mavlink data using websocket in backend python.

working on a web client version, where in java script  sends serial data over websocket: socket.emit('mavlinkData', {mavlik_msg });

server side. 

def connect_ws(self): 
        """Connect to MAVLink via WebSocket."""
        if not self.ws_uri:   #self.ws_uri = "wsserver:0.0.0.0:5760"
            mavlink_logger.error("❌ No WebSocket URI provided")
            return False

        try:
            mavlink_logger.info(f"🔌 Connecting to WebSocket: {self.ws_uri}")
            self.mav_conn = mavutil.mavlink_connection(self.ws_uri, dialect="ardupilotmega")
            mavlink_logger.info("✅ Connected successfully!")

            # Wait for Heartbeat
            mavlink_logger.info("⏳ Waiting for heartbeat...")
            self.mav_conn.wait_heartbeat()
            self.last_heartbeat = time.time()
            mavlink_logger.info("✅ Heartbeat received!")

            # Set target info
            self.target_system = self.mav_conn.target_system
            self.target_component = self.mav_conn.target_component
            self.is_connected = True

            # Start heartbeat monitoring
            threading.Thread(target=self._check_heartbeat_timeout, daemon=True).start()

            return True

        except Exception as e:
            mavlink_logger.error(f"❌ WebSocket connection error: {e}")
            return False
    
------------------------------------------
#Code to handle MAVLink data sent from the Browser to the server
# This function is called when a new MAVLink message is received over websocket
@socketio.on('mavlinkdata')
def handle_mavlink_data(data):
    """Handle incoming MAVLink data from the client"""
    global mavlink_buffer

    try:
        validator.mav_conn.write(data)  # Send the data to mavlink recv
        # _message_loop is running in a separate thread and will handle the incoming data
        logger.debug(f"Received MAVLink message: {data}")
    except Exception as e:
        logger.error("error processing MAVLink data ")

PR #974, opened 2024-09-02
Add a .pylintrc file

amilcarlucas:


peterbarker:
How do the rules interact with the flake8 tests which we already do?

amilcarlucas:
They do not, pylint is much more compreensive than flake8 and it runs a lot slower. This file is just to run pylint localy for now.
Once more of the code gets linted, we can automate this and activate fail on errors.

amilcarlucas:
BTW better than flake8 is ruff

tridge:
ruff sounds like better option, pylint is too slow

PR #979, opened 2024-09-17
Update MavFTP code, add MavFTPfs example/tool

Williangalvani:
This mounts mavftp as a virtual filesystem, which allows for more "human" interfacing with it.

There are caveats. It is obviously slower than doing things on your own machine. but this does allow interfacing using any filebrowser, for instance.

So far the filesystem is mounted as read-only. as there's no mavftp interface to write to a sector of a file, we'd have to create an abstraction layer that caches the whole file...

[Screencast from 2024-09-17 12-21-47.webm](https://github.com/user-attachments/assets/60fcbc26-295a-419a-87e4-89f83f856005)

### DIFF
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -1,0 +1,35 @@
+name: Pylint
+
+on: [push]
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: ["3.8", "3.12"]
+    steps:
+
+    - name: Checkout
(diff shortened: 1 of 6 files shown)


amilcarlucas:
@peterbarker I need to run pylint localy to make sure it is still 100% clean (no regressions)

In hindsight it was a bad idea to reject pylint github workflow. I will create a new PR that adds a pylint workflow for just these and any other pylint clean files. 

Nice work @Williangalvani, really like the way you structured the OP and return enums.

tridge (approved):

PR #983, opened 2024-09-27
Generator: Lua: add support for checksum on decode

IamPete1:
This adds checking of the checksum on message decode. This allows AP to pass unknown messages which have not passed its checksum.


IamPete1:
I have added a change to swap the null character in the generated output for `string.char(0)` this upsets GitHub less when you try and commit the generated output.

tridge (approved):

PR #986, opened 2024-10-03
.github: Use / as a file path separator

amilcarlucas:
Fix the commit that got merged without passing CI


peterbarker (approved):

peterbarker:
A few tests to fix, but this is obviously going to be right at some stage :-)

PR #988, opened 2024-10-09
added rewrite_log.py example

tridge:


peterbarker (approved):

danbaggs:
@tridge Please forgive my ignorance here, but i'm trying to use this example as the base for some log rewrites. I need to rewrite the `GLOBAL_POSITION_INT` values (along with a couple of others) but I'm struggling to understand how it's happening (particularly with line 31). I've searched through the mavlink docs and i can't find any messages that use range and uncertainty. Would you mind explaining how to use this example in a more generic way?

PR #994, opened 2024-10-29
Update requirements.txt for serial connection

sethome2:
Although pymavlink supports the use of serial connections with a flight control board and utilizes a serial connection package, this information is not reflected in the requirements.txt file.

Simply running "pip install -r pymavlink/requirements.txt" and attempting to use the mavutil module over serial will not work as expected.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,7 @@
 fastcrc
 lxml>=3.6.0
 setuptools>=42
+pyserial>=3.5
 wheel>=0.37.1
 
 # dev dependencies:


amilcarlucas (approved):

amilcarlucas:
Can you please rebase and force push this?

peterbarker (commented):
As silly as this sounds, we don't actually require pyserial because it's not technically required to use pymavlink.  You can create other sorts of backends that don't involve serial.  This is also why the `import` for serial happens when you try to create a `mavserial` object.

If you can find somewhere to put this in as a "recommends" that would be better.

PR #996, opened 2024-12-06
DFReader: In dump verbose show hex value for bitmask fields

shancock884:
The aim of this PR is to add the hex value alongside the decimal value, when displaying bitmask fields with the "dump --verbose" command in MavExplorer, as I think it is easier to follow which bits are set in hex.

This is also more in-line with how bitmasks are shown from TLog files when using "dump --verbose".

For example, on XKF4:
```
> dump --verbose XKF4
2024-12-04 14:53:37.722: XKF4
    ...
    SS: 107327 (0x1a33f)
        ATTITUDE (attitude estimate valid)
        HORIZ_VEL (horizontal velocity estimate valid)
        VERT_VEL (vertical velocity estimate valid)
    ...
 ```

Tested locally with a variety of messages containing bitmasks.

### DIFF
--- a/DFReader.py
+++ b/DFReader.py
@@ -384,6 +384,11 @@ class DFMessage(object):
                 except UnicodeDecodeError:
                     f.write("    %s: %s" % (c, to_string(val)))
 
+            # If this is a bitmask, then append the hex value
+            if c in field_metadata_by_name:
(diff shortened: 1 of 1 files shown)


peterbarker:
... eg.  `0b01011101` would seem to be better still for your goal..

Issue #976, opened 2024-09-13
Please straighten out your supported Python versions and dependencies


sarnold:
I just came back to this after several years and there seems to be *zero progress* in terms of cleaning up Python version support or packaging.  Meaning `setup.py` *still requires future pkg* even though several other places claim only py36 or later (and several variants thereof).  This is getting ridiculous trying to maintain a package that won't let go of out-dated dependencies.  The world moves on and even bionic has been declared lifeless.  Please lurch forward into at least the early 21st century...

peterbarker:
> I just came back to this after several years and there seems to be _zero progress_ in terms of cleaning up Python version support or packaging. Meaning `setup.py` _still requires future pkg_ even though several other places claim only py36 or later 

We're not going out of our way to support Python2 any more - if it breaks it breaks.

But we are also not going out of our way to kill it right now in pymavlink.  That day may come - and it will be a good day!

> (and several variants thereof). This is getting ridiculous trying to maintain a package that won't let go of out-dated 

I think we may be unaware of problems that our laggardly behaviour may be causing other people here.

Are you attempting to create system packages or the like?

> dependencies. The world moves on and even bionic has been declared lifeless

.... python3 *is* a possibility on Bionic.

> Please lurch forward into at least the early 21st century...

Wait, who *doesn't* like steam trains?

bobpaul:
Am I missing something? I don't see any reference to `future` in the *.py files in the project. I just built the package with python13 and without installing `future`. I haven't done extensive testing (so maybe something is broken), but I was able to run the [sender/receiver examples from mavlink-routerd](https://github.com/mavlink-router/mavlink-router?tab=readme-ov-file#sender--receiver)

Is future only required for python2?

peterbarker:
Partially closed by https://github.com/ArduPilot/pymavlink/pull/1003 which officially says we're no longer doing Python2 and happy to actively remove support code.

peterbarker:
> Derp, I can be kinda dumb sometimes. I had the mavlink project checked out (with pymavlink as a submodule) and I was doing `git grep` from within `mavlink` instead of in `mavlink/pymavlink` :-/
> 
> Attached is a patch that I think should maintain python2 compatibility without requiring the future module in python 3.11+. I understand `sorted(mydict.items())` is a bit slower on python2.7 than using `iteritems()` from future, but it should still return the same result.
> 
[...]

PR #1000, opened 2024-12-24
Update mavgen_c to python3

jmackay2:
Most of the code was already compatible with python3 except for the small change for iteritems. The builtins and print_function were put in to make it python3 compatible, so those are no longer necessary.

### DIFF
--- a/generator/mavgen_c.py
+++ b/generator/mavgen_c.py
@@ -1,15 +1,10 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 '''
 parse a MAVLink protocol XML file and generate a C implementation
 
 Copyright Andrew Tridgell 2011
 Released under GNU GPL version 3 or later
 '''
-from __future__ import print_function
-from future.utils import iteritems
-
-from builtins import range
-from builtins import object
 
 import os
 from . import mavparse, mavtemplate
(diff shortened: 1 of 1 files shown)


bobpaul (changes requested):
I'm not a project maintainer, but from reading other comments on github, I understand pymavlink is not yet ready to intentionally remove python2 support.

`python-future` at least partially requires (for running the tests for sure) lib2to3 which was removed in python13. Removing pymavlink's dependence on the `future` module seems reasonable and certainly for this file can be done without breaking python2 compatibility.

bobpaul:
I had read that discussion, too, and it wasn't clear to me they had made a decision. Peter's comment a few months ago in #976 look clear to me that they still don't want us to intentionally remove Python2 support, just don't jump through hoops to maintain support:

>We're not going out of our way to support Python2 any more - if it breaks it breaks.

> But we are also not going out of our way to kill it right now in pymavlink. That day may come - and it will be a good day!

One of the things to keep in mind is that a lot of use of mavlink is on embedded devices in the air. While it's easy to get python3 support on our computers, getting python3 on an out of production intel edison or some other product that never had an official BSP for any embedded linux new enough to include python3 could be challenging for some users. (That's a situation I was in about 3 years ago).

tridge:
I think we are ready to be py3 only now

bobpaul (approved):
In light of tridge's comment, I think this looks fine as is and you can ignore my previous suggestions.

peterbarker:
Merged, thanks!

PR #1002, opened 2024-12-30
Mode:add Plane AUTOLAND mode

Hwurzburg:


mavutil.py
```diff
@@ -2148,6 +2148,7 @@ def mode_string_v09(msg):
     23 : 'QACRO',
     24 : 'THERMAL',
     25 : 'LOITERALTQLAND',
+    26 : 'AUTLOAND',
```
peterbarker, line 2151:
```suggestion
    26 : 'AUTOLAND',
```

peterbarker (changes requested):

mavutil.py
Hwurzburg: duh

mavutil.py
```diff
@@ -2148,6 +2148,7 @@ def mode_string_v09(msg):
     23 : 'QACRO',
     24 : 'THERMAL',
     25 : 'LOITERALTQLAND',
+    26 : 'AUT0LAND',
```
peterbarker, line 2151:
```suggestion
    26 : 'AUTOLAND',
```

peterbarker (changes requested):

mavutil.py
Hwurzburg: double duh

peterbarker (approved):

PR #1003, opened 2025-01-28
DFReader: change to python3 only

tridge:
I think we can drop py2 support now
overall performance improvement of about 30% in loading a bin log

for release, we should make it 2.5.0

### DIFF
--- a/DFReader.py
+++ b/DFReader.py
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 '''
 APM DataFlash log file reader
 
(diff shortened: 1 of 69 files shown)


DFReader.py
```diff
@@ -1122,15 +1089,16 @@ def rewind(self):
 
     def init_arrays(self, progress_callback=None):
         '''initialise arrays for fast recv_match()'''
-        self.offsets = []
-        self.counts = []
+        offsets = []
+        counts = []
+        formats = self.formats
```
peterbarker, line 1094: This would not seem to be related to the PR title?

I mean, probably not a bad change, but...

DFReader.py
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
peterbarker, line 1: So this doesn't *reeally* accomplish much, as DFReader is usually called from `mavlogdump.py` via `mavutil.py`.

We should probably change `mavlogdump.py`'s shebang line as part of this PR.

peterbarker (approved):
LGTM

tridge:
@peterbarker added change to py3 shebang for a tools/examples/modules

peterbarker:
> @peterbarker added change to py3 shebang for a tools/examples/modules

Nothing exceeds like excess :-)

PR #1005, opened 2025-02-17
mavwp: display DO_LAND_START paths as different edge colours and no "to" edge

peterbarker:
…to" edge

stops showing us flying to these edges

changes the colour to match the label colour

Before:
image

After:
image

Before:
image

After:
image


peterbarker:
I've tested this with a DLS at 0,0,0 and it is the same as master, showing an edge from 0,0,0. 

I don't have a clear idea of how to do better there ATM.

PR #1006, opened 2025-02-22
Generator: python: set display bitmask based on enum if no value is provided directly

IamPete1:
Because enums are represented as lists with no class it would be a big change to add the bitmask property to the enum. Instead this sets the existing display property on the field which uses the enum to "bitmask" if no value is directly provided and the enum is marked as a bitmask in the xml.

https://github.com/mavlink/mavlink/pull/2215
https://github.com/ArduPilot/mavlink/pull/383

As far as I can tell the other generators deal with this correctly already.

### DIFF
--- a/generator/mavgen_python.py
+++ b/generator/mavgen_python.py
@@ -518,16 +518,42 @@ def byname_hash_from_field_attribute(m, attribute):
     return ", ".join(strings)
 
 
-def generate_classes(outf, msgs, enable_type_annotations):
+def generate_classes(outf, msgs, enums, enable_type_annotations):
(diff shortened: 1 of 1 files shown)


IamPete1:
This is the diff in generated output. `AVAILABLE_MODES`, `VIDEO_STREAM_INFORMATION`, `AIRSPEED` and `STORM32_GIMBAL_MANAGER_CONTROL` all get a display bitmask flag. This is because they are cases where the xml "display=bitmask" was not set but the "bitmask=true" was set on the enum which they point to.

image

tridge (approved):

PR #1012, opened 2025-04-02
Add .bitmask attribute to enumerations

peterbarker:
... also add a sanity check that the bitmask values are valid - which fails on a lot of XML at the moment


hamishwillee (approved):
This looks great, but can you run int on the upstream master branch before merging please.

peterbarker:
> This looks great, but can you run it/succeed on the upstream master branch before merging please.

Nothing untoward happens running this branch against mavlink/master branch.  flake8 passes on the results file.

tridge (approved):

amilcarlucas:
Very nice one. Long overdue.

For added points: enums can not have a range attribute

PR #1022, opened 2025-04-12
mavutil.py: set flightmode, mav_type and base_mode only within sysid_…

peterbarker:
…state

if we just set the attribute then the setter will set the mnessage for the currently-configured target, rather than the source of the message

This work sponsored by Freespace Operations


tridge (approved):

peterbarker:
Tested during flights out at SV, no problems.

ArduPilot autotests also seem to be absolutely fine with the change, and fresh patches to use multiple vehicles in the ArduPilot test suite benefit from these changes.

PR #1023, opened 2025-04-14
[new] adding new unit of consumption: L/h

TomasTwardzik:
[new]
Adding a new consumption unit (L/h), that is better suited for large scale drones with Internal Combustion Engines.
Adding an already existing unit of volume litre "L". This is done to prevent breaking changes in various MAVLink CI runners that depend on unit definition using lower case "l" for the litre unit. Unit "l" is to be deprecated in the future.

### DIFF
--- a/generator/mavschema.xsd
+++ b/generator/mavschema.xsd
@@ -170,9 +170,11 @@
     <!-- flow -->
     <xs:enumeration value="g/min"/>    <!-- grams/minute -->
     <xs:enumeration value="cm^3/min"/> <!-- cubic centimetres/minute -->
+    <xs:enumeration value="L/h"/>      <!-- Litres/hour -->
     <!-- volume -->
     <xs:enumeration value="cm^3"/>     <!-- cubic centimetres -->
     <xs:enumeration value="l"/>        <!-- litres -->
+    <xs:enumeration value="L"/>        <!-- Litres -->
   </xs:restriction>
 </xs:simpleType>
 


amilcarlucas (changes requested):
Please remove white space changes

amilcarlucas:
@peterbarker can this be merged now?

amilcarlucas:
It does look better now than it did originally, so you are making progress.

TomasTwardzik:
Just a side question, why the trailing whitespaces? My Editor is removing them all the time for me, so it was slightly cumbersome to put them back in.

amilcarlucas:
preserve git history. So that the `git blame` command produces correct results.

TomasTwardzik:
Hello. 
I am still waiting for workflow and merge approval. I believe the white space issue was resolved.
Can you please have a look?

amilcarlucas:
Re-triggered workflows.

TomasTwardzik:
Hello.
Can we merge this, or is there some change that is blocking this?

generator/mavschema.xsd
```diff
@@ -169,6 +169,7 @@
     <!-- flow -->
     <xs:enumeration value="g/min"/>    <!-- grams/minute -->
     <xs:enumeration value="cm^3/min"/> <!-- cubic centimetres/minute -->
+    <xs:enumeration value="l/h"/>       <!-- litres/hour -->
```
peterbarker, line 172:
```suggestion
    <xs:enumeration value="L/h"/>       <!-- litres/hour -->
```

... SI FTW.

peterbarker (changes requested):

generator/mavschema.xsd
TomasTwardzik: Considering unit litre already exists and is not capitalized, wouldn't it make sense to keep this in line with that?

amilcarlucas: Good point, I would prefer to fix it in every place it is used.
L is the correct unit.

MaEtUgR: We're happy to change litres from l to L but it will cause CI to fail on many MAVLink messages and should probably be done separately from this pull request no?

amilcarlucas: My vote is to fix it all here. What do you think @peterbarker ?

TomasTwardzik:
@peterbarker Did I understand the assignment correctly by adding capitalized "Litre" for transition? This leads to eventually deprecating lower case "l" over time.

PR #1025, opened 2025-04-20
Remove Python2 support; inline typing markup into templates

peterbarker:
This removes support for generating Python2 bindings, and for running the generator under Python2.

Working with the non-inline type annotations was rather painful when modifying them recently.  It was a good hack, but we really shouldn't need it any more given how dead Python2 is now.

### DIFF
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -27,8 +27,8 @@ jobs:
 
     - name: Analysing the code known to be pylint clean
       run: |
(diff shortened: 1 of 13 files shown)


peterbarker:
I've tested the annotations by pip-installing from the source and `diff -ur` on the generated packages.

tridge (approved):

tridge:
just some testing with new pymavlink + mavproxy and old pymavlink, new flight controller build

peterbarker:
> just some testing with new pymavlink + mavproxy and old pymavlink, new flight controller build

Done

PR #1027, opened 2025-04-27
recv_match condition adapted to multi vehicle

noahredon:
The condition now verify the right message received from m.get_srcSystem() instead of that of self.sysid

### DIFF
--- a/mavutil.py
+++ b/mavutil.py
@@ -523,7 +523,7 @@ class mavfile(object):
                 return None
             if type is not None and not m.get_type() in type:
                 continue
-            if not evaluate_condition(condition, self.messages):
+            if not evaluate_condition(condition, self.sysid_state[m.get_srcSystem()].messages):
                 continue
             return m
 


peterbarker:
This does feel right to me.

How have you tested this?

My chief concern is that this is a base-class implementation (`mavfile.recv_match`).  Your patches here make `recv_match` specific to handling mavlink messages, and I don't see anything else in there currently which makes it specific to mavlink messages.

We use the same infrastructure for parsing messages of types other than MAVLink.  DFReader is a good example of that, it creates `DFMessage` objects.  In a slightly different world DFReader would only override `recv_msg`, not `recv_match`, in which case this patch would kill DFReader (DFMessage objects don't have `get_srcSystem`.

I can't find an in-tree example of this *actually* happening, but I think it's a reasonable concern.

At the very least we need a comment / apology to whoever this trips up where we are doing mavlink-message-specific things, and a plea to fix it.  More reasonable would be to check the message to see if it has a `get_srcSystem()` attribute before attempting to use it as a method.

noahredon:
Sorry for my inactivity, I fixed this problem.

peterbarker (approved):
LGTM.

I've rebased this to get rid of the merge commit.

PR #1030, opened 2025-05-20
DFReader bugfix

robertlong13:
Sorry about this.

### DIFF
--- a/dfindexer/dfindexer.c
+++ b/dfindexer/dfindexer.c
@@ -36,12 +36,17 @@ static void ensure_capacity(OffsetArray *arr) {
 }
 
 OffsetArray* scan_offsets(const uint8_t *data, size_t len,
-                          uint8_t fmt_type,
+                          uint8_t fmt_type, uint8_t fmt_length,
                           uint8_t type_offset, uint8_t length_offset,
                           uint8_t head1, uint8_t head2) {
     uint8_t lengths[NUM_TYPES] = {0};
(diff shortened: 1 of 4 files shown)


robertlong13:
Found a second bug too. I have a log where FMT of FMT isn't the first message in the log (it came second, after DCM, for some reason).

dfindexer/dfindexer.c
```diff
@@ -36,13 +36,18 @@ static void ensure_capacity(OffsetArray *arr) {
...
                           uint8_t type_offset, uint8_t length_offset,
                           uint8_t head1, uint8_t head2) {
     uint8_t lengths[NUM_TYPES] = {0};
     OffsetArray *results = calloc(NUM_TYPES, sizeof(OffsetArray));
     if (!results) panic("Memory allocation failed");
 
+    // Make an a-priori assumption about the FMT message length, since
+    // sometimes, somehow, the FMT for FMT ends up NOT being the first message
+    // in the file.
+    lengths[fmt_type] = fmt_length;
```
Ryanf55, line 49: Is this repeatable on the same log, or does it only occur every N times when you load the same log?

robertlong13: image

robertlong13: (i.e., it's repeatable; it's very clearly in the log file, not the code)

tridge (approved):

PR #1032, opened 2025-05-20
DFReader pandas module

robertlong13:
WIP

I was tempted to make this as a totally separate repo, but maybe it's the right fit here. Opening a draft PR in case anyone else is interested in it.

This is handy for writing analysis scripts and jupyter notebooks. It resamples all your fields of interest to a common time base and returns everything together as a pandas dataframe. I also have it cache that dataframe so that you can quickly rerun your scripts during development without needing to handle caching manually.

### DIFF
--- a/tools/dfreader_pandas.py
+++ b/tools/dfreader_pandas.py
@@ -88,7 +88,7 @@ def parse_log_data(reader: DFReader_binary, fields, frequency):
             rows.append(new_row(reader, fields))
 
     df = pd.DataFrame(rows)
-    df["timestamp"] = pd.to_datetime(df["timestamp"], unit="s")
+    df["timestamp"] = pd.to_datetime(df["timestamp"], unit="s", utc=True)
     df.set_index("timestamp", inplace=True)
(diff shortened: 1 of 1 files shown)


dfreader_pandas.py
```diff
@@ -0,0 +1,215 @@
+"""
+Module to parse ArduPilot logs into pandas DataFrames with optional caching.
```
tridge, line 2: should be in tools/ or examples/

tridge (changes requested):
i don't think this should be in top level

dfreader_pandas.py
robertlong13: Yeah, it's out of place at top level for sure. I do want it to be importable though, like `from pymavlink.tools.dfreader_pandas import parse_log_to_df`. I'll figure out how to do that. I think that means I need to add tools to packages, which now that I say it, is probably better than installing all these tools as scripts. We should be calling them like `python -m pymavlink.tools.magfit`

tridge (approved):

tridge:
@robertlong13 looks good, just needs to be out of draft to merge

PR #1035, opened 2025-05-21
CI: added python -m build to all CI

tridge:


robertlong13 (approved):

.github/workflows/test.yml
```diff
@@ -60,6 +60,9 @@ jobs:
         run: |
           python3 -m pip install .
 
+      - name: Build package
+        run: python -m build
```
robertlong13, line 64: This might need some extra pip installs first though

PR #1038, opened 2025-05-22
CI: ensure we build properly tagged wheels

tridge:


robertlong13:
@tridge, according to ChatGPT, his only builds wheels, not source distributions. Should we also include a `python -m build --sdist`?

robertlong13:
https://cibuildwheel.pypa.io/en/stable/deliver-to-pypi/#:~:text=Make%20your%20SDist%20with%20the,on%20your%20CI%20provider's%20options).

PR #1042, opened 2025-05-23
Saving send_callback after the use of auto_mavlink_version

noahredon:
In the function auto_mavlink_version, the save of the mav.callback was here, but not the one related to the mav.send_callback, so I added it.


peterbarker:
What problem does this fix?

noahredon:
I tried to set a callback and a send_callback before receiving any MAVLink message from the master, but when I received one, auto_mavlink_version was called and only my send_callback was erased. So I think it makes sense to add this

PR #1044, opened 2025-05-25
CI Speedup and More Wheels

robertlong13:
Brought the time from 30m to 8m, which I think we can live with. Also added Linux ARM, Windows, and Mac (x86, ARM, and universal). Funny enough, the bottleneck is still Linux x86. The Linux builds use docker, and the musl 32-bit wheel takes most of the time.

I've set it so that it only builds everything on push to master or release. PRs get only a fraction of the full build, and take about 3 minutes instead of 8. On PRs, I build Linux and Windows, 64-bit only, and for only 3.8 and 3.13 (which I deemed most likely to accidentally break in a PR). I could honestly leave all the python versions and all the OS and just disable 32-bit and it'd likely still be about 3 minutes, but I think I'll keep it the way I have it. We're still building everything after we merge, so we should be quick to catch any regressions. I also added the ability to manually trigger a run in case you see a PR that you think is risky enough to try a full build before merging.

The only wheels we're missing now are Windows ARM, because the Windows ARM runner doesn't come with enough stuff to compile lxml and lxml doesn't supply wheels WinARM. I'll just wait until lxml adds wheels for that, much to the dismay of all 0 people asking me for Windows ARM wheels.

I added concurrency rules so that subsequent PR pushes, or even rapid merging of several PRs, don't bog down CI.

Tested everything pretty thoroughly on my fork before opening.

### DIFF
--- a/.github/workflows/pylint.yml
+++ b/.github/workflows/pylint.yml
@@ -2,6 +2,10 @@ name: Pylint
 
 on: [push, pull_request]
 
+concurrency:
+  group: ci-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   build:
     runs-on: ubuntu-latest
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -14,38 +14,51 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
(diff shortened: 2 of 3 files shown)


tridge (approved):

tridge:
@robertlong13 thanks!

PR #1051, opened 2025-05-27
WLua: add custom dissector for FTP Mavlink message

shancock884:
This PR allows custom dissectors to be added for specific fields within the wireshark dissector, to allow more details on fields that have their own internal formatting.
The first example implemented is for the FILE_TRANSFER_PROTOCOL.payload field, and looks like this:

image

This has been tested both interactively/locally, and by a new snapshot test case.


peterbarker:
Merged, thanks!

PR #1055, opened 2025-06-03
tools: Add mavtranslatelog.py for updating tlog files

Maarrk:
This can be used to recover an existing `.tlog` file after introducing breaking changes in message definitions. For now it can handle changing the message id number, adding or removing fields, as long as the names don't change.

An example use case is when a custom dialect is requested to update the message ids into allocated range, like I needed to do in mavlink/mavlink#2272. This can be handled this way:
```sh
mavtranslatelog.py /path/to/old.tlog -f 5f24327765e5061acc3f17487ac12e441f067177 -F https://github.com/marsh-sim/mavlink.git -t dialect -T https://github.com/marsh-sim/mavlink.git
```

I do not know how often that happens in general (@hamishwillee may have comments?), thus whether it should be included in the package. If that's a small niche, no hard feelings if y'all prefer to reject this PR.

The `--help` output has a complete description of what's supported:
```
usage: mavtranslatelog.py [-h] [-o OUTPUT] [-f FROM] [-F FROM_REPOSITORY] [-t TO] [-T TO_REPOSITORY] [--dialect DIALECT] [--mav10] [-q] [-v] LOG

translate a MAVLink telemetry log (.tlog) file between incompatible versions of the protocol, mapping messages and fields by name between the two. Copyright Marek S. Łukasiewicz 2025 Released under GNU GPL version 3 or later

positional arguments:
  LOG

options:
  -h, --help            show this help message and exit
  -o OUTPUT, --output OUTPUT
                        output matching packets to given file, by default will append '_translated' to input file name
  -f FROM, --from FROM  git reference or xml file with definitions for input messages
  -F FROM_REPOSITORY, --from-repository FROM_REPOSITORY
                        local path or git clone URL for definitions, default github.com/mavlink/mavlink
  -t TO, --to TO        output messages, see '--from'
  -T TO_REPOSITORY, --to-repository TO_REPOSITORY
                        repository with output definitions, see '--from-repository'
  --dialect DIALECT     MAVLink dialect, ignored when specifying definitions in XML file, default all
  --mav10               parse as MAVLink1
  -q, --quiet           display less output
  -v, --verbose         display more output

Specify at least 'from' or 'to' definition, with the unspecified one assumed to be installed pymavlink.
```

### DIFF
--- a/tools/mavtranslatelog.py
+++ b/tools/mavtranslatelog.py
@@ -1,0 +1,181 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+
+'''
+translate a MAVLink telemetry log (.tlog) file between incompatible versions
+of the protocol, mapping messages and fields by name between the two.
+
+Copyright Marek S. Łukasiewicz 2025
+Released under GNU GPL version 3 or later
+'''
+
+from argparse import ArgumentParser
+import importlib
+import os
+import struct
+import subprocess
+import sys
+import tempfile
+
+import pymavlink
+from pymavlink.generator import mavgen
+from pymavlink.mavutil import mavlink_connection
+
+FROM_TO_REQUIRED = "Specify at least 'from' or 'to' definition, with the unspecified one assumed to be installed pymavlink."
+parser = ArgumentParser(description=__doc__, epilog=FROM_TO_REQUIRED)
+
+parser.add_argument("-o", "--output", help="output matching packets to given file, by default will append '_translated' to input file name", default=None)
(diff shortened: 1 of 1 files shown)


hamishwillee:
It is very uncommon for things to change in common.xml, standard.xml or minimal.xml, but it does happen. I think this might be useful thanks @Maarrk but it's not my call. I'm sure @peterbarker et al will look at it sometime soon.

peterbarker (approved):

peterbarker:
@Maarrk whoops - 3.7 doesn't like this for some reason.

PR #1060, opened 2025-06-17
XSD: Remove display attribute from fields

hamishwillee:
This removes the XSD to allow `display="bitmask"` on a field. We now expect the bitmask attribute to be set on the enum itself.

The trigger for this was the accidental reintroduction of the attribute in https://github.com/mavlink/mavlink/pull/2289/ 

My concern about this is that it might break other systems that aren't directly managed by MAVLink org.

### DIFF
--- a/generator/mavschema.xsd
+++ b/generator/mavschema.xsd
@@ -20,7 +20,6 @@
 <xs:attribute name="id" type="mavlinkMsgId"/> <!-- used in message elements -->
 <xs:attribute name="print_format" type="xs:string"/> <!-- used in field elements -->
 <xs:attribute name="enum" type="xs:string"/> <!-- used in field,param elements -->
-<xs:attribute name="display" type="xs:string"/> <!-- used in field elements -->
 <xs:attribute name="units" type="SI_Unit"/> <!-- this will get changed on the fly to xs:string if no strict-units command line option is used -->
 <xs:attribute name="multiplier" type="factor"/>
 <xs:attribute name="instance" type="xs:boolean"/>
@@ -243,7 +242,6 @@
         <xs:attribute ref="name" use="required"/>
         <xs:attribute ref="print_format" />
         <xs:attribute ref="enum" />
-        <xs:attribute ref="display" />
         <xs:attribute ref="units" />
         <xs:attribute ref="increment"/>
         <xs:attribute ref="minValue"/>


hamishwillee:
@peterbarker This works upstream, but we need get rid of the last instances here. 

But, the question stands, what about other dialects that don't talk to this project?

amilcarlucas (approved):

hamishwillee:
@peterbarker This is failing because the pymavlink repo still has a version of common that includes `display="bitmap"` in fields. These were purged upstream. I thought had been removed in ardupilot/mavlink. IFF that's the case, I think ArduPilot/pymavlink needs to update from ArduPilot/mavlink - right?

peterbarker:
On Tue, 30 Sep 2025, Hamish Willee wrote:

> @peterbarker This is failing because the pymavlink repo still has a version of common that includes display="bitmap" in fields. These were
> purged upstream. I thought had been removed in ardupilot/mavlink. IFF that's the case, I think ArduPilot/pymavlink needs to update from
> ArduPilot/mavlink - right?

pymavlink doesn't reference mavlink.  It's actually the other way 
around...

Perhaps a pymavlink release is required to fix the problem - where's your 
pymavlink coming from?

hamishwillee:
> On Tue, 30 Sep 2025, Hamish Willee wrote: @peterbarker This is failing because the pymavlink repo still has a version of common that includes display="bitmap" in fields. These were purged upstream. I thought had been removed in ardupilot/mavlink. IFF that's the case, I think ArduPilot/pymavlink needs to update from ArduPilot/mavlink - right?

> pymavlink doesn't reference mavlink. It's actually the other way around... Perhaps a pymavlink release is required to fix the problem - where's your pymavlink coming from?

I'm not sure what you mean. This repo is failing its tests. If it's taking upstream mavlink then it shouldn't.

peterbarker:
> I'm not sure what you mean. This repo is failing its tests. If it's taking mavlink/mavlink then it shouldn't.

image

.... so that's pretty clear.... do we need to pull patches back into ardupilot/mavlink to fix these attributes?  Do we have a PR?

Notionally we could check against both repositories.

hamishwillee:
> Notionally we could check against both repositories.

I thought we'd fixed all these in ardupilot/mavlink. Perhaps some crept in. We'll need a new PR I guess.

PR #1062, opened 2025-06-19
generator: Fix TypeError str must be encoded before hashing 

cclauss:
Fixes: #1081
* #1081

Detect and fix `TypeError: Strings must be encoded before hashing`.

How was this tested?
* Run the fixed old version and compare the results with the proposed version.
* `python3 generator/C/test/posix/sha256_test.py ` # Usage...
* `python3 generator/C/test/posix/sha256_test.py sha256`
* `python3 generator/C/test/posix/sha256_test.py a b ` # Uasge...
* `pytest generator/C/test/posix/sha256_test.py ` # One test passed
* `pytest generator ` # One test passed

### DIFF
--- a/generator/C/test/posix/sha256_test.py
+++ b/generator/C/test/posix/sha256_test.py
@@ -1,16 +1,35 @@
-#!/usr/bin/env python3
+import hashlib
+import sys
 
-from builtins import range
-
-import hashlib, sys
-
-h = hashlib.new('sha256')
-h.update(sys.argv[1])
-res = h.digest()[:6]
-for i in range(6):
-    sys.stdout.write("%02x " % ord(res[i]))
-sys.stdout.write("\n")
+def hash_sha256(s) -> bytes:
+    if isinstance(s, str):
+        s = s.encode("utf-8")
+    h = hashlib.new('sha256')
+    h.update(s)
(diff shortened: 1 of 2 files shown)


tpwrules:
This is broken, but it is not a pytest test.

What is the motivation for removing `pytest.ini` anyway? It is doing its job here.

cclauss:
We can also make this a test file merely by adding a `test_hash_sha256()` function to ensure that it works and detect any regressions.  `@pytest.mark.parametrize` could have been used, but I worried about adding that dependency here.

peterbarker:
Well, test change looks good.

@tpwrules are you convinced re. the pytest.ini deletion?

tpwrules:
The problem is definitely solved now so I'm happy with this PR on its face.

But it's implied precedent for our testing procedure that I'd like to be a bit more thoughtful about. Namely that every file that pytest finds should indeed be something pytest can use.

And that the test should hang around instead of just being deleted. I doubt sha256 will break soon :)

tpwrules:
I suspect this is actually intended to be a companion to a manual test of mavlink using `sha256_test.c` in the same directory. Do we have an automated test for that? If so let's delete both files.

PR #1069, opened 2025-06-30
.github: Let's drop the `--exit-zero` tests that are being ignored

cclauss:
https://flake8.pycqa.org/en/latest/user/options.html#cmdoption-flake8-exit-zero


khancyr:
That is wrong, it removes the tests ...

PR #1073, opened 2025-07-02
cibuildwheel updates

robertlong13:
This PR contains two fixes:

## Remove "lxml" from pyproject.toml
This is necessary to fix a [recent build failure](https://github.com/ArduPilot/pymavlink/actions/runs/16010298504). lxml just released a new version 6.0.0, which does not have a wheel for cp38-manylinux_aarch64, where the previous version did. The thing is, there's really no reason for us to need to pip install lxml into the isolated build environment; it's a runtime dependency only. `pyproject.toml` does not need this in there.

This also lets me build the windows-arm wheels now (surely few, if any, care about that yet).

I'm pretty sure it doesn't need "future" in there either, but I didn't bother removing that as it isn't hurting anything like the lxml thing is.

## Update cibuildwheel to 3.0.0
Split out from #1072

I went through the [change log](https://cibuildwheel.pypa.io/en/stable/changelog/) and it looks like the only change that impacts us was that they dropped 32-bit linux from the "auto" group. I have changed that to "auto64 auto32". They mentioned that they are considering doing the same to windows in the future, so I added that to the windows line too to cover us when that happens.

[I have tested the build on my fork, using workflow-dispatch to make sure all archs/versions build](https://github.com/robertlong13/pymavlink/actions/runs/16015744853)

### DIFF
--- a/.github/workflows/python-publish.yml
+++ b/.github/workflows/python-publish.yml
@@ -38,7 +38,7 @@ jobs:
                   ("linux-arm", "ubuntu-24.04-arm", "auto"),
                   ("macos", "macos-latest", "arm64 x86_64 universal2"),
                   ("windows", "windows-latest", "auto"),
-                  # ("windows-arm", "windows-11-arm", "auto"), # Can't do this one yet, as lxml doesn't have Windows ARM wheels, and the runner doesn't have libxml2 pre-installed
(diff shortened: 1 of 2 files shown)


peterbarker:
Thanks!

PR #1078, opened 2025-07-02
Tools: mavlogdump: preseve units allong with format when trimming logs

IamPete1:
Now were using instance messages extensively log parsers need the units to work out what should be shown as a instance.

### DIFF
--- a/tools/mavlogdump.py
+++ b/tools/mavlogdump.py
@@ -235,6 +235,9 @@ while True:
         if (isbin or islog) and m_type == "FMT":
             output.write(m.get_msgbuf())
             continue
+        if (isbin or islog) and m_type == "FMTU":
+            output.write(m.get_msgbuf())
+            continue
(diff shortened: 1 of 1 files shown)


tools/mavlogdump.py
```diff
@@ -235,6 +235,9 @@ def match_type(mtype, patterns):
         if (isbin or islog) and m_type == "FMT":
             output.write(m.get_msgbuf())
             continue
+        if (isbin or islog) and m_type == "FMTU":
```
peterbarker, line 238: I think you're supposed to pass in `--meta` to get this, `MULT` and `UNIT`, all of which are required for sensible display...

I reckon we probably should include those up here, however....
```suggestion
        if (isbin or islog) and m_type in ["FMTU", "MULT", "UNIT"]:
```

amilcarlucas: Nice, good catch

peterbarker (approved):

PR #1079, opened 2025-07-02
Spin2

RbtsEvrwhr-Riley:
Bug fixes. Needed to change an assignment to a BYTEMOVE in a couple places, and made the pack/send more efficient.


RbtsEvrwhr-Riley:
So I might have accidentally clicked "main" instead of "spin2" in that last pull request and broken some stuff...sorry, don't mix pull requests and dental work.

I plan to add more unit tests now that I know stuff is largely getting rubber stamped in the spin2 space, we should have more testers soon.

peterbarker (changes requested):
Please tidy the commit list.

In particular, we don't like Merge commits in the master branch.

PR #1088, opened 2025-07-10
mavextra: speedup radio_margin() with a LRU cache

tridge:
approx 10x faster, very useful for BVLOS planning
image


peterbarker (approved):
LGTM

annotations AKA black magic.

I'm a touch surprised we can't get away with an `@memoise` if it exists...

PR #1089, opened 2025-07-12
Replace cast(...) in generated Python code with type annotations

ntamas:
This patch removes the occurrences of `cast(some_type, expr)` in the generated Python code and replaces these expressions with type annotations. The patch increases the speed of the MAVLink parser from about 93K packets/sec to 111K packets/sec in my benchmarks on a Macbook Pro M1 while still maintaining type safety (all `mypy` checks pass).

The reason for the runtime cost of `cast(some_type, expr)` is that `cast()` is a proper function in Python that simply ignores the first argument and returns the second argument. `cast()` only has a special meaning for type checkers. However, `some_type` (the first argument) still needs to be _constructed_ when `cast()` is called even though it will be discarded immediately, and the cost is non-negligible for complex types like `Tuple[bytes, int, int, int, int, int, int, int, int]`. In contrast, inline type annotations used in this patch are completely stripped away when the source is compiled into bytecode so they have zero runtime cost.

### DIFF
--- a/generator/mavgen_python.py
+++ b/generator/mavgen_python.py
@@ -39,7 +39,7 @@ import struct
 import sys
 import time
 from builtins import object, range
-from typing import Any, Callable, Dict, Iterable, List, Mapping, Optional, Sequence, Tuple, Type, Union, cast
+from typing import Any, Callable, Dict, Iterable, List, Mapping, Optional, Sequence, Tuple, Type, Union
 
 WIRE_PROTOCOL_VERSION = "${wire_protocol_version}"
 DIALECT = "${DIALECT}"
@@ -200,7 +200,7 @@ class MAVLink_message(object):
 
     def format_attr(self, field: str) -> Union[str, float, int]:
         """override field getter"""
-        raw_attr = cast(Union[bytes, float, int], getattr(self, field))
+        raw_attr: Union[bytes, float, int] = getattr(self, field)
         if isinstance(raw_attr, bytes):
             return raw_attr.decode(errors="backslashreplace").rstrip("\\x00")
         return raw_attr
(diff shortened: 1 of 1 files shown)


ntamas:
CI failure in the `test pymavlink / build (3.7)` job is due to an outdated `mypy` in that environment; newer versions of `mypy` are smarter and accept the generated code without problems. Not sure whether it's worth trying to work around this issue - let me know what you would prefer.

ntamas:
> Are your packets/sec benchmarks in this repo, or is that custom code that you created? 

Custom code; not secret at all but it requires a large pickle file (about 68 MB) that contains actual signed MAVLink packets plus noise (fragmented packets, serial line noise etc) that we captured from ~1000 drones running at a same time on an UDP/IP network, and the benchmark kinda doesn't make sense without real data. I can publish it somewhere if needed. Usually the casts don't cause a problem in normal Python code, but this was special because it was on a hot path _and_ contained a complex type signature.

ntamas:
@peterbarker Done.

tridge:
@ntamas sorry, it has conflicts, I github is not allowing me to rebase it (maybe you unticked the allow maintainer edits button?)

ntamas:
> maybe you unticked the allow maintainer edits button?

No, I didn't, but it's a cross-organization pull request (our fork is under an org account) so GitHub does not show this checkbox.

Conflicts fixed - would you like me to rebase to get rid of the merge commit?

tridge:
> Conflicts fixed - would you like me to rebase to get rid of the merge commit?

yes please, and the PR still shows conflicts for me "This branch cannot be rebased due to conflicts"

PR #1090, opened 2025-07-12
DRAFT: Detect runtime-cast-value issues on generated code

cclauss:
Related to
* #1089

* [ ] Can we detect the issues raised in #1089?
* [ ] Can we disable mypy on legacy Python 3.7?
* [ ] Should we upgrade `target-version` in `.github/workflows/mypy-generated.ini`?

@ntamas


.github/workflows/flake8-generated.cfg
```diff
@@ -0,0 +1,11 @@
+# Enable different config settings than this repo's .flake8 file.
```
amilcarlucas, line 1:
```suggestion
# These flake8 settings apply only to auto-generated python code and are
# different than this repo's .flake8 file settings that apply to hand written code.
```

amilcarlucas (changes requested):

.github/workflows/flake8-generated.cfg
amilcarlucas: It's better to ignore as little a possible, and this PR does exactly that. Thanks.

cclauss:
```suggestion
# These flake8 settings apply only to auto-generated Python code and are
# different than this repo's .flake8 file settings that apply to hand written code.```

**Review by peterbarker — APPROVED** [review]
LGTM.

Would be nice to thin that list at some stage, too :-)

**Comment by cclauss** [comment]
My favorite is:
```
max-line-length = 5519
```
Should we try to `ruff format` the generated code to see if it can wrap some lines to be shorter than a football field?

PR #1091, opened 2025-07-12
fixed CRC bypass issue

### LINKED ISSUE #1019: Failing CRCs can sometimes pass due to signing
I believe I may have found a bug in the mavlink library.  It starts with this buffer:

uint8_t TestBuffer[]={0xFD,0x1C,0x00,0x00,0xFD,0x01,0x01,0x1E,0x00,0x00,0xE5,0x87,0x10,0x00,0x7c,0xdd,0x8c,0xba,0xfb,0xa6,0xa6,0xba,0x74,0x5a,0xdc,0xbc,0x98,0x97,0x99,0xb9};

Suppose that in the previous message one or two bytes is missed for whatever reason, this means that the starting 0xFD and 0x1C will be consumed as the end of the prior message and ultimately lost due to a CRC fail.  A new message will start on 0xFD,0x01,0x01, which means it has an incompatibility flag of 0x01 meaning it is a signed message.  Using mavlink_char_frame_buffer we end up here:

```
case MAVLINK_PARSE_STATE_GOT_BAD_CRC1:
        if (status->parse_state == MAVLINK_PARSE_STATE_GOT_BAD_CRC1 || c != (rxmsg->checksum >> 8)) {
            // got a bad CRC message
            status->msg_received = MAVLINK_FRAMING_BAD_CRC;
        } else {
            // Successfully got message
            status->msg_received = MAVLINK_FRAMING_OK;
        }
        rxmsg->ck[1] = c;

        if (rxmsg->incompat_flags & MAVLINK_IFLAG_SIGNED) {
            status->parse_state = MAVLINK_PARSE_STATE_SIGNATURE_WAIT;
            status->signature_wait = MAVLINK_SIGNATURE_BLOCK_LEN;

            // If the CRC is already wrong, don't overwrite msg_received,
            // otherwise we can end up with garbage flagged as valid.
            if (status->msg_received != MAVLINK_FRAMING_BAD_CRC) {
                status->msg_received = MAVLINK_FRAMING_INCOMPLETE;
            }
        } else {
            if (status->signing &&
                   (status->signing->accept_unsigned_callback == NULL ||
                    !status->signing->accept_unsigned_callback(status, rxmsg->msgid))) {

                // If the CRC is already wrong, don't overwrite msg_received.
                if (status->msg_received != MAVLINK_FRAMING_BAD_CRC) {
                    status->msg_received = MAVLINK_FRAMING_BAD_SIGNATURE;
                }
            }
            if (r_message != NULL) {
                memcpy(r_message, rxmsg, sizeof(mavlink_message_t));
            }
        }
        break;

```
status->msg_received=MAVLINK_FRAMING_BAD_CRC is set, but because the incompat_flags is set to 1 we end up setting status->parse_state = MAVLINK_PARSE_STATE_SIGNATURE_WAIT;
This means that 13 more chars are consumed and we end up here:

```
case MAVLINK_PARSE_STATE_SIGNATURE_WAIT:
        rxmsg->signature[MAVLINK_SIGNATURE_BLOCK_LEN-status->signature_wait] = c;
        status->signature_wait--;
        if (status->signature_wait == 0) {
            // we have the whole signature, check it is OK
#ifndef MAVLINK_NO_SIGNATURE_CHECK
            MAVLINK_START_SIGN_STREAM(status->signing->link_id);
            bool sig_ok = mavlink_signature_check(status->signing, status->signing_streams, rxmsg);
            MAVLINK_END_SIGN_STREAM(status->signing->link_id);
#else
            bool sig_ok = true;
#endif
            if (!sig_ok &&
                   (status->signing->accept_unsigned_callback &&
                    status->signing->accept_unsigned_callback(status, rxmsg->msgid))) {
                // accepted via application level override
                sig_ok = true;
            }
            if (sig_ok) {
                status->msg_received = MAVLINK_FRAMING_OK;
            } else {
                status->msg_received = MAVLINK_FRAMING_BAD_SIGNATURE;
            }
            if (r_message !=NULL) {
                memcpy(r_message, rxmsg, sizeof(mavlink_message_t));
            }
        }
        break;

```
signing is by default set to 0 so mavlink_signature_check just returns true.  When true status->msg_received is set to MAVLINK_FRAMING_OK and whatever garbage was in the frame is happily passed along as a valid message.  

Any string of garbage bytes starting with 0xFD,0xXX,0x01 will pass as a valid message.

tridge:
This fixes an issue where a corrupt packet can bypass the CRC check

fixes #1019 

many thanks to @greg-draganfly for reporting this!


peterbarker (approved):
LGTM

greg-draganfly:
Looks good.

PR #1093, opened 2025-07-12
delete unused arg error_limit of mavgen.py

oneWayOut:
The arg 'error_limit' of [mavgen.py](https://github.com/ArduPilot/pymavlink/blob/master/generator/mavgen.py) is not used anymore. So it could be deleted.


peterbarker (approved):
LGTM

PR #1099, opened 2025-07-23
mavgen_python.py: correct attribute used to retrieve link ID

peterbarker:
see post_message in mavutil.py

Busted since 2016 :-)


peterbarker:
Wrong fix; we have _link which is the inbound-to-MAVProxy-bottonm-2-bits information and _link_id which is the signing-outbound-link.

PR #1103, opened 2025-07-29
allow javascript generator to work with both node and browser

tridge:
This builds on the work by @IamPete1 in #995 and gets the generated javascript to pass tests, plus work in both a browser and in command line node tools

example usage: https://github.com/ArduPilot/WebTools/pull/265

also adds wss support to mavutil.py, so mavproxy can use wss: connections

### DIFF
--- a/generator/javascript/test/mavlink10.js
+++ b/generator/javascript/test/mavlink10.js
@@ -5,7 +5,6 @@ var    fs = require('fs');
 // Actual data stream taken from APM.
 global.fixtures = global.fixtures || {};
 global.fixtures.serialStream = fs.readFileSync("test/capture.mavlink");
-//global.fixtures.heartbeatBinaryStream = fs.readFileSync("javascript/test/heartbeat-data-fixture");
 
 describe("Generated MAVLink 1.0 protocol handler object", function() {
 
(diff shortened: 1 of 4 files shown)


tridge:
@Huibean could you test this with your browser efforts?

Huibean:
@tridge I just have a try using mavlink.js generated in this pr, the program freeze up every time after connect websocket, tomorrow will digging into it
image

tridge:
> something wrong inside mavlinkProcessor.send, tomorrow will digging into it

ok, thanks!

tridge:
> there is a dead loop here

does it keep getting messages?

Huibean:
> > there is a dead loop here
> 
> does it keep getting messages?

I set a breakpoint there once enter while never return, then memory leak and crash

Huibean:
image
runtime error logs is going into the parse buffer, I temporary fix it by checking BAD_DATA, now I will fix the "Pack" import error and continue
```MAVLink20Processor.prototype.parseBuffer = function(s) {

    // Get a message, if one is available in the stream.
    var m = this.parseChar(s);

    // No messages available, bail.
    if ( null === m ) {
        return null;
    }

    // While more valid messages can be read from the existing buffer, add
    // them to the array of new messages and return them.
    var ret = [];
    ret.push(m);
    while(true) {
        m = this.parseChar(null);
        if ( null === m || m._id === mavlink20.MAVLINK_MSG_ID_BAD_DATA) {
            // No more messages left.
            return ret;
        }
        ret.push(m);
    }

}
```

tridge:
discussed with @Huibean, working now

PR #1111, opened 2025-08-06
setup: detect if python dev package is installed

tridge:
and build without fast indexing if not

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -113,6 +113,35 @@ with open("README.md", "r", encoding = "utf-8") as fh:
 
 ext_modules = []
 
+def test_python_h_available():
+    import tempfile
+    import distutils.ccompiler
+    import distutils.sysconfig
+    import sysconfig
+
+    with tempfile.TemporaryDirectory() as tmpdir:
+        test_file = os.path.join(tmpdir, "test_python_h.c")
+        with open(test_file, "w") as f:
+            f.write("#include <Python.h>\nint main() { return 0; }")
+
+        try:
+            compiler = distutils.ccompiler.new_compiler()
+            distutils.sysconfig.customize_compiler(compiler)
+
+            # Add Python include dir (e.g. /usr/include/python3.11 or C:\... on Windows)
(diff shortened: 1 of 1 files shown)


robertlong13:
This detection doesn't seem to work correctly on Windows, so it's breaking cibuildwheel.

tridge:
> This detection doesn't seem to work correctly on Windows, so it's breaking cibuildwheel.

should be fixed now I hope, changed to compile only, not link

robertlong13:
Can you do a workflow_dispatch on Build and Optionally Publish for this branch? That will trigger a full build of every wheel we make instead of the reduced list we do for PRs triggers

tridge:
> Nevermind, testing it now on my fork

great, thanks! CI on main repo is glacially slow today

tridge:
@robertlong13 looks like it passed!

robertlong13 (approved):
LGTM

robertlong13:
Now that I think of it (not to change my mind on my approval), since our wheel build is so comprehensive, we probably don't even need fast-index to build by default on Linux. Very few people will be running the build, and it's easy enough for those who both
1. need to build from source
2. care about fast log analysis
can just set the environment variable and be done with it.

At the time I introduced this, I didn't fully understand the details of distribution and thought a lot more users would need to actually compile.

tridge:
> It only matters for people who both

it's breaking CI in ArduPilot for an attempt to update to the latest pymavlink, so this change is worth it just for that I think
see https://github.com/ArduPilot/ardupilot/pull/30675

robertlong13:
> it's breaking CI in ArduPilot for an attempt to update to the latest pymavlink, so this change is worth it just for that I think

Right, your PR nicely fixes that CI while still leaving my fast indexer compiled by default on Linux. Alternatively, we could have just made the fast indexing only ever build when explicitly forced to. Everyone who installs the official release wheels would still get it, and people like you and I would just set PYMAVLINK_FAST_INDEX=1 in our .bashrc and be done with it.

PR #1118, opened 2025-08-16
gitignore: remove generator/C/ from ignored files

Bogdanov-am:
This PR updates .gitignore so that the generator/C directory is no longer ignored and will be included in the source distribution (sdist).

Currently, files under generator/C are skipped during sdist creation because of the ignore rules, which results in an incomplete source package. This can cause issues when installing from source (e.g. from PyPI or Linux distributions) where the full generator code is required.

With this change, the generator sources will always be shipped as part of the distribution, ensuring consistent and reproducible builds.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,7 +5,6 @@ build*
 .nfs*
 share/
 __pycache__
-generator/C/
 generator/CPP11/include_v2.0
 generator/python/
 generator/message_definitions


tridge:
@Bogdanov-am I checked the .tar.gz download from pypi and it does have the files

Bogdanov-am:
I have a project that uses C++ and pybind11. The project https://github.com/ArduPilot/mavlink is added to it as a submodule to generate MAVLink header files during the build process. The build is performed using https://github.com/pypa/build. During the build, I get the following error: `FileNotFoundError: [Errno 2] No such file or directory: '/tmp/***/thirdparty/mavlink/pymavlink/generator/C/include_v2.0/protocol.h'`. When exploring the temporary build directory, the generator files are indeed not found there. I have tried many options, and for a reason that is still unclear to me, specifically modifying the .gitignore file fixes my problem. If desired, I can provide a link to my open-source project.

PR #1119, opened 2025-08-24
mavlogdump.py follow argument non-functional

rrhan0:
This argument in mavlogdump does not appear to do anything if the file is a .log, .bin, or .csv
https://github.com/ArduPilot/pymavlink/blob/03210f40551378e84c776d96b42c2842d608e4df/tools/mavlogdump.py#L28

When the log file is initialized, it returns a DFReader or CSVReader
https://github.com/ArduPilot/pymavlink/blob/03210f40551378e84c776d96b42c2842d608e4df/tools/mavlogdump.py#L81-L85
https://github.com/ArduPilot/pymavlink/blob/03210f40551378e84c776d96b42c2842d608e4df/mavutil.py#L2063-L2089

The blocking arg in recv_match is unused in DFReader and CSVReader.
https://github.com/ArduPilot/pymavlink/blob/03210f40551378e84c776d96b42c2842d608e4df/tools/mavlogdump.py#L221
https://github.com/ArduPilot/pymavlink/blob/03210f40551378e84c776d96b42c2842d608e4df/DFReader.py#L1032-L1050
https://github.com/ArduPilot/pymavlink/blob/03210f40551378e84c776d96b42c2842d608e4df/CSVReader.py#L165-L179

I am not sure if there is any other purpose for the follow argument in mavlogdump. If there is a purpose, I feel like a warning would be good. If there isn't, maybe it should be removed?


peterbarker:
Good description.

Just tested this locally and it does actually work for tracking a live file.

If I run a simulation (and leave it running)
```
pbarker@crun:~/rc/ardupilot(master)$ ./Tools/autotest/sim_vehicle.py -v Plane --gdb --debug
```
Then `mav.tlog` is created:
```
-rw-rw-r-- 1 pbarker pbarker 313686 Aug 24 10:57 mav.tlog
```
If I `mavlogdump.py mav.tlog` then it dumps the contents of the file and exits.

If I add `--follow` then it dumps the contents of the file, but does not exit and continues to dump content as `mavproxy` appends data to `mav.tlog`.

So I do believe the parameter is functional and useful.

rrhan0:
It may work in that case, but that wasn't exactly what I was trying to point out. It was probably because I was trying to do something weird with the dataflash logs by using mavproxy to write dataflash logs live and trying to read the .bin with the blocking argument, which wouldn't work. Not sure if anyone else will try doing that, but a warning would hopefully stop them.

peterbarker (approved):
LGTM

Issue #1133, opened 2025-09-29
Error loading generated WLua Wireshark plugin


yonatan8070:
I'm generating a MAVLink dissector for Wireshark from my custom dialect, which includes `common.xml`. In Wireshark 4.4.8, I'm now getting this error on startup from my dialect:

```Lua: Error during loading:
/home/yonatan/.local/lib/wireshark/plugins/MY_DISSECTOR.lua:6915: bad argument #5 to 'new' (ProtoField_new: Base must be either base.DEC or base.UNIT_STRING)
stack traceback:
	[C]: in field 'new'
	/home/yonatan/.local/lib/wireshark/plugins/MY_DISSECTOR.lua:6915: in main chunk
	[C]: in function 'require'
```

This error originates from the `CAMERA_IMAGE_CAPTURED` message, for which the generated output looks like this, I added a comment on the problematic line:

```lua
f.CAMERA_IMAGE_CAPTURED_time_boot_ms = ProtoField.new("time_boot_ms (uint32_t) [ms]", "mavlink_proto.CAMERA_IMAGE_CAPTURED_time_boot_ms", ftypes.UINT32, nil)
f.CAMERA_IMAGE_CAPTURED_time_utc = ProtoField.new("time_utc (uint64_t) [us]", "mavlink_proto.CAMERA_IMAGE_CAPTURED_time_utc", ftypes.UINT64, nil)
f.CAMERA_IMAGE_CAPTURED_camera_id = ProtoField.new("camera_id (uint8_t)", "mavlink_proto.CAMERA_IMAGE_CAPTURED_camera_id", ftypes.UINT8, nil)
f.CAMERA_IMAGE_CAPTURED_lat = ProtoField.new("lat (int32_t) [degE7]", "mavlink_proto.CAMERA_IMAGE_CAPTURED_lat", ftypes.INT32, nil)
f.CAMERA_IMAGE_CAPTURED_lon = ProtoField.new("lon (int32_t) [degE7]", "mavlink_proto.CAMERA_IMAGE_CAPTURED_lon", ftypes.INT32, nil)
f.CAMERA_IMAGE_CAPTURED_alt = ProtoField.new("alt (int32_t) [mm]", "mavlink_proto.CAMERA_IMAGE_CAPTURED_alt", ftypes.INT32, nil)
[...]

PR #1134, opened 2025-09-29
mavgen_lua.py: copy with MAV_BOOL / int8 type bitmask

peterbarker:
This looks like it would die any time you have an enumeration which isn't of type float, double or int32.... no idea why you would have enumerations in those first two types...

Closes https://github.com/ArduPilot/pymavlink/issues/1133

@shancock884 does this look like it might work to you?


shancock884:
> @shancock884 does this look like it might work to you?

I'll make a quick test for this later today to cover the range of data types - for example, I wonder whether INT16/64 should be included too, given that INT32 and INT8 both are.

Having FLOAT can make sense because some MAV_CMD params are float enums, but not sure I've seen any double enums - although if its not explicitly disallowed by Mavlink rules, parser should be able to cope with it.

shancock884:
I added a suggestion above so that it should be able to handle any cases of bitmasks being attached to any sort of signed integer types in the future.

It seems very odd to me that MAV_BOOL is a bitmask rather than a straight-up enum... but given that there seems to be lots of discussion about on the original PR (https://github.com/mavlink/mavlink/pull/2242), I will leave that can of worms alone!

The net result of that decision is that it looks kind-of silly in Wireshark, as it is extracted like this:
```
param1: Arm (MAV_BOOL): 0x00000000 (0)
    ...0 = MAV_BOOL_TRUE: False
```
Instead of a normal enum, which would look like this:
```
param1: Arm (MAV_BOOL): MAV_BOOL_FALSE (0)
```

peterbarker:
@yonatan8070 @shancock884 I've applied Simon's suggestions here.  What sort of testing needs to be done here, @shancock884 ?

@yonatan8070 could you at least test this patch still solves your problem, please?

PR #1136, opened 2025-10-08
Fix use of Set() in mavftp

andyp1per:
This is not allowed in newer versions of python

### DIFF
--- a/mavftp.py
+++ b/mavftp.py
@@ -984,7 +984,7 @@ class MAVFTP:  # pylint: disable=too-many-instance-attributes
         if file_size % self.write_block_size != 0:
             write_blockcount += 1
 
-        self.write_list = Set(range(write_blockcount))
+        self.write_list = set(range(write_blockcount))
         self.write_acks = 0
         self.write_total = write_blockcount
         self.write_idx = 0


mavftp.py
```diff
@@ -984,7 +984,7 @@ def cmd_put(
         if file_size % self.write_block_size != 0:
             write_blockcount += 1
 
-        self.write_list = Set(range(write_blockcount))
+        self.write_list = set(range(write_blockcount))
```
tridge: please remove the import on line 28

peterbarker:
@amilcarlucas can you weigh in on this, please?

It's breaking things for Andy, so we're very inclined to merge this!

andyp1per:
> Shouldn't you also remove importing Set?

Done, thanks

peterbarker (changes requested):
Needs to pass CI.

On the bright side, it has highlighted another code path which would have broken your setup when it was crossed :-)

andyp1per:
I think this just demonstrates how time-consuming invalid requests for changes can be. Turns out the import is required by the reference to ```Set[int]`` which is both valid (its using typing.Set in the correct way) on 3.12 and required by the code. On 3.9 or later the lower case form can be used but it is not required and of course this would break 3.8 compatibility. So I am back at my original change - which is the one I tested - and is correct.

peterbarker:
Merged, thanks!

PR #1143, opened 2025-10-30
Revert "Mavgen C: Use strncpy instead of memcpy to pack char[n] fields"

julianoes:
This reverts commit ff57508f739f49bf09e8ef3f523f17a44478d6c5 from https://github.com/ArduPilot/pymavlink/pull/922.

The change broke the `param_value` field of `PARAM_EXT_...` messages which is defined as `char[128]` and can contain arbitrary data such as int/float or binary data.

Ideally, the type would be defined as `uint8_t[128]` and the issue avoided, however, this can't be changed at this stage as it would change the CRC_EXTRA and make the message incompatible.

Unfortunately, I don't see another way to fix this than to revert this change.

@shancock884, @hamishwillee

### DIFF
--- a/generator/C/include_v0.9/protocol.h
+++ b/generator/C/include_v0.9/protocol.h
@@ -169,6 +169,13 @@ static inline void byte_copy_8(char *dst, const char *src)
 */
 static void mav_array_memcpy(void *dest, const void *src, size_t n)
 {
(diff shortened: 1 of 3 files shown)


peterbarker:
@julianoes I'd like to see a second patch in this PR adding a comment as to *why* we're using memcpy - so we don't break your toys again.

PR #1144, opened 2025-11-05
XSD: Add remove_after option to deprecated tag

hamishwillee:
This follows on from #809 last week, allowing an optional `removal_date` to be added to deprecated tags. This is optional - often we will deprecate things but not be ready to decide a hard date up front. 

This is suggestion by PeterB in https://github.com/ArduPilot/pymavlink/pull/809#pullrequestreview-3398470519 

@peterbarker Was this what you were thinking? Or were you thinking of replacing `since` with `removal_date` in deprecated?

### DIFF
--- a/generator/mavschema.xsd
+++ b/generator/mavschema.xsd
@@ -73,6 +73,7 @@
   </xs:simpleType>
 </xs:attribute>
 <xs:attribute name="replaced_by" type="xs:string"/> <!-- used in elements: superseded, deprecated -->
+<xs:attribute name="remove_on_date" type="xs:string"/> <!-- optional in elements: deprecated -->
 
 <!-- mavlink message IDs are unsigned 24-bit values. -->
 <xs:simpleType name="mavlinkMsgId" id="mavlinkMsgId">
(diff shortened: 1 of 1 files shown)


generator/mavschema.xsd
```diff
@@ -73,6 +73,7 @@
   </xs:simpleType>
 </xs:attribute>
 <xs:attribute name="replaced_by" type="xs:string"/> <!-- used in elements: superseded, deprecated -->
+<xs:attribute name="remove_after" type="xs:string"/> <!-- optional in elements: deprecated -->
```
hamishwillee:
```suggestion
<xs:attribute name="removal_date" type="xs:string"/> <!-- optional in elements: deprecated -->
```

peterbarker (approved):

hamishwillee:
> Please keep the commit list neat - unfortunately the github interface doesn't really allow for that (I was tempted to use it myself but resisted!)

What precisely do you mean by "commit list"?

What I normally do is a squash merge. If there are multiple commits you end up with a single commit named after the PR and an extended commit message that includes notes on the squashed commits. 

I think you are suggesting to just manually squash commits and give them a good name?

hamishwillee:
Oh, and thank you for this!

PR #1145, opened 2025-11-05
Add missing include

Ryanf55:
Messages use types such as uint8_t but those are not defined unless you include cstdint.


peterbarker (changes requested):
Single patch, please.

peterbarker (approved):
LGTM

PR #1163, opened 2026-01-28
feat: add new base MAVlink dissector port association for TCP

TomasTwardzik:
Adding a new TCP port association in a generated Lua dissector.
The TCP port is an Auterion-specific MAVLink-router TCP endpoint. Adding this will ease our development and debugging.

### DIFF
--- a/generator/mavgen_wlua.py
+++ b/generator/mavgen_wlua.py
@@ -55,7 +55,7 @@ def get_field_info(field):
 
 def generate_preamble(outf):
     print("Generating preamble")
-    t.write(outf, 
+    t.write(outf,
 """
 -- Wireshark dissector for the MAVLink protocol (please see https://mavlink.io/en for details)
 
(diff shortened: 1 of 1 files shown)


generator/mavgen_wlua.py
```diff
@@ -718,6 +718,7 @@ def generate_epilog(outf):
 -- register common Mavlink TCP ports
 
 DissectorTable.get("tcp.port"):add("5760-5763", mavlink_proto)
+DissectorTable.get("tcp.port"):add("5790", mavlink_proto)
```
hamishwillee: @peterbarker It seems reasonable to me that we would make it easier for people to debug endpoints like this, given it has little or no cost to others - right? But suggest this because it isn't a "Common" port.

```suggestion

-- register vendor-specific Mavlink TCP ports

DissectorTable.get("tcp.port"):add("5790", mavlink_proto) # Auterion MAVlink-router TCP endpoing
```

PR #1192, opened 2026-03-30
Fix: specify UTF-8 encoding when opening XML files in mavgen

youziLab:
Fix: specify UTF-8 encoding when opening XML files in mavgen
potential encoding issues when opening XML files on different operating systems (especially Windows).

### DIFF
--- a/generator/mavgen.py
+++ b/generator/mavgen.py
@@ -56,7 +56,7 @@ def mavgen(opts, args):
     if opts.validate:
         try:
             from lxml import etree
-            with open(schemaFile, 'r') as f:
+            with open(schemaFile, 'r', encoding='utf-8') as f:
                 xmlschema_root = etree.parse(f)
                 if not opts.strict_units:
                     # replace the strict "SI_Unit" list of known unit strings with a more generic "xs:string" type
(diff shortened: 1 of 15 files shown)


peterbarker:
How was this tested?

cclauss (approved):
I approve of these changes.

> it's a touch surprising we have to do this in 2026!

Python v3.15 (to be released in October 2026) will finally ___uses [UTF-8](https://en.wikipedia.org/wiki/UTF-8) as the default encoding, independent of the system’s environment.___
https://docs.python.org/3.15/whatsnew/3.15.html?utm_source=chatgpt.com#other-language-changes

----
To consider for a ___separate___ pull request.... A few of these files use the recommended `with open(filename) as:` syntax, which automatically closes file handles even when exceptions are raised.  Unfortunately, many others may be leaving file handles open.  This can be detected via the command: [`ruff check --select=SIM115`](https://docs.astral.sh/ruff/rules/open-file-with-context-handler)

peterbarker:
> I approve of these changes.

Thanks!

> > it's a touch surprising we have to do this in 2026!
> Python v3.15 (to be released in October 2026) will finally _**uses [UTF-8](https://en.wikipedia.org/wiki/UTF-8) as the default encoding, independent of the system’s environment.**_ https://docs.python.org/3.15/whatsnew/3.15.html?utm_source=chatgpt.com#other-language-changes

Yay!  So we ought to be able to use that by, say, 2030 ;-)

peterbarker:
> In my case, I've completely replaced the original XML definitions because I’m applying MAVLink to industrial equipment via RS232/RS485, rather than traditional drones or flight controllers. Even in 2026, it's still surprisingly hard to find a robust protocol tool as good as MAVLink for general UART communication.

Huh!  Interesting, thanks for mentioning this, it's absolutely a from-left-field sort of thing for me and I'm sure tridge.

peterbarker:
Merged, thanks!

PR #1197, opened 2026-04-06
Replace deprecated argument with its new name

hunt0r:
A tiny improvement: `git clone --recurse-submodules` is preferred now, replacing old `--recursive`.

Related to https://github.com/ArduPilot/ardupilot/pull/32690

**Reviewers** I only tested this by confirming that I spelled it correctly. If that's insufficient, please either reject the PR or contribute additional desired testing to the PR via comments/editing this description.

### DIFF
--- a/tools/mavtranslatelog.py
+++ b/tools/mavtranslatelog.py
@@ -84,7 +84,7 @@ def protocol_module(identifier, repository, dialect):
                     repo_path = tempfile.mkdtemp()
                     if args_verbosity >= 1:
                         print("Cloning from", repository, "to", repo_path)
-                    subprocess.check_call(['git', 'clone', '--recursive', repository, repo_path])
+                    subprocess.check_call(['git', 'clone', '--recurse-submodules', repository, repo_path])
 
             assert os.path.isdir(repo_path), "Has a local copy of mavlink repository"
             try:


peterbarker:
> **Reviewers** I only tested this by confirming that I spelled it correctly. If that's insufficient, please either reject the PR or contribute additional desired testing to the PR via comments/editing this description.

So I think what you're saying here is that the change does potentially break things, and the original wording is extremely unlikely to break, ever.  So it is cosmetic.

This is a change that could take perhaps a minute to make and PR.  You are then asking someone *else* to sit down and work out how to test the thing, which is probably 30 minutes of work.

This is *not* really a good contribution when put in those terms, correct?

hunt0r:
> the change does potentially break things

In the sense that every change always "potentially" breaks things: yes.

> the original wording is extremely unlikely to break, ever.

I don't know how to estimate the likelihood that git will ever start rejecting the deprecated naming of the argument. If you estimate git's willingness to fully delete a deprecated thing (breaking the workflows of folks using the deprecated version) as "extremely unlikely", that's fine with me.

> So it is cosmetic.

If you dismiss "using the preferred way over the deprecated-but-works way" as "cosmetic", this is cosmetic.
(I don't personally chose that label for that case.)

> This is a change that could take perhaps a minute to make and PR. You are then asking someone _else_ to sit down and work out how to test the thing, which is probably 30 minutes of work.

Let's explicitly state the core question: "For this specific change, what constitutes _enough_ testing?"

Because this is a 1-liner, 1-name, targeted change, I personally am comfortable with "visual" testing. AKA: Not testing. That is: Looking at the change, understanding how simple it is, understanding that the chance it breaks things is _extremely_ low, and not doing/requiring further testing.

Your statement "You are then asking someone _else_ to sit down and work out how to test the thing, which is probably 30 minutes of work" seems to imply:
- You are not comfortable with "visual" testing (aka Not Testing).
- The fact that "visual" testing (aka Not Testing) is not acceptable should have been obvious to me.

> This is _not_ really a good contribution ... correct?

The most important statement _seems_ to be only implied. Let me attempt stating it explicitly:
"I, your reviewer, do not agree that 'visual testing' (aka Not Testing) is sufficient in this specific case."

@peterbarker Is that what you're saying?

peterbarker (approved):
LGTM

Issue #882, opened 2023-12-02
recv_match() error over UDP connection


jordancormack:
I've been using pymavlink to successfully send and receive mavlink messages using a serial connection, and I am trying to move to UDP.

When I try to receive a message using `msg = master.recv_match()` I get the following error:

"data, new_addr = self.port.recvfrom(UDP_MAX_PACKET_LEN)
OSError: [WinError 10022] An invalid argument was supplied"

However, if I send a message immediately before (e.g. `master.mav.heartbeat_send(...)` I don't get an error, and a burst of messages are successfully received, before a pause, and another burst of messages.

DevankGarg:
hi, did u figure out this ?

musluogluu:
I need help for this topic immediately. What do you mean by "I send a message immediately before (e.g. master.mav.heartbeat_send(...)" Can anyone help ???

1Blademaster:
Bumping this; I've ran into the same error while on Python 3.11.3.

```
DEBUG:fgcs:Trying to setup master with port udpout:127.0.0.1:14550 and baud 115200
ERROR:fgcs:Error while waiting for heartbeat: [WinError 10022] An invalid argument was supplied
Traceback (most recent call last):
  File "E:\projects\FGCS\radio\app\drone.py", line 175, in __init__
    heartbeat = self.master.recv_match(
                ^^^^^^^^^^^^^^^^^^^^^^^
  File "E:\projects\FGCS\radio\venv\Lib\site-packages\pymavlink\mavutil.py", line 518, in recv_match
    m = self.recv_msg()
        ^^^^^^^^^^^^^^^
  File "E:\projects\FGCS\radio\venv\Lib\site-packages\pymavlink\mavutil.py", line 1122, in recv_msg
    s = self.recv()
        ^^^^^^^^^^^
  File "E:\projects\FGCS\radio\venv\Lib\site-packages\pymavlink\mavutil.py", line 1080, in recv
    data, new_addr = self.port.recvfrom(UDP_MAX_PACKET_LEN)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
OSError: [WinError 10022] An invalid argument was supplied
```

peterbarker:
@stephendade have you ever seen this?  Looks to be Windows-specific error message.

The odd thing here is that MAVProxy should hit this all the time if you're doing udp - perhaps it's udp-in vs udp-out codepaths?

[...]

PR #1206, opened 2026-04-20
mavutil: fix missing socket.bind under Windows in udpout connections

stephendade:
Fix for #882.

Under Windows, a udpout connection needs to be first initialised with ``socket.bind(0.0.0.0, port)``, to avoid a ``OSError: [WinError 10022] An invalid argument was supplied`` error.

If a packet is transmitted before a ``recv_match()``, this error does not occur as a data transmission seemingly calls ``socket.bind()`` anyway.

Tested on Windows with a CubeOrange (via serial->udpout MAVProxy)


peterbarker (approved):
LGTM

PR #1210, opened 2026-05-01
mavutil: fix interpret_px4_mode for OFFBOARD on PX4 v1.12+ (#793)

### LINKED ISSUE #793: mavutil.py: interpret_px4_mode requires a simple fix for OFFBOARD mode in PX4 version 1.12 and higher
OFFBOARD mode is incorrectly reported as UNKNOWN by mavutil.py in the interpret_px4_mode function when using with PX4 version 1.12 and higher.  The function looks like this:
```
def interpret_px4_mode(base_mode, custom_mode):
    custom_main_mode = (custom_mode & 0xFF0000)   >> 16
    custom_sub_mode  = (custom_mode & 0xFF000000) >> 24

    if base_mode & mavlink.MAV_MODE_FLAG_MANUAL_INPUT_ENABLED != 0: #manual modes
        if custom_main_mode == PX4_CUSTOM_MAIN_MODE_MANUAL:
            return "MANUAL"
        elif custom_main_mode == PX4_CUSTOM_MAIN_MODE_ACRO:
            return "ACRO"
        elif custom_main_mode == PX4_CUSTOM_MAIN_MODE_RATTITUDE:
            return "RATTITUDE"
        elif custom_main_mode == PX4_CUSTOM_MAIN_MODE_STABILIZED:
            return "STABILIZED"
        elif custom_main_mode == PX4_CUSTOM_MAIN_MODE_ALTCTL:
            return "ALTCTL"
        elif custom_main_mode == PX4_CUSTOM_MAIN_MODE_POSCTL:
            return "POSCTL"
    elif (base_mode & auto_mode_flags) == auto_mode_flags: #auto modes
        if custom_main_mode & PX4_CUSTOM_MAIN_MODE_AUTO != 0:
            if custom_sub_mode == PX4_CUSTOM_SUB_MODE_AUTO_MISSION:
                return "MISSION"
            elif custom_sub_mode == PX4_CUSTOM_SUB_MODE_AUTO_TAKEOFF:
                return "TAKEOFF"
            elif custom_sub_mode == PX4_CUSTOM_SUB_MODE_AUTO_LOITER:
                return "LOITER"
            elif custom_sub_mode == PX4_CUSTOM_SUB_MODE_AUTO_FOLLOW_TARGET:
                return "FOLLOWME"
            elif custom_sub_mode == PX4_CUSTOM_SUB_MODE_AUTO_RTL:
                return "RTL"
            elif custom_sub_mode == PX4_CUSTOM_SUB_MODE_AUTO_LAND:
                return "LAND"
            elif custom_sub_mode == PX4_CUSTOM_SUB_MODE_AUTO_RTGS:
                return "RTGS"
	    elif custom_sub_mode == PX4_CUSTOM_SUB_MODE_OFFBOARD:
		return "OFFBOARD"
    return "UNKNOWN"
```

But I believe it should be this, note the final elif is moved to the left:
```
def interpret_px4_mode(base_mode, custom_mode):
    custom_main_mode = (custom_mode & 0xFF0000)   >> 16
    custom_sub_mode  = (custom_mode & 0xFF000000) >> 24

    if base_mode & mavlink.MAV_MODE_FLAG_MANUAL_INPUT_ENABLED != 0: #manual modes
        if custom_main_mode == PX4_CUSTOM_MAIN_MODE_MANUAL:
            return "MANUAL"
        elif custom_main_mode == PX4_CUSTOM_MAIN_MODE_ACRO:
            return "ACRO"
        elif custom_main_mode == PX4_CUSTOM_MAIN_MODE_RATTITUDE:
            return "RATTITUDE"
        elif custom_main_mode == PX4_CUSTOM_MAIN_MODE_STABILIZED:
            return "STABILIZED"
        elif custom_main_mode == PX4_CUSTOM_MAIN_MODE_ALTCTL:
            return "ALTCTL"
        elif custom_main_mode == PX4_CUSTOM_MAIN_MODE_POSCTL:
            return "POSCTL"
    elif (base_mode & auto_mode_flags) == auto_mode_flags: #auto modes
        if custom_main_mode & PX4_CUSTOM_MAIN_MODE_AUTO != 0:
            if custom_sub_mode == PX4_CUSTOM_SUB_MODE_AUTO_MISSION:
                return "MISSION"
            elif custom_sub_mode == PX4_CUSTOM_SUB_MODE_AUTO_TAKEOFF:
                return "TAKEOFF"
            elif custom_sub_mode == PX4_CUSTOM_SUB_MODE_AUTO_LOITER:
                return "LOITER"
            elif custom_sub_mode == PX4_CUSTOM_SUB_MODE_AUTO_FOLLOW_TARGET:
                return "FOLLOWME"
            elif custom_sub_mode == PX4_CUSTOM_SUB_MODE_AUTO_RTL:
                return "RTL"
            elif custom_sub_mode == PX4_CUSTOM_SUB_MODE_AUTO_LAND:
                return "LAND"
            elif custom_sub_mode == PX4_CUSTOM_SUB_MODE_AUTO_RTGS:
                return "RTGS"
    elif custom_sub_mode == PX4_CUSTOM_SUB_MODE_OFFBOARD:
	return "OFFBOARD"
    return "UNKNOWN"
```

This is because in PX4 version 1.12 and higher, as shown in file HEARTBEAT.hpp, base_mode is no longer used for offboard processing. Offboard does not use the `vehicle_control_mode.flag_control_auto_enabled` flag, offboard has its own flag, `vehicle_control_mode.flag_control_offboard_enabled` which does not set base_mode or auto_mode_flags.  So when they are checked in interpret_px4_mode in mavutil, this check will fail returning false.  I believe the fix is to simply move the PX4_CUSTOM_SUB_MODE_OFFBOARD check to the top level of the function, as I showed above, since it is a unique value of 0.

As a result of this bug, control software seeking to check "flightmode" using pymavlink will receive an UNKNOWN mode when the platform is actually in OFFBOARD mode.

FabianDumitrascu:
Closes #793. PX4's [`px4_custom_mode.h`][1] is the authoritative spec for
HEARTBEAT.custom_mode, but `mavutil.interpret_px4_mode` ignores it and
gates on `base_mode` flags instead -- which PX4 v1.12+ no longer sets
reliably for OFFBOARD. Result: OFFBOARD heartbeats decode as `"UNKNOWN"`.

[1]: https://github.com/PX4/PX4-Autopilot/blob/v1.14.3/src/modules/commander/px4_custom_mode.h

## Fix

Dispatch on `custom_main_mode` first (matches QGroundControl, MAVSDK, and
mavros, all of which copy `px4_custom_mode.h` from PX4 firmware). Only
descend into `custom_sub_mode` for `MAIN_MODE_AUTO`. `base_mode` stays in
the signature for API stability but is no longer consulted.

This is the inverse of the existing `px4_map` encoder table in the same
file -- the two sides now agree by inspection.

## Test plan

- [x] Added `tests/test_interpret_px4_mode.py` -- regression test for #793
      plus a contract test confirming `base_mode` is ignored.
- [x] `python3 -m pytest tests/test_interpret_px4_mode.py` passes.
- [x] `flake8 . --select=E9,F63,F7,F82` clean.
- [x] Single rebased commit, no merge or fixup commits.

### DIFF
--- a/mavutil.py
+++ b/mavutil.py
@@ -2538,40 +2538,41 @@ px4_map = { "MANUAL":        (mavlink.MAV_MODE_FLAG_CUSTOM_MODE_ENABLED | mavlin
 
 
 def interpret_px4_mode(base_mode, custom_mode):
+    # Dispatch on custom_main_mode (authoritative per PX4's px4_custom_mode.h);
+    # base_mode is no longer reliable for mode identity on PX4 v1.12+ (#793).
+    del base_mode
     custom_main_mode = (custom_mode & 0xFF0000)   >> 16
     custom_sub_mode  = (custom_mode & 0xFF000000) >> 24
 
-    if base_mode & mavlink.MAV_MODE_FLAG_MANUAL_INPUT_ENABLED != 0: #manual modes
-        if custom_main_mode == PX4_CUSTOM_MAIN_MODE_MANUAL:
-            return "MANUAL"
-        elif custom_main_mode == PX4_CUSTOM_MAIN_MODE_ACRO:
-            return "ACRO"
-        elif custom_main_mode == PX4_CUSTOM_MAIN_MODE_RATTITUDE:
-            return "RATTITUDE"
-        elif custom_main_mode == PX4_CUSTOM_MAIN_MODE_STABILIZED:
-            return "STABILIZED"
-        elif custom_main_mode == PX4_CUSTOM_MAIN_MODE_ALTCTL:
-            return "ALTCTL"
-        elif custom_main_mode == PX4_CUSTOM_MAIN_MODE_POSCTL:
-            return "POSCTL"
-    elif (base_mode & auto_mode_flags) == auto_mode_flags: #auto modes
-        if custom_main_mode & PX4_CUSTOM_MAIN_MODE_AUTO != 0:
-            if custom_sub_mode == PX4_CUSTOM_SUB_MODE_AUTO_MISSION:
-                return "MISSION"
-            elif custom_sub_mode == PX4_CUSTOM_SUB_MODE_AUTO_TAKEOFF:
-                return "TAKEOFF"
-            elif custom_sub_mode == PX4_CUSTOM_SUB_MODE_AUTO_LOITER:
-                return "LOITER"
-            elif custom_sub_mode == PX4_CUSTOM_SUB_MODE_AUTO_FOLLOW_TARGET:
-                return "FOLLOWME"
-            elif custom_sub_mode == PX4_CUSTOM_SUB_MODE_AUTO_RTL:
-                return "RTL"
-            elif custom_sub_mode == PX4_CUSTOM_SUB_MODE_AUTO_LAND:
-                return "LAND"
-            elif custom_sub_mode == PX4_CUSTOM_SUB_MODE_AUTO_RTGS:
-                return "RTGS"
-            elif custom_sub_mode == PX4_CUSTOM_SUB_MODE_OFFBOARD:
-                return "OFFBOARD"
+    if custom_main_mode == PX4_CUSTOM_MAIN_MODE_MANUAL:     
+        return "MANUAL"
+    elif custom_main_mode == PX4_CUSTOM_MAIN_MODE_ACRO:       
+        return "ACRO"
+    elif custom_main_mode == PX4_CUSTOM_MAIN_MODE_RATTITUDE:  
+        return "RATTITUDE"
+    elif custom_main_mode == PX4_CUSTOM_MAIN_MODE_STABILIZED: 
+        return "STABILIZED"
+    elif custom_main_mode == PX4_CUSTOM_MAIN_MODE_ALTCTL:     
+        return "ALTCTL"
+    elif custom_main_mode == PX4_CUSTOM_MAIN_MODE_POSCTL:     
+        return "POSCTL"
+    elif custom_main_mode == PX4_CUSTOM_MAIN_MODE_OFFBOARD:   
+        return "OFFBOARD"
+    elif custom_main_mode == PX4_CUSTOM_MAIN_MODE_AUTO:
+        if custom_sub_mode == PX4_CUSTOM_SUB_MODE_AUTO_MISSION:       
+            return "MISSION"
+        elif custom_sub_mode == PX4_CUSTOM_SUB_MODE_AUTO_TAKEOFF:       
+            return "TAKEOFF"
+        elif custom_sub_mode == PX4_CUSTOM_SUB_MODE_AUTO_LOITER:        
+            return "LOITER"
+        elif custom_sub_mode == PX4_CUSTOM_SUB_MODE_AUTO_FOLLOW_TARGET: 
+            return "FOLLOWME"
+        elif custom_sub_mode == PX4_CUSTOM_SUB_MODE_AUTO_RTL:           
+            return "RTL"
+        elif custom_sub_mode == PX4_CUSTOM_SUB_MODE_AUTO_LAND:          
+            return "LAND"
+        elif custom_sub_mode == PX4_CUSTOM_SUB_MODE_AUTO_RTGS:          
+            return "RTGS"
     return "UNKNOWN"
 
 def mode_mapping_byname(mav_type):
--- a/tests/test_interpret_px4_mode.py
+++ b/tests/test_interpret_px4_mode.py
@@ -1,0 +1,27 @@
+#!/usr/bin/env python3
+
+
+"""Unit tests for mavutil.interpret_px4_mode -- regression coverage for #793."""
+
+import unittest
+
+from pymavlink import mavutil
+
+
+class InterpretPX4ModeTest(unittest.TestCase):
+
+    def test_offboard_decoded_correctly(self):
+        # #793: pre-fix this returned "UNKNOWN" because PX4 v1.12+ stopped
+        # setting MAV_MODE_FLAG_AUTO_ENABLED in base_mode for OFFBOARD.
+        custom_mode = mavutil.PX4_CUSTOM_MAIN_MODE_OFFBOARD << 16
+        assert mavutil.interpret_px4_mode(0, custom_mode) == "OFFBOARD"
+
+    def test_base_mode_ignored(self):
+        # Locks in the post-fix contract that base_mode is no longer consulted.
+        custom_mode = mavutil.PX4_CUSTOM_MAIN_MODE_OFFBOARD << 16
+        assert mavutil.interpret_px4_mode(0x00, custom_mode) == "OFFBOARD"
+        assert mavutil.interpret_px4_mode(0xFF, custom_mode) == "OFFBOARD"
+
+
+if __name__ == '__main__':
+    unittest.main()


peterbarker:
What versions of PX4 will we break when we merge this, and how old are they?

peterbarker:
@hamishwillee FYI

FabianDumitrascu:
> What versions of PX4 will we break when we merge this, and how old are they?

**TL;DR:** No PX4 version breaks. Only OFFBOARD changes — from `"UNKNOWN"` back to `"OFFBOARD"` on PX4 v1.12+ (Apr 2021). Every other return string is byte-identical to today's behavior on every PX4 version since v1.0 (Sep 2014).

## When and why OFFBOARD broke

In **PX4 v1.11 and earlier**, [`src/modules/mavlink/mavlink_messages.cpp` lines 256-259](https://github.com/PX4/PX4-Autopilot/blob/v1.11.0/src/modules/mavlink/mavlink_messages.cpp#L256-L259) built `base_mode` with a big `switch (status->nav_state)` and the OFFBOARD case explicitly OR'd in `auto_mode_flags`:

```cpp
case vehicle_status_s::NAVIGATION_STATE_OFFBOARD:
    *mavlink_base_mode |= auto_mode_flags;            // = AUTO_ENABLED | STABILIZE_ENABLED | GUIDED_ENABLED
    custom_mode->main_mode = PX4_CUSTOM_MAIN_MODE_OFFBOARD;
    break;
```

That bit pattern is byte-identical to pymavlink's `auto_mode_flags`, so the `(base_mode & auto_mode_flags) == auto_mode_flags` gate in `interpret_px4_mode` passed and OFFBOARD decoded correctly (via a dead-code branch — `custom_sub_mode == 0` happens to match `PX4_CUSTOM_SUB_MODE_OFFBOARD = 0`).

In **PX4 v1.12+**, the heartbeat path was refactored. [`src/modules/mavlink/streams/HEARTBEAT.hpp` lines 79-99](https://github.com/PX4/PX4-Autopilot/blob/v1.14.3/src/modules/mavlink/streams/HEARTBEAT.hpp#L79-L99) now derives `base_mode` from `vehicle_control_mode` rather than from a per-nav-state switch:

```cpp
if (vehicle_control_mode.flag_control_auto_enabled) {
    base_mode |= MAV_MODE_FLAG_AUTO_ENABLED | MAV_MODE_FLAG_STABILIZE_ENABLED | MAV_MODE_FLAG_GUIDED_ENABLED;
}
```

And [`src/modules/commander/ModeUtil/control_mode.cpp` lines 115-130](https://github.com/PX4/PX4-Autopilot/blob/v1.14.3/src/modules/commander/ModeUtil/control_mode.cpp#L115-L130) — the new authoritative populator of `vehicle_control_mode` — gives OFFBOARD its own dedicated `flag_control_offboard_enabled` and **never sets `flag_control_auto_enabled`**:

```cpp
case vehicle_status_s::NAVIGATION_STATE_OFFBOARD:
    vehicle_control_mode.flag_control_offboard_enabled = true;
    // ... position/velocity/attitude flags from offboard_control_mode setpoint ...
    // (flag_control_auto_enabled is never set)
```

Post-refactor, OFFBOARD heartbeats no longer carry `auto_mode_flags`, pymavlink's gate fails, and the decoder falls through to `return "UNKNOWN"`. This is what issue #793 reports.

## Why this is a pymavlink-side fix, not a PX4-side fix

The PX4 refactor is correct: OFFBOARD really isn't an auto mode and getting its own dedicated control flag is the right design. The accidental coupling was on pymavlink's side — `interpret_px4_mode` had been using `base_mode` flags as a coarse pre-filter ahead of looking at `custom_mode`, and PX4 was never obligated to keep that pre-filter satisfied.

Per the [MAVLink common message spec](https://github.com/mavlink/mavlink/blob/master/message_definitions/v1.0/minimal.xml), `HEARTBEAT.custom_mode` carries no `enum=` attribute and is documented as *"A bitfield for use for autopilot-specific flags"* — i.e., the protocol explicitly defers decoding to the autopilot. PX4's [`px4_custom_mode.h`](https://github.com/PX4/PX4-Autopilot/blob/v1.14.3/src/modules/commander/px4_custom_mode.h) is the canonical spec, and the authoritative state lives entirely in `custom_mode.data` (bits 16-23 = main_mode, bits 24-31 = sub_mode). QGroundControl, MAVSDK, and mavros all decode this way — they ignore `base_mode` and read `custom_main_mode` directly. pymavlink's same-file `px4_map` encoder side already does it this way too. Only the decoder was relying on the `base_mode` heuristic.

## Why no PX4 version breaks

In the [old function](https://github.com/ArduPilot/pymavlink/blob/3cedcf10/mavutil.py#L2540-L2575), `base_mode` only gates *whether* we reach the lookup — every `return` is picked by `custom_main_mode`/`custom_sub_mode` alone. There is no `(main_mode, sub_mode)` pair where two `base_mode` values produce two different return strings. So removing the gate keeps every existing return identical and only converts some former `"UNKNOWN"`s into real names.

PX4's `custom_mode` bit layout and the enum values this decoder uses (`MAIN_MODE_MANUAL=1` … `STABILIZED=7`, `SUB_MODE_AUTO_*` 1-8) have been **byte-identical since [v1.0.0](https://github.com/PX4/PX4-Autopilot/blob/v1.0.0/src/modules/commander/px4_custom_mode.h) (Sep 2014)** — verified across [v1.0.0](https://github.com/PX4/PX4-Autopilot/blob/v1.0.0/src/modules/commander/px4_custom_mode.h), [v1.5.0](https://github.com/PX4/PX4-Autopilot/blob/v1.5.0/src/modules/commander/px4_custom_mode.h), [v1.10.0](https://github.com/PX4/PX4-Autopilot/blob/v1.10.0/src/modules/commander/px4_custom_mode.h), [v1.11.0](https://github.com/PX4/PX4-Autopilot/blob/v1.11.0/src/modules/commander/px4_custom_mode.h), [v1.12.0](https://github.com/PX4/PX4-Autopilot/blob/v1.12.0/src/modules/commander/px4_custom_mode.h), [v1.14.3](https://github.com/PX4/PX4-Autopilot/blob/v1.14.3/src/modules/commander/px4_custom_mode.h), and [main](https://github.com/PX4/PX4-Autopilot/blob/main/src/modules/commander/px4_custom_mode.h). Values were appended (RATTITUDE in v1.5, PRECLAND in v1.8, ORBIT in v1.9), never renumbered.

## Summary table

| PX4 version | base_mode includes auto_mode_flags for OFFBOARD? | pre-PR result | post-PR result |
|---|---|---|---|
| v1.0 – v1.11 | Yes (explicit `\|= auto_mode_flags`) | `"OFFBOARD"` | `"OFFBOARD"` |
| v1.12 – main | No (OFFBOARD only sets `flag_control_offboard_enabled`) | `"UNKNOWN"` | `"OFFBOARD"` |

All other modes: pre-PR result == post-PR result, on every PX4 version.

peterbarker:
@squilter this was your code initially - almost a decade ago now :-)

The AI is convincing when just glancing at its arguments, but I'm pretty damned sure this won't have actually been tested on older PX4.

I'd like someone who deals with PX4 on a regular basis to weigh in before I merge this - unless I see testing evidence.

jeroenzwan:
To weigh in on this, there was never a PX4_CUSTOM_SUB_MODE_OFFBOARD going back to v1.1.0 of PX4. This just happened to resolve to OFFBOARD, because PX4_CUSTOM_MAIN_MODE_OFFBOARD equates to 6 and PX4_CUSTOM_MAIN_MODE_AUTO to 4 resolving `if custom_main_mode & PX4_CUSTOM_MAIN_MODE_AUTO != 0: ` to True. With no custom sub mode specified, i.e. 0 for the case of just setting a main mode as in OFFBOARD mode, and PX4_CUSTOM_SUB_MODE_OFFBOARD being defined as 0, `elif custom_sub_mode == PX4_CUSTOM_SUB_MODE_OFFBOARD: ` also resolves to True. This returns OFFBOARD purely by coincidence.

Going all the way back to [PX4 v1.1.0](https://github.com/PX4/PX4-Autopilot/blob/63da175b3fbd9f912b3332df01b628363a29e238/src/modules/commander/px4_custom_mode.h) (v1.0.0 gives the warning "This commit does not belong to any branch on this repository, and may belong to a fork outside of the repository."), the enum on PX4 side has always been appended and not altered. The PX4 side has always had offboard mode defined as PX4_CUSTOM_MAIN_MODE_OFFBOARD (6), so it should be backwards compatible to the very start.

I think the base_mode flag doesn't add any useful catch coming from PX4 anyway. The mode is fully defined by the main mode and sub mode. Now it just breaks OFFBOARD resolving, because it fails the sub_mode check which was corrected in v1.12 as our friend Claude pointed out.

What might be nice is to add all the new sub modes and main modes that were added in newer PX4 versions (PX4_CUSTOM_MAIN_MODE_SIMPLE, PX4_CUSTOM_MAIN_MODE_TERMINATION, PX4_CUSTOM_SUB_MODE_AUTO_READY, PX4_CUSTOM_SUB_MODE_AUTO_PRECLAND, PX4_CUSTOM_SUB_MODE_AUTO_VTOL_TAKEOFF, PX4_CUSTOM_SUB_MODE_POSCTL_POSCTL, PX4_CUSTOM_SUB_MODE_POSCTL_ORBIT, PX4_CUSTOM_SUB_MODE_POSCTL_SLOW), making this PR more future proof.

peterbarker:
@jeroenzwan I take it that you're happy for this to go in as-is, then?

jeroenzwan:
I am happy to have it go in as is, but it might be nice to have the new messages added as well. So I would propose to add the new messages before merging the PR.

peterbarker:
> I am happy to have it go in as is, but it might be nice to have the new messages added as well. So I would propose to add the new messages before merging the PR.

Thanks - but perfect is the enemy of better, and we're not running out of PR numbers :-)

@FabianDumitrascu - feel free to raise another PR if you'd like to add the new stuff in.

peterbarker:
Merged, thanks!

PR #1216, opened 2026-06-05
mavgen_c: Helper functions to get MAV_CMD metadata and range check params

hamishwillee:
MAVLink command parameters (p1–p7) carry numeric values whose valid ranges are defined in the MAVLink XML — for example, a heading param may only accept 0–360°, or a speed param may require a non-negative value.
Validating these ranges at the MAVLink layer lets a vehicle or GCS reject malformed commands an mission items early, before they reach the flight stack, reducing the risk of undefined behaviour caused by out-of-range inputs.

This PR provides efficient range checking in a generated mavgen_c header `mav_cmd_helpers.h`.
These can be used by any flight stack to check inputs.

In addition to explicit range checking, it also provides methods for checking that lat/lon values are in valid ranges.

Note that what is actually range-checked depends on the library you build. Docs in https://github.com/ArduPilot/pymavlink/pull/1216/changes#diff-0009f780fff194d866e51fab6ff08d0765c0182b1a2ffb0b85651ff34bc9dde3 and in-source.


docs/cmd_range_checking.md
```diff
@@ -0,0 +1,176 @@
...
+
+// cmd_flags_table[]
+{ 179, CMD_FLAG_HAS_LOCATION }, // DO_SET_HOME
+```
+
+---
+
+### Unified handler (`COMMAND_INT`, `MISSION_ITEM_INT`, and `COMMAND_LONG`)
+
+In practice both PX4 and ArduPilot normalise all three message types into a single command handler before range-checking.
```
peterbarker: What?  No.....

hamishwillee: Yes, I know their is some bullshit in the doc. I'm hoping the code looks sensible.

hamishwillee: Fixed up all that crap

hamishwillee:
Note, fixes by Julian look good to me - they are individually explained in each commit.

hamishwillee:
This is the header built for common - note, others don't have mav_cmd, so no header built.

[mav_cmd_helpers.zip](https://github.com/user-attachments/files/29274798/mav_cmd_helpers.zip)

PR #1218, opened 2026-06-22
🍎 mavutil: auto-detect serial ports on macOS

drnic:
Add /dev/cu.usbmodem* and /dev/cu.usbserial* to the device globs on darwin/macos to allow auto-detect like a champ.

🍎


peterbarker (approved):
LGTM

PR #1219, opened 2026-06-22
:apple: tests: release UDP port between MAVFTP tests

drnic:
`TestMAVFTPPayloadDecoding.setUp()` opens a `udp:localhost:14550` connection for every test, but `tearDown()` never closed it. Re-binding the same port in the next test's `setUp()` succeeds on Linux (the socket sets `SO_REUSEADDR`) but fails on macOS with `OSError: [Errno 48] Address already in use`, since macOS additionally requires `SO_REUSEPORT` to rebind.

This makes the test suite fail for anyone running it on macOS, while passing in CI (Linux).


peterbarker (commented):
LGTM

PR #1221, opened 2026-06-22
README: document running the tests locally

drnic:
## Summary

The README explains how to install pymavlink but not how to run the test suite locally. Running it from a bare checkout has two non-obvious requirements, so this documents them.

If there is a better/preferred way to setup for local development, let me know and I'll update PR. This is how I got it working.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,8 @@ Starting from September 2022, mavnative, a C extension for parsing mavlink, was 
 From the pymavlink directory, you can use :
 
 ```bash
-MDEF=PATH_TO_message_definitions python3 -m pip install . -v
+(cd ..; git clone https://github.com/ArduPilot/mavlink.git)
+MDEF=$PWD/../mavlink/message_definitions python3 -m pip install . -v
 ```
 
 Since pip installation is executed from /tmp, it is necessary to point to the directory containing message definitions with MDEF. MDEF should not be set to any particular message version directory but the parent folder instead. If you have cloned from mavlink/mavlink then this is ```/mavlink/message_definitions``` . Using pip should auto install dependencies and allow you to keep them up-to-date. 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -85,6 +85,21 @@ Or:
...
+From inside the pymavlink directory, fetch the message definitions and run the
+suite with `pytest`:
+
+```bash
+git clone https://github.com/ArduPilot/mavlink.git
+ln -sf "$PWD/mavlink/message_definitions" ../message_definitions
+PYTHONPATH=.. python3 -m pytest
+```
+
+The tests need the message definitions at `../message_definitions` to generate
```
peterbarker: That's not actually quite true.

You should be able to specify `MDEF=$PWD/mavlink/message_definitions` in the environment to get the same effect.

drnic: Awesome, thanks. Commit updated.

README.md
```diff
@@ -105,6 +105,22 @@ Or:
...
 ```
 
+## Running the tests
+
+From inside the pymavlink directory, fetch the message definitions and run the
+suite with `pytest`:
+
+```bash
+git clone https://github.com/ArduPilot/mavlink.git
+MDEF=./mavlink/message_definitions PYTHONPATH=.. python3 -m pytest
```
peterbarker: Usually you wouldn't have this in the same directory as pymavlink - you'd have it one-level-up.

I just always use the definitions from an ArduPilot checkout, using that MDEF= method.

This is the reason you're getting a warning and trying to suppress "mavlink" as a directory in this directory from git's notice.

```suggestion
pushd ..
git clone https://github.com/ArduPilot/mavlink.git
popd
MDEF=$PWD/../mavlink/message_definitions PYTHONPATH=.. python3 -m pytest
```

peterbarker (approved):
LGTM

README.md
drnic: Ahh, cool. I've updated the PR; gone with terser `(cd ..; git clone ...)` since the magic of the brackets isolates the `cd`. I can revert back to pushd/popd if its your preference.

I also updated the development instructions to use the same change to make the command runnable.

PR #1222, opened 2026-06-22
Ignore test output artifacts and local mavlink dependency clone

drnic:
Test runs leave behind tmp.dump, prms.txt, test_wp-wp*.txt and a .tmp/ directory in the working tree (from test_mavlogdump, test_mavparm, test_wp and test_mavgen_typescript). The mavlink/ directory is a local clone of ArduPilot/mavlink used as the message_definitions source. None of these belong in version control.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,10 @@ include/mavlink/v1.0
 dfindexer/dfindexer_cy.c
 dfindexer/*.so
 dfindexer/*.pyd
+
+# test output artifacts
+.tmp/
+tmp.dump
+prms.txt
+test_wp-wp.txt
+test_wp-wp2.txt


.gitignore
```diff
@@ -28,3 +28,13 @@ include/mavlink/v1.0
 dfindexer/dfindexer_cy.c
 dfindexer/*.so
 dfindexer/*.pyd
+
+# locally-checked-out mavlink dependency (message_definitions source)
+mavlink/
+
```
peterbarker: This doesn't need to be checked out here....

Which workflows suggest this is a good idea - I'm wondering if something needs adjusting.

.gitignore
```diff
@@ -28,3 +28,13 @@ include/mavlink/v1.0
...
+# test output artifacts
+.tmp/
```
peterbarker, line 33: That actually looks pretty bad as something to create, let alone ignore.

.gitignore
drnic: That was from my other PR where I thought I needed a symlink to the folder; but now I know to use `MDEF=./mavlink/message_definitions PYTHONPATH=.. python3 -m pytest` instead.

.gitignore
drnic: Comes from https://github.com/ArduPilot/pymavlink/blob/master/tests/test_mavgen_typescript.py#L19

peterbarker (approved):

PR #1223, opened 2026-06-25
MAV_CMD missionOnly attribute - should default to false

hamishwillee:
When looking at https://github.com/mavlink/mavlink/pull/1871 I realised that this means all MAV_CMD are only intended to be used for missions. That's not correct - we want to apply this to a subset.

Note, this hasn't been applied anywhere yet, but I am about to do so. It might reasonably be changed to commandsOnly since that at least implies command protocol. With mission only, you might think you couldn't apply this to a fence command.


peterbarker (approved):
LGTM

hamishwillee:
Thanks!

PR #1224, opened 2026-06-25
XSD: add scaling value 2E-3

hamishwillee:
Allow efficient scaling of a % value in UINT16 value. With this, 0 to 100 is represented by 50000.

This is needed for https://github.com/mavlink/mavlink/pull/2526/changes#r3465455513 - but would do no harm anyway.

### DIFF
--- a/generator/mavschema.xsd
+++ b/generator/mavschema.xsd
@@ -182,6 +182,7 @@
 <xs:simpleType name="factor">
   <xs:restriction base="xs:string">
     <xs:enumeration value="1E-2"/>        <!-- actual value = stated value / 100 -->
+    <xs:enumeration value="2E-3"/>        <!-- actual value = stated value / 500 -->
     <xs:enumeration value="360/255"/>     <!-- actual value = stated value * 360/255, as used for GPS_STATUS.satellite_azimuth -->
   </xs:restriction>
 </xs:simpleType>


peterbarker:
This works, but why not the full range?  Leaving room for flag "not supplied values, of course.

I also suggest making the soc 0-1 rather than as a percentage to make things simpler.  Usually that will be scaled to a percentage, but it may not need to go via that before the GCS renders a little "gauge" icon, for example.

We're trying to move away from percentages generally in ArduPilot, preferring a scaler of 0-1 (or -1 to 1 in places...)

hamishwillee:
@peterbarker There's literally no difference in the data conveyed by a scaled percentage or a normalized value from 0 to 1, so if you want to call SoC a normalized scaled value then go for it :-).

I'm opposed to scaling across the full range though because it makes the maths a little bit harder, and a lot less intuitive. You could argue that doesn't matter, but I think it increases the chance for error.

If the argument is that the extra range gained from scaling from 50K up to 64K ish will make a difference, then you're arguing that it makes sense to move back to a float :-)

EDIT PS This is just my opinion. You're welcome to take this to the dev call since we're not getting this resolved right now unless you agree with me. I would like to come out of the call on Wednesday with agreement that this is the message. 3 years is a long wait.

peterbarker:
On Thu, 25 Jun 2026, Hamish Willee wrote:

> If the argument is that the extra range will make a difference, then you're arguing that it makes sense to move back to a float :-)

Yeah, I don't think *that* logics ;-)

hamishwillee:
> Yeah, I don't think *that* logics ;-)

I'm sure you get my point 
- if we don't need the extra resolution, choosing to provide it should be balanced against the calculation cost, ease of scaling, and so on. 
- if we really do need more resolution it is unlikely that increasing the range by 15% oir whatever it is will be sufficient long time.

I know I'd prefer to work with numbers that are easy to convert to decimal unless there is real value in it.

hamishwillee:
@peterbarker Thoughts? Because we also need this for EFI performance ^^^.

generator/mavschema.xsd
```diff
@@ -182,6 +182,7 @@
 <xs:simpleType name="factor">
   <xs:restriction base="xs:string">
     <xs:enumeration value="1E-2"/>        <!-- actual value = stated value / 100 -->
+    <xs:enumeration value="2E-3"/>        <!-- actual value = stated value / 500 -->
```
hamishwillee, line 185: The alternative?
```suggestion
    <xs:enumeration value="Uint16toPercent"/>        <!-- actual value = stated value / 65534 (% scaling into uint16 with max value indicating not supported -->
```

hamishwillee:
@peterbarker Can we resolve this?

The discussion is "what is the best way to scale 0-100% in a uint16. The potential range is 0 to 65534, leaving UINT_MAX for the unused value. We can also scale by 10 (0-1000), 100 (0-10000), or 2E3 (0 - 50000).

The argument for scaling the full range is that it gives you maximum resolution - there'd be no later "gee, I wish I'd used more of that range". The argument for the other scaling is that it is mentally easier to convert to percentage. If you don't see any advantage to the mental scaling, then by all means accept https://github.com/ArduPilot/pymavlink/pull/1224#discussion_r3584568560 and merge :-) 

i.e. I defer to your knowledge as an implementer, and as a user.

tridge:
I'd prefer to add float16 to MAVLink

hamishwillee:
> I'd prefer to add float16 to MAVLink

@tridge For sure. But that sounds like something that would take time to do and adopt and test. I have this problem now.

hamishwillee:
@tridge Even if you had an implementation today on mavgen, it would take some time for catchup from all the other generators.
Agree it would be great to have, and if you have an implemetnation we'll take it, but should not block this. 

PS [#1230 Support for float16](https://github.com/ArduPilot/pymavlink/pull/1230#top) for getting some discussion on the topic. But we can't wait on this. I have 5 year PRs that haven't been merged because it was too hard.

tridge (approved):

hamishwillee:
Thanks. @peterbarker Can you merge this?

hamishwillee:
Thanks very much.

PR #1227, opened 2026-07-14
C: remove broken and unused put_bitfield_n_by_index helper

julianoes:
Disclaimer: found by Claude.

`put_bitfield_n_by_index()` seems to be dead code, and not doing anything if used anywhere.

```
buffer[i_byte_index] |= ((0x00 << curr_bits_n) & v);
```


julianoes:
Good to go in?

peterbarker (approved):
Did it ever work? No — it was born broken. The earliest commit containing put_bitfield_n_by_index in the history this repo carries (43ded8f9d, "Upgraded to python generator", Aug 24 2011,
  from upstream MAVLink) already has the fatal no-op buffer[i_byte_index] |= ((0x00 << curr_bits_n) & v) — 0x00 << n is always zero, so the value is never written. The line was never touched
  in any of the ~15 years since; the v1.0 and v2.0 copies inherited it verbatim in 2012 and 2015. Codex also found it has additional original defects (an infinite loop when bit_index == 8,
  a wrong clear-mask, v never shifted between iterations), and nothing in-repo has ever called it. So the PR's premise is correct: this is dead code that never functioned, and removal loses
  no working behavior.

  The review's findings are about the cleanup being incomplete rather than wrong: the identical broken helper still ships in the include_v0.9 headers, its companion union __mavlink_bitfield
  is left unused in all three header sets, and deleting a publicly declared symbol is technically a source-compat break worth a changelog note (both Codex and the workflow flagged that one).

peterbarker:
Thanks!

PR #1229, opened 2026-07-15
implement 32 bit system IDs

tridge:
This is a WIP effort to implement 32 bit system IDs in MAVLink2. It is based on https://github.com/mavlink/rfcs/pull/20 with a change to use 32 bit instead of 16 bit wide system ID.
Only C and python generation so far, plus addition of CS and JavaScript dropping of unknown incompatible flags
I am also planning on resolving the 32 bit uint32_t issue with ArduPilot parameters as part of this effort which will make using IPv4 addresses more practical


generator/C/include_v2.0/mavlink_helpers.h
```diff
@@ -120,6 +120,69 @@ MAVLINK_HELPER uint8_t _mav_trim_payload(const char *payload, uint8_t length)
...
+	buf[ofs++] = msgid & 0xFF;
+	buf[ofs++] = (msgid >> 8) & 0xFF;
+	buf[ofs++] = (msgid >> 16) & 0xFF;
+	if (incompat_flags & MAVLINK_IFLAG_TARGETTED) {
+		buf[ofs++] = target_sysid & 0xFF;
+		buf[ofs++] = (target_sysid >> 8) & 0xFF;
+		buf[ofs++] = (target_sysid >> 16) & 0xFF;
+		buf[ofs++] = (target_sysid >> 24) & 0xFF;
+		buf[ofs++] = target_compid;
+	}
```
nexton-winjeel, line 167: Does `MAVLINK_IFLAG_TARGETTED` require `MAVLINK_IFLAG_SYSID32`?

```suggestion
	if (incompat_flags & MAVLINK_IFLAG_TARGETTED) {
		buf[ofs++] = target_sysid & 0xFF;
		if (incompat_flags & MAVLINK_IFLAG_SYSID32)
		    buf[ofs++] = (target_sysid >> 8) & 0xFF;
		    buf[ofs++] = (target_sysid >> 16) & 0xFF;
		    buf[ofs++] = (target_sysid >> 24) & 0xFF;
		}
		buf[ofs++] = target_compid;
	}
```

generator/C/include_v2.0/mavlink_helpers.h
```diff
@@ -234,44 +306,46 @@ MAVLINK_HELPER uint16_t mavlink_finalize_message_buffer(mavlink_message_t* msg,
...
-	msg->compat_flags = 0;
+	if (!mavlink1) {
+		if (system_id > 255) {
+			msg->incompat_flags |= MAVLINK_IFLAG_SYSID32;
+		}
+		if (target_sysid > 255) {
+			msg->incompat_flags |= MAVLINK_IFLAG_TARGETTED;
+			msg->target_sysid = target_sysid;
+			msg->target_compid = target_compid;
+		}
```
nexton-winjeel, line 339:
```suggestion
		if (target_sysid > 0) {
			msg->incompat_flags |= MAVLINK_IFLAG_TARGETTED;
			msg->target_sysid = target_sysid;
			msg->target_compid = target_compid;
		    if (target_sysid > 255) {
			    msg->incompat_flags |= MAVLINK_IFLAG_SYSID32;
		    }
		}
```

generator/C/include_v2.0/mavlink_helpers.h
```diff
@@ -120,6 +120,69 @@ MAVLINK_HELPER uint8_t _mav_trim_payload(const char *payload, uint8_t length)
...
+ */
+MAVLINK_HELPER uint8_t _mav_header_pack(uint8_t *buf, uint8_t magic, uint8_t len,
+					uint8_t incompat_flags, uint8_t compat_flags,
+					uint8_t seq, uint32_t sysid, uint8_t compid, uint32_t msgid,
+					uint32_t target_sysid, uint8_t target_compid)
+{
+	buf[0] = magic;
+	buf[1] = len;
+	if (magic == MAVLINK_STX_MAVLINK1) {
+		if (sysid > 255 || target_sysid > 255) {
```
nexton-winjeel, line 137: Elsewhere, equivalent checks also check if `msgid > 255`.

generator/C/include_v2.0/mavlink_types.h
```diff
@@ -305,8 +323,15 @@ typedef struct __mavlink_msg_entry {
...
-#define MAVLINK_IFLAG_MASK    0x01 // mask of all understood bits
+#define MAVLINK_IFLAG_SIGNED    0x01 // this message has a MAVLink2 signature attached
+#define MAVLINK_IFLAG_SYSID32   0x02 // this message uses a 32 bit system ID
+#define MAVLINK_IFLAG_TARGETTED 0x04 // this message has target sysid/compid in an extended header
+#define MAVLINK_IFLAG_MASK      0x07 // mask of all understood bits
+
+/*
+  compat_flags bits
+ */
+#define MAVLINK_CFLAG_SYSID32 0x01 // system is capable of understanding 32 bit system ID
```
nexton-winjeel, line 334: Would this be better as a new flag in [`MAV_PROTOCOL_CAPABILITY_MAVLINK2`](https://mavlink.io/en/messages/common.html#MAV_PROTOCOL_CAPABILITY_MAVLINK2)?

generator/C/include_v2.0/mavlink_helpers.h
julianoes: Agreed, it doesn't make sense to send a corrupt message over Mavlink1.

generator/C/include_v2.0/mavlink_helpers.h
```diff
@@ -120,6 +120,69 @@ MAVLINK_HELPER uint8_t _mav_trim_payload(const char *payload, uint8_t length)
...
+			// 32 bit IDs cannot be represented in MAVLink1; never truncate
+			return 0;
+		}
+		buf[2] = seq;
+		buf[3] = sysid & 0xFF;
+		buf[4] = compid;
+		buf[5] = msgid & 0xFF;
+		return MAVLINK_CORE_HEADER_MAVLINK1_LEN+1;
+	}
+	uint8_t ofs = 2;
```
julianoes, line 147: You could do `uint8_t ofs = 0` at the very top and use it throughout.

generator/C/include_v2.0/mavlink_helpers.h
julianoes: I'm not sure. It almost looks like the two flags are independent which has the benefit of not always burning the 3 bytes in the header.

generator/C/include_v2.0/mavlink_types.h
julianoes: MAV_PROTOCOL_CAPABILITY flags are a level higher. They document what Mavlink services/features are supported but they don't describe the protocol bytes changed here. And MAV_PROTOCOL_CAPABILITY flags are only received, once mavlink implementations are actually working/talking.

generator/C/include_v2.0/mavlink_types.h
```diff
@@ -305,8 +323,15 @@ typedef struct __mavlink_msg_entry {
 /*
   incompat_flags bits
  */
-#define MAVLINK_IFLAG_SIGNED  0x01
-#define MAVLINK_IFLAG_MASK    0x01 // mask of all understood bits
+#define MAVLINK_IFLAG_SIGNED    0x01 // this message has a MAVLink2 signature attached
+#define MAVLINK_IFLAG_SYSID32   0x02 // this message uses a 32 bit system ID
+#define MAVLINK_IFLAG_TARGETTED 0x04 // this message has target sysid/compid in an extended header
```
julianoes, line 328: I'd make it clear that this is related to sysid 32. Just "targetted" could also mean compid, etc.

```suggestion
#define MAVLINK_IFLAG_TARGETTED_SYSID32 0x04 // this message has target sysid/compid in an extended header
```

generator/C/include_v2.0/mavlink_helpers.h
julianoes: I think the confusion is in the name `MAVLINK_IFLAG_TARGETTED`. It should be called `MAVLINK_IFLAG_TARGETTED_SYSID32`. And then the original implementation is correct and your suggestion is wrong. A `target_sysid > 0` is perfectly valid with the existing mavlink implementation and doesn't need a incompat flag.

generator/C/include_v2.0/mavlink_types.h
```diff
@@ -200,7 +209,16 @@ typedef enum {
     MAVLINK_PARSE_STATE_GOT_CRC1,
     MAVLINK_PARSE_STATE_GOT_BAD_CRC1,
     MAVLINK_PARSE_STATE_SIGNATURE_WAIT,
-    MAVLINK_PARSE_STATE_SIGNATURE_WAIT_BAD_CRC
+    MAVLINK_PARSE_STATE_SIGNATURE_WAIT_BAD_CRC,
+    // new states appended to keep existing state numbering stable
```
julianoes, line 213: Does that have an actual benefit? I suppose if implementations check against magic numbers?

julianoes (commented):
1. Should sysid32 be behind a #define, so that we can keep updating implementations that don't support sysid32, or should this be rolled out when updating universally and potentially break builds? Nevermind, seeing all the functions such as `_pack(...)` I suppose it would add a lot of duplication and be very messy.
2. Is there any benefit of allowing 32 bit compids as well? I think I like the idea of the symmetry but I'm not sure it would be the right thing to do.
3. Can we do a clean format PR before this. I see spaces and tabs added in this PR and I just find that a bit yucky.

I think I'm actually somewhat convinced that this could work.

julianoes:
I've spent some tokens on an Opus 5 review. See details below.

@tridge let me know if you want to tackle these or whether you want to defer them to future PRs.

Edit: and I'm not trying to slow you down or block you, just want to make sure we don't end up breaking too much. I'm eager to add this to PX4 and MAVSDK.

<details>

# Review: `pr-sysid-32` vs `master`

Branch: `pr-sysid-32` (10 commits, ~1400 lines across 15 files)
Repo: `pymavlink`
Scope: MAVLink 2.1 32-bit system IDs (`MAVLINK_IFLAG_SYSID32`) and extended target headers (`MAVLINK_IFLAG_TARGETTED`)

Method: read of the full diff plus surrounding code. No builds or tests were run.

## Commits under review

```
d8cd90c3 generator: C: clear target IDs when parsing a MAVLink1 frame
0ede65f6 generator: mavgen_javascript: fix link_id extraction from signed messages
3b3381b8 generator: mavgen_javascript: 32 bit sysid and extended target support
43c6e224 generator: mavgen_lua: handle the 32 bit sysid mavlink_message_t layout
bff89dc8 generator: reject unknown incompat_flags in CS and JavaScript parsers
b6aa784c tests: add 32 bit sysid tests
53d39dd7 generator: mavgen_python: 32 bit sysid and extended target support
32c6d19f generator: C: test 32 bit sysid and extended targets
c88459af generator: mavgen_c: generate 32 bit sysid/target aware APIs
7c34da73 generator: C: add 32 bit system ID and extended target support
```

## Overall

The core wire work is solid and internally consistent:

- `_mav_header_pack()` is a good centralisation of what was previously four copies of hand-written header serialisation.
- The CRC and signing offset arithmetic is correct everywhere I checked — `header_len` now consistently includes the magic byte, and `header_len-1` is passed into `crc_calculate`.
- The parser state additions are non-overlapping and appended, so existing state numbering stays stable.
- The stale-target clearing in both `GOT_LENGTH` and the MAVLink1 `GOT_STX` path is the right fix.
- The C to Python golden-frame cross-check in `tests/test_sysid32.py` is exactly the right way to pin this down.
- Lua struct offsets (295/299) check out against a `MAVPACKED` `mavlink_message_t`.

Findings follow in rough severity order.

---

## 1. C `_decode()` turns a 32-bit target into a broadcast, and diverges from Python/JS

`generator/mavgen_c.py:759`

```python
f.decode_call = '_MAV_RETURN_uint8_t(msg, %u)' % f.wire_offset
```

For a TARGETTED frame, `mavlink_msg_command_long_decode()` yields `pkt.target_system == 0`. In MAVLink, 0 means *broadcast*. So the near-universal receiver idiom

```c
if (packet.target_system == 0 || packet.target_system == mavlink_system.sysid) {
    handle();
}
```

accepts a message addressed to some *other* 32-bit system. The testsuite even locks this in:

```c
assert(pkt.target_system == (targetted?0:target));
```

Python (`mavgen_python.py`) and JS both do the opposite — they overlay the full 32-bit value onto the decoded field, so `msg.target_system` is correct there. Same feature, three languages, two different semantics for the same frame.

The constraint is understood: `mavlink_X_t.target_system` is `uint8_t` and the `#else` branch memcpy's the struct straight off the wire, so it cannot be widened. But 0 is the worst possible sentinel. Options worth weighing:

- Use `0xFF` (or another reserved value) so the frame fails an equality check rather than passing a broadcast check.
- Emit a companion `target_system32` field in the struct.
- At minimum, make this a documented, loud caveat rather than a comment in the generator.

## 2. `MAVLINK_CFLAG_SYSID32` is set unconditionally on every MAVLink2 frame

`generator/C/include_v2.0/mavlink_helpers.h:341`, `generator/mavgen_python.py:363`, `generator/mavgen_javascript.py:378`

Every outgoing MAVLink2 packet from every pymavlink user now has `compat_flags == 0x01` instead of `0x00`. That changes the bytes and CRC of literally every frame — the JS reference-buffer churn in this diff is a preview of what downstream golden tests, tlog comparisons and interop fixtures will hit.

Two concerns beyond the churn:

- It is an unconditional capability claim for a bit assignment that, as far as I can tell from the branch, is not finalised upstream. If the bit moves, every shipped build is lying.
- Combined with finding 1, a plain recompile makes a C application advertise "I understand 32-bit system IDs" while its `_decode()`-based routing demonstrably does not.

Suggest making this opt-in — a `MAVLINK_STATUS_FLAG_OUT_SYSID32` or a compile-time define — rather than always-on.

## 3. C# `MAVLinkMessage.processBuffer()` is unguarded

`generator/CS/MAVLinkMessage.cs:157`

The commit "reject unknown incompat_flags in CS and JavaScript parsers" only covers `MavlinkParse.ReadPacket(Stream)`. `processBuffer` reads `srcSystem` from `buffer[5]`, `msgid` from `buffer[7..9]`, and CRC from `MAVLINK_CORE_HEADER_LEN + payloadlength + 1/+2` — all fixed offsets.

It is reachable from the public `new MAVLinkMessage(byte[])`, which is how packet-oriented transports (UDP datagrams, log replay) feed it. An extended-header frame there yields a garbage msgid and a misplaced CRC read rather than a clean rejection. `generator/CS/MAVLinkMessage.cs:76` (`ByteArrayToStructure(..., MAVLINK_NUM_HEADER_BYTES, payloadlength)`) has the same fixed-offset assumption.

The discard-length arithmetic in `MavlinkParse.cs` itself is correct — worked through: `lengthtoread + extlen - (headerlengthstx - 2)` equals `payload + extlen + 2 + sig`.

## 4. Lua layout detection breaks if `MAVLINK_MAX_PAYLOAD_LEN` is reduced

`generator/mavgen_lua.py:89`

```lua
return #message >= 299
```

The old Lua code only used offsets counted from the *start* of the struct (at most `12 + payload_length`), so it was immune to `MAVLINK_MAX_PAYLOAD_LEN`. The new code reads `target_sysid`/`target_compid` at absolute offsets 295-299, and detects the layout by total size — both of which shift if a build lowers `MAVLINK_MAX_PAYLOAD_LEN` (a supported `#ifndef` knob).

On such a build, `sysid32_layout()` returns false and the script silently decodes with the *old* offsets: wrong sysid, wrong msgid, no error. Worth at least a comment stating the assumption, or deriving the offset from `#message` rather than hardcoding 295.

## 5. Generator coverage gaps

Java (`generator/mavgen_java.py:420`), WLua (`generator/mavgen_wlua.py:676` — `if version == 0xfd and incompatibility_flag == 0x01`), Swift, spin2, C++11 and TypeScript are untouched.

Java and the Wireshark dissector will *misparse* extended-header frames at fixed offsets rather than reject them. Not necessarily blocking, but the Wireshark dissector in particular is the tool people reach for when debugging exactly this feature. Worth an explicit "out of scope, follow-up" note in the PR.

---

## Smaller items

- **Orphaned doxygen** — `generator/C/include_v2.0/mavlink_helpers.h:290`: moving `_mav_parse_error()` up left the `@brief Finalize a MAVLink message with channel assignment / @param system_id / @param length` block attached to `_mav_parse_error`.
- **`pack_status` non-CRC_EXTRA branch** — `generator/mavgen_c.py:277`: the `#else` arm still calls plain `mavlink_finalize_message_buffer` without `${target_suffix}`, so with `MAVLINK_CRC_EXTRA=0` a >255 target is silently dropped (payload byte already zeroed, no extended header). That branch was already arity-broken on master, but it is now also silently lossy.
- **Leaked loop variable** — `generator/mavgen_c.py:693`: `sysid32_capable` is assigned inside the first `for m in xml.message` loop and read in the second. Works by scope leakage; should be hoisted above both.
- **Stale comment** — `generator/mavgen_javascript.py:750`: "the frame is rejected in decode()" is no longer true now that `MAVLINK_IFLAG_MASK = 0x07`. It *is* accurate in `mavgen_javascript_stable.py`, where it was presumably copied from.
- **MAVLink1 JS dialects** get `MAVLINK_IFLAG_MASK = 0x07`, claiming to understand flags the v1 decoder cannot see. Harmless (v1 forces `incompat_flags = 0`) but misleading.
- **Inconsistent state naming** — `generator/C/include_v2.0/mavlink_types.h`: `GOT_SYSID1..3` mean "1..3 bytes consumed", but `GOT_TARGET_SYSID0..3` mean "1..4 bytes consumed". Same concept, off by one apart.
- **Test readability** — `tests/test_sysid32.py:95`: `assert payload[30] == 0 if len(payload) > 30 else True` parses as intended but reads as a precedence bug. `if len(payload) > 30: assert payload[30] == 0` is clearer.
- **ABI note for the PR description**: `mavlink_message_t` grows 291 to 299 bytes, `mavlink_system_t.sysid` and `__mavlink_signing_stream.sysid` widen to `uint32_t`, `MAVLINK_MAX_PACKET_LEN` grows by 8, and every `_pack`/`_encode`/`get_target_system` signature changes type. Anything sharing these headers needs a lockstep rebuild — and any downstream buffer sized by a hardcoded constant instead of `MAVLINK_MAX_PACKET_LEN` will now overflow.

## Incidental fix worth calling out

`generator/mavgen_python.py:1197` changes `m._payload = msgbuf[6:...]` to `msgbuf[headerlen:...]`. On master that was wrong for MAVLink2 (4 header bytes leaked into `_payload`). Genuine fix, but it is unrelated to sysid32 and buried in a large commit — it deserves its own commit and a changelog line, since anything reading `msg._payload` on v2 gets different bytes now.

## Checked and found correct

- CRC offsets in `mavlink_finalize_message_buffer_target`, `_mav_finalize_message_chan_send_target`, `mavlink_msg_to_send_buffer` and `_mavlink_resend_uart`.
- `MAVLINK_START/END_UART_SEND` lengths.
- `mavlink_signature_check` re-serialising the header, and never hitting the `header_len == 0` path since MAVLink1 frames are never signed.
- MAVLink1 guards refusing to truncate rather than aliasing a sysid.
- Python `mav20_sysid32_unpacker` layout (13 bytes, `<` so unaligned).
- JS `header_len` propagation into the length validation and `payloadBuf` slice.
- JS `_link_id` fix (`msgbuf[-13]` to `msgbuf.slice(-13)[0]`) — that was a real pre-existing bug.
- C# discard arithmetic.
- Lua CRC reconstruction byte order.
- `dialects/*.py` are gitignored, so no checked-in generated output goes stale.



</details>

julianoes (approved):

generator/C/include_v2.0/mavlink_helpers.h
tridge: I've changed it so that 8 bit targetting is supported

generator/C/include_v2.0/mavlink_helpers.h
tridge: I've left it as MAVLINK_IFLAG_TARGETTED but made it independent of MAVLINK_IFLAG_TARGETTED_SYSID32

generator/C/include_v2.0/mavlink_types.h
tridge: yep

tridge:
@julianoes @nexton-winjeel @peterbarker @hamishwillee I've updated this ready for next round of review

generator/C/include_v2.0/mavlink_types.h
tridge: yep

generator/C/include_v2.0/mavlink_types.h
tridge: I made targeting separate from sysid32 so left as MAVLINK_IFLAG_TARGETTED, but could be MAVLINK_IFLAG_TARGETTED_SYSID

generator/C/include_v2.0/mavlink_helpers.h
tridge: could do, not really less code though

AP-Review:

Full report: [https://uav.tridgell.net/DevCallReviews/2026_09_24_AIReview/devcall_pr_reviews.html#prpymavlink-1229](https://uav.tridgell.net/DevCallReviews/2026_09_24_AIReview/devcall_pr_reviews.html#prpymavlink-1229)

Reviewed at head `1e72acc133`. **COMMENT** — careful, well-instrumented work. The framing arithmetic is consistent across all fourteen generators, the flags really are independent, the buffers really did grow, the CRC and signature ranges are right, and the test suite demonstrably catches errors in it. Two things I would want settled before the RFC freezes.

**The four-combination framing table, computed from the code and confirmed against real generated frames:**

| incompat_flags | hdr len | field offsets (0-based) | payload@ |
|---|---|---|---|
| `0x00` neither | **10** | sysid5 (1B), compid6, msgid7-9 | 10 |
| `0x02` SYSID32 | **13** | sysid5-8 (4B LE), compid9, msgid10-12 | 13 |
| `0x04` TARGETTED | **12** | sysid5, compid6, msgid7-9, tgt_sys10 (1B), tgt_comp11 | 12 |
| `0x06` both | **18** | sysid5-8, compid9, msgid10-12, tgt_sys13-16 (4B LE), tgt_comp17 | 18 |

`0x01` SIGNED is orthogonal (+13 trailing). CRC covers `buf[1 .. header_len+len-1]` + crc_extra in every combination; the signature hash covers `header[0..header_len-1]` + payload + ck + sig[0..6], same as master. I hand-checked all seven independent implementations of the extended-header length formula and they agree (+3 for SYSID32, +2 for TARGETTED, +3 more when both), including the equivalent `len + 9` remaining-bytes form used by Java/Swift/Spin2/Ada and the C# and TypeScript whole-frame variants. **No generator got the byte order, the flag test or a header length wrong.** Constants measured from a compiled probe: `MAVLINK_MAX_HEADER_LEN=18`, `MAVLINK_MAX_EXT_HEADER_BYTES=8`, `MAVLINK_MAX_PACKET_LEN=288`, `sizeof(mavlink_message_t)=299` (was 291) — no undersized buffer anywhere.

### Two items for the RFC

**1. The legacy C `_decode()` sentinel collides with 255** — `generator/mavgen_c.py:690`. The generated `decode_post` writes `target_system = msg->target_sysid > UINT8_MAX ? UINT8_MAX : (uint8_t)msg->target_sysid`, which correctly fixes the previous review's "0 means broadcast" complaint — but 255 is not reserved, it is the default sysid of MAVProxy, QGC and Mission Planner. Compiled probe against your own generated headers: target 70000 decodes to 255, and the near-universal receiver idiom `if (pkt.target_system == 0 || pkt.target_system == mavlink_system.sysid)` on a node whose own sysid is 255 **accepts** the frame, while the same idiom on sysid 1 correctly does not.

There is no safe 8-bit sentinel, and the documented contract is that receivers use `mavlink_msg_get_target_sysid()` when TARGETTED is set (which does return 70000) — so this is a spec decision rather than a code defect. But the failure mode is a silent misroute in code that predates the contract, so it is worth settling deliberately, perhaps with a loud caveat in the generated header.

**2. `MAVLink.decode()` does not enforce `MAVLINK_IFLAG_MASK`** — `generator/mavgen_python.py:1169`. Handing a frame with incompat flag `0x08` and a recomputed CRC to the two entry points gives `parse_char()` → `MAVLink_bad_data` (correct) but `decode()` → **accepted** as a HEARTBEAT with `incompat_flags=0x08`. The mask check lives only at :1056 and :1063. The C parser checks in `GOT_LENGTH` and both JS decoders check inside `decode()`, so this is a cross-language divergence. It is pre-existing on master — I reproduced it on both parent and head — but "an unknown incompat flag MUST drop the frame" is the single MUST this whole RFC rests on, and `decode()` is public and used directly by third-party log tooling. Moving the check into `_decode_header` is a two-line fix.

### Notes

- `mavlink_helpers.h:345` — `MAVLINK_CFLAG_SYSID32` is set unconditionally, so every ordinary MAVLink 2 frame changes on the wire. I generated `ardupilotmega` at master and at your head and encoded the same messages with the same seq and sysid: byte 3 goes `0x00 → 0x01` and the CRC changes, on all four messages tested, with header length and payload otherwise identical. It is *compat*_flags so no conforming receiver breaks and the advertisement is deliberate — but it invalidates every golden fixture, tlog byte-comparison and interop test downstream, and your own `message20.js` churn is a preview. julianoes raised this in August and it is the one item from that review still open; it deserves an explicit line in the description or the RFC.
- `mavgen_python.py:1279` — `m._payload` changes for every MAVLink 2 message (`msgbuf[6:...]` → `msgbuf[headerlen:...]`). Measured on a v2 HEARTBEAT: master gives 13 bytes with four header bytes leaked in, head gives the correct 9. A genuine pre-existing bug fix and required for a variable header offset, but anything reading `msg._payload` on v2 gets different bytes now, and it is buried inside a 44-file squashed commit. Worth a changelog line.
- `mavlink_types.h:31` — `MAVLINK_NUM_NON_PAYLOAD_BYTES` is now a *minimum* (12, against up to 20) rather than an absolute, but unlike its siblings at :27-29 it carries no doc comment saying so. Nothing in-tree is wrong, but downstream code sizing a buffer from it directly will be short.
- `mavlink_helpers.h:436` — no explicit-target variant of the chan-send API, so in C explicit 8-bit and explicit-broadcast header targets are reachable only via the buffer API while Python and JS get `set_target()` everywhere. Documented in `tests/sysid32/README.md`, so an API-convenience gap rather than missing functionality.
- `mavlink_types.h:214` — `GOT_SYSID1..3` mean "1..3 bytes consumed" while `GOT_TARGET_SYSID0..3` mean "1..4". Cosmetic, raised last round, still open.
- `mavgen_javascript.py:1028` — pre-existing typo, present identically on master:892 (`msgID` vs `msgId`). Drive-by; mentioned only because the surrounding function was rewritten.
- Ten commits in the August round have become one commit spanning 14 generators, the C helpers, the wire types and two test suites. A per-subsystem split would also give the `_payload` fix and the compat-flag decision their own visible history.

### What I actually ran

`tests/test_sysid32.py` **128 passed**; `test_sysid32_generators.py` 8 passed / 9 skipped after installing lua5.4, tshark and the JS/TypeScript deps to un-skip C, NextGen JS, stable JS x2, C++11, Lua, wlua and TypeScript; regenerated the C headers and ran `testmav2.0_ardupilotmega`, `testmav1.0_ardupilotmega` and `test_issues`, all passing; the full existing suite at head **276 passed / 17 skipped**; the JS mocha suite **1657 passing**; and a hand-written round-trip of all four flag combinations plus explicit broadcast and payload-target>255.

**Two mutation tests confirmed the suites actually bite**, both reverted afterwards: making `_mav_header_pack` always emit a 4-byte target under TARGETTED produced `testmav.c:456: Assertion 'header_len == MAVLINK_NUM_HEADER_BYTES + MAVLINK_TARGETTED_HEADER_EXTRA' failed` — and note the binary had to be removed first, because the Makefile does not depend on the headers and the first attempt silently ran the stale binary and "passed". Shifting the Python 8-bit target read by one byte gave 44 failed / 84 passed.

Also verified: MAVLink 1 never truncates (it returns 0 and raises a parse error rather than aliasing another system or broadcasting); stale target IDs are cleared on both paths; signing byte ranges are unchanged from master; and four of julianoes's five findings are fixed.

**Coverage, honestly:** of the 2776 added lines I read roughly 1100 deeply — the entire C helpers/types/protocol change, `mavgen_c.py`, `mavgen_python.py`, `mavgen_javascript.py`, `mavgen_lua.py`, `mavgen_javascript_stable.py` and `mavparse.py` — plus ~200 lines of Java/Swift/Spin2/CS/Ada/TS/wlua/ObjC/C++11 reject logic read line by line for arithmetic divergence. I sampled `testmav.c`, the new Python test code (proving behaviour by mutation rather than reading every assertion) and the fixture sources in 12 languages. **The main gap:** eight generator paths have no toolchain on this box — Java, C#, Ada x2, Swift, ObjC x2, Spin2 x2 — and are read-verified only, not executed.

One finding was dropped after Codex refuted it: C#'s unwidened `MAVLINK_MAX_PACKET_LEN` is fine, because `MavlinkParse.cs:201` allocates a separate discard buffer including the extension bytes and rejects unsupported flags before interpreting later fields.

AP-Review:

Re-reviewed at head `a90e6135ba`; my earlier comment above is superseded.

Full report: https://uav.tridgell.net/DevCallReviews/2026_09_25_AIReview/devcall_pr_reviews.html#prpymavlink-1229

**One of the two RFC items is properly closed, the other is not — and it now has a second, worse mechanism alongside it. Verdict stays COMMENT.** One new plain bug, with a two-line fix.

The real delta since `1e72acc133` is **3 files, +70/−5** (both heads share merge-base `a3d28c31`): the mask check moved into `_decode_header`, a doc comment on `mavlink_types.h:31`, and two new tests. Everything else is the same content re-split from 1 squashed commit into **16 per-subsystem commits**.

## Resolved

- **`MAVLink.decode()` now enforces `MAVLINK_IFLAG_MASK`** — moved to `generator/mavgen_python.py:1128-1129`, inside `_decode_header`. Flags `0x08`/`0x10`/`0x80` with a recomputed CRC are now rejected by **both** entry points, closing the cross-language divergence. Mutation-proved twice independently: deleting the two lines gives **80 failed / 130 passed**. And no over-rejection — MAVLink 1 with `seq=0xFF` and MAVLink 2 with `compat_flags=0xFF` both still decode.
- **`MAVLINK_NUM_NON_PAYLOAD_BYTES` is documented** at `mavlink_types.h:31` as a minimum, pointing at `MAVLINK_MAX_PACKET_LEN` for buffer sizing.
- **The `_payload` offset fix has its own changelog line** — commit `ba0d55c1`.
- **Per-subsystem history** — the squashed-commit note is addressed.

## The one plain bug

**`setup.py:194` — the new Ada tests fail under the project's own CI invocation.**

This PR adds `'TypeScript/*.ts'` to `package_data`, but `Ada/` is still not shipped, while `mavgen_ada.py` copies from it. `.github/workflows/test.yml` runs a bare `python3 -m pytest` *after* `pip install .`, deliberately — the workflow's own comment says "we must do all testing on the installed python package, not on the build tree". Under that invocation the two new `test_ada_rejects_extensions` cases fail with `FileNotFoundError`: 1.0 on `Ada/mavlink_v1.gpr`, 2.0 on `Ada/mavlink.ads`.

Reproduced independently by both reviewers and proved by mutation — adding `'Ada/*.*'`, `'Ada/v2/*.*'`, `'Ada/examples/*.*'` and `'Ada/v2/examples/*.*'` to `package_data` makes both pass. The cross-check had **GNAT 14.3.0 installed and the tests genuinely executed**, so this is not a skip being misread as a pass. The packaging omission predates the PR; its new tests are what expose it. CI is green here (21/21) because the runner has no `gnatmake` so the test skips before reaching the copy — though neither of us could inspect the runner image to confirm that last step. Anyone with `gnat` installed following the CI recipe gets a hard failure.

## Two items for the RFC — spec decisions, not defects

**1. `generator/mavgen_c.py:690` — the legacy `_decode()` sentinel still collides with 255.** Unchanged since last round. Re-derived with fresh compiled probes: target 70000 decodes to 255, and the usual receiver idiom accepts on a node whose own sysid is 255 while correctly rejecting on sysid 1. 255 is the default sysid of MAVProxy, QGC and Mission Planner. No caveat has appeared in the generated header, `mavlink_types.h` or `tests/sysid32/README.md`.

**2. `generator/mavgen_c.py:743` — the widened getter truncates silently in unmodified legacy code.** *(New this round.)* The per-message target getter's return type changed `uint8_t` → `uint32_t`, so existing C storing it in a `uint8_t` — the *correct* type before this PR — truncates. Built with the project's own `-Wall -Werror` it produces **zero diagnostics**. Target 70000 becomes 112, so a node with sysid 112 accepts the frame.

The cross-check found the sharper case: **a target of 65536 — any multiple of 256 — truncates to 0, which every receiver using the standard idiom treats as broadcast.** That is worse than the sentinel above, which at least pins to one value. Same framing — a migration-contract question rather than evidence the full target is wrong to return — but it argues for the caveat living in the generated header next to the getter, not only in the RFC.

## Notes

- **`mavlink_helpers.h:345`** — `MAVLINK_CFLAG_SYSID32` is still unconditional, so byte 3 of every ordinary MAVLink 2 frame goes `0x00 → 0x01` and the CRC changes. It is *compat*_flags so nothing conforming breaks, but it invalidates every golden fixture, tlog byte-comparison and interop test downstream. The description covers flag independence and per-language scope but never states this. julianoes raised it in August and it is the last item from that review still open — worth an explicit line in the description or the RFC.
- **`tests/test_sysid32_generators.py:114-117`** — `node_environment()` probes only for `jspack`, but the stable JS generator needs `underscore`. Without it the two stable-JS cases fail hard with `MODULE_NOT_FOUND` instead of skipping, unlike every other optional runtime in that file. Seen on two independent machines.
- **`mavlink_types.h:214`, `:218`** — `GOT_SYSID1..3` mean "1..3 bytes consumed" while `GOT_TARGET_SYSID0..3` mean "1..4". Cosmetic, still open.
- Cosmetic side effect of the mask-check move: a buffer with a bad magic byte *and* high bits in byte 2 now raises `invalid incompat_flags` where master raised `invalid MAVLink prefix`. Unreachable via `parse_char`.

## Dropped since last round

- **The missing explicit-target chan-send variant** — `mavlink_finalize_message_buffer_targetted` exists at `mavlink_helpers.h:384` and the new `tests/sysid32/README.md` documents the path including MAVLink 1 returning zero. A documented convenience gap, not worth carrying.
- **The `mavgen_javascript.py:1028` `msgID`/`msgId` typo** — traced this round: it is the v1 fallback branch of a v2 dialect and the function throws on the magic byte before `msgId` is read. Dead code, identical on master.

## Re-derived independently rather than carried forward

Both reviewers rebuilt the four-combination framing table from scratch with fresh C probes — `0x00`→10, `0x02`→13, `0x04`→12, `0x06`→18 — matching last round exactly, no generator divergence. All four `GOLDEN` hex strings in `tests/test_sysid32.py` reproduced **byte-for-byte** from a C encoder built at this head, twice, so the C↔Python pinning is real and not self-referential. Constants re-measured: `MAVLINK_MAX_HEADER_LEN=18`, `MAVLINK_MAX_EXT_HEADER_BYTES=8`, `MAVLINK_MAX_PACKET_LEN=288`, `sizeof(mavlink_message_t)=299`. julianoes's Lua finding is properly fixed rather than papered over — `mavgen_lua.py:90-93` now errors instead of silently decoding with the old offsets.

## What I ran

`tests/test_sysid32.py` **210 passed** (128 last round). Full suite: installed-package mode 360 passed / 13 skipped / 2 failed (the Ada pair); `PYTHONPATH=..` mode 362 passed / 13 skipped / 0 failed. Regenerated all C v1.0/v2.0/CPP11 headers and ran `testmav1.0_ardupilotmega`, `testmav2.0_ardupilotmega`, `test_issues` — all clean. JS mocha **2775 passing**. Three mutations, all reverted: the mask check (80 failed), a forced 4-byte extended target (C assertion at `testmav.c:456`), and the `setup.py` Ada fix (2 failures → 0).

Two environment notes worth recording. The **stale-binary trap is real** — the C test Makefile has no header dependency, so the binary must be removed before rebuilding or the old one silently passes. And contrary to the usual pymavlink rule, here **pytest imported the installed package, not the checkout**, so mutations had to go into site-packages; mutating the checkout alone would have inverted every result.

**Not covered:** the ~2800 unchanged lines were not re-read in full — effort went into the 70-line delta line by line, re-deriving both headline findings from code, and re-running everything. Generator paths that could not be executed: C# (`mcs` absent), the TypeScript runtime, Objective-C, the GNUstep syntax check, and the Spin2 runtime harness. Newly executable this round — a real coverage gain from the PR's own test additions — are Java, Swift, Ada 1.0 and 2.0, Spin2 compile, WLua/tshark, C++11 and Lua, all passing.

---
*Reviewed by: Claude (primary) + Codex (independent cross-check). The cross-check confirmed five of six findings, corrected the Ada failure ordering, and added the 65536→0 broadcast case above, which is the sharpest form of that finding.*

PR #1237, opened 2026-07-28
mavutil: add type annotations to mavfile_state and param_state

khancyr:
This Only set annotation and don't touch anything else ! No code modification

Declares real attribute types on the per-sysid/param state helper classes so mavfile's proxying properties (target_system, messages, flightmode, etc.) can be typed accurately in a follow-up step instead of falling back to Any.


peterbarker (approved):
LGTM.

PR #1238, opened 2026-07-28
mavgen_python: accept str for char and char[] message fields

### LINKED ISSUE #1225: str.split() does not accept bytes
When running the mavproxy_example.py example module, I received this:

```
  File "ArduPilot/mavproxy/venv/bin/mavproxy.py", line 1102, in periodic_tasks
    m.idle_task()
  File "ArduPilot/mavproxy/venv/lib/python3.12/site-packages/MAVProxy/modules/mavproxy_example.py", line 81, in idle_task
    self.master.mav.statustext_send(mavutil.mavlink.MAV_SEVERITY_NOTICE,
  File "ArduPilot/mavproxy/venv/lib/python3.12/site-packages/pymavlink/dialects/v20/all.py", line 32199, in statustext_send
    self.send(self.statustext_encode(severity, text, id, chunk_seq), force_mavlink1=force_mavlink1)
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "ArduPilot/mavproxy/venv/lib/python3.12/site-packages/pymavlink/dialects/v20/all.py", line 32183, in statustext_encode
    return MAVLink_statustext_message(severity, text, id, chunk_seq)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "ArduPilot/mavproxy/venv/lib/python3.12/site-packages/pymavlink/dialects/v20/all.py", line 17970, in __init__
    self.text = text.split(str(b"\x00"), 1)[0].decode("ascii", errors="replace")
                ^^^^^^^^^^^^^^^^^^^^^^
TypeError: must be str or None, not bytes
```

This appears to fix the error:
```
$ diff -u all.py.orig all.py
--- all.py.orig 2026-07-02 10:30:02.297639272 -0400
+++ all.py      2026-07-02 10:30:51.411888176 -0400
@@ -17967,7 +17967,7 @@
         self._instance_offset = MAVLink_statustext_message.instance_offset
         self.severity = severity
         self._text_raw = text
-        self.text = text.split(b"\x00", 1)[0].decode("ascii", errors="replace")
+        self.text = text.split(str(b"\x00"), 1)[0].decode("ascii", errors="replace")
         self.id = id
         self.chunk_seq = chunk_seq
 
```

This may be the source of the error plus a 2nd case of the same problem:
```
$ grep -r split | grep 00
generator/mavgen_python.py:                init_fields.append('self.%s = %s.split(b"\\x00", 1)[0].decode("ascii", errors="replace")' % (f.name, f.name))
mavftp.py:            dentries = sorted(op.payload.split(b"\x00"))
```

MuhammadFathy:
Passing a str for a char/char[] field raised

    TypeError: must be str or None, not bytes

from the generated message constructor, e.g.

    mav.statustext_send(mavutil.mavlink.MAV_SEVERITY_NOTICE, "hello")

as MAVProxy's example module does. Commit 667cfcd4 started splitting the raw value on b"\x00" in __init__, which only works for bytes-like input, so every caller that had been passing a str broke.

Add a char_field_to_bytes() helper to the generated module and route char field assignment through it, so str, bytes and bytearray are all accepted and normalised to bytes before the NUL split and before being handed to struct.pack. str is encoded as ascii with errors="replace", matching the decode already performed on the same line.

The constructor and *_encode/*_send annotations for char fields are widened from bytes to the new CharFieldLike alias so the generated dialects still type-check.

The second case mentioned in the issue, mavftp.py's op.payload.split(b"\x00"), is not affected: op.payload is a bytearray and bytearray.split() accepts a bytes separator.

Adds tests/test_mavgen_python.py, which generates the common dialect and exercises str/bytes/bytearray input, NUL truncation, non-ascii handling and a full pack/parse round trip.

Fixes #1225

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,6 +25,7 @@ include/mavlink/v1.0
 /.vscode
 .idea/
 .venv/
+.tmp/
 dfindexer/dfindexer_cy.c
 dfindexer/*.so
 dfindexer/*.pyd
--- a/generator/mavgen_python.py
+++ b/generator/mavgen_python.py
@@ -83,6 +83,23 @@ except Exception:
 
 BytesLike = Union[List[int], Tuple[int], bytes, bytearray, str]
 
+# char and char[] fields are bytes on the wire, but callers have always been
+# allowed to pass a str for them (e.g. mav.statustext_send(sev, "hello")).
+CharFieldLike = Union[bytes, bytearray, str]
+
+
+def char_field_to_bytes(value: CharFieldLike) -> bytes:
+    """Normalise a char/char[] field value to bytes.
+
+    str is encoded as ascii to match the decoding done on the receiving side;
+    characters outside ascii are replaced rather than raising.
+    """
+    if isinstance(value, bytes):
+        return value
+    if isinstance(value, str):
+        return value.encode("ascii", errors="replace")
+    return bytes(value)
+
 
 class _x25crc_slow(object):
     """CRC-16/MCRF4XX - based on checksum.h from mavlink library"""
@@ -517,8 +534,8 @@ def generate_classes(outf, msgs, enums):
         init_fields = []
         for f in m.fields:
             if f.type == "char":
-                init_fields.append("self._%s_raw = %s" % (f.name, f.name))
-                init_fields.append('self.%s = %s.split(b"\\x00", 1)[0].decode("ascii", errors="replace")' % (f.name, f.name))
+                init_fields.append("self._%s_raw = char_field_to_bytes(%s)" % (f.name, f.name))
+                init_fields.append('self.%s = self._%s_raw.split(b"\\x00", 1)[0].decode("ascii", errors="replace")' % (f.name, f.name))
             else:
                 init_fields.append("self.%s = %s" % (f.name, f.name))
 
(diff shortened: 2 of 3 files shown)


fallenmi (changes requested):
The intended input paths work: on exact head `3154af13`, `str`, `bytes`, and `bytearray` produce identical wire bytes for both `char` and `char[]`, and the five added tests pass. I also applied the head over current master `449d03c2`; the same focused tests pass there.

There is one blocking regression in `char_field_to_bytes()`: after the `bytes` and `str` branches, `return bytes(value)` does not only handle the declared `bytearray` case. It also accepts integers, integer iterables, `memoryview`, and even dictionaries with integer keys. The base constructor rejects all of these immediately, but this head makes `MAVLink_statustext_message(6, 100)` produce 100 NUL bytes and successfully pack as an empty STATUSTEXT; `{65: "x"}` packs as text `A`. That turns common caller type errors into valid but unintended telemetry, and a large integer can allocate that many bytes before the fixed-width pack truncates it.

Please handle `bytearray` explicitly and raise `TypeError` for values outside `CharFieldLike`, with a regression covering at least an integer or integer iterable. The exact head currently has no GitHub check contexts or Actions run, so upstream CI also still needs to execute after the update.

Reviewed with OpenAI Codex assistance; I generated dialects from the exact PR base, exact head, and current-master merge and reproduced the constructor and wire-packet behavior locally.

PR #1239, opened 2026-07-31
mavutil: add type annotations to mavfile subclasses (mavserial, mavud…

khancyr:
…p, ...)

Types FakeSerial and all mavfile subclasses: mavserial, mavudp, mavmcast, mavtcp, mavtcpin, mavlogfile, mavmmaplog, mavchildexec, mavwebsocket, and mavwebsocket_client. Widens mavfile.recv()'s return type to bytes | str to honestly reflect what these overrides actually return (several fall back to "" on error paths).

This Only set annotation and don't touch anything else ! No code modification

### DIFF
--- a/mavutil.py
+++ b/mavutil.py
@@ -13,8 +13,9 @@ import copy
 import json
 import re
 import platform
+import mmap
 from types import ModuleType
-from typing import Any
+from typing import Any, IO, cast
 from pymavlink import mavexpression
 import ssl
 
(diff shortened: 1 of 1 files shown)


mavutil.py
```diff
@@ -1136,7 +1138,7 @@ def write(self, buf):
         except socket.error:
             pass
 
-    def recv_msg(self):
+    def recv_msg(self) -> Any:
```
peterbarker, line 1145: Why `Any`?

SHouldn't this be message-or-None?

Similarly elsewhere

PR #1240, opened 2026-08-01
mavgen_c.py: make MAVLINK_*_XML_HASH deterministic

peterbarker:
xml_hash was computed with Python's builtin hash(), which is salted per interpreter run since Python 3.3 (PYTHONHASHSEED), so each generator run emits different MAVLINK_*_XML_HASH values for identical XML.  The hash only needs to be distinct per dialect name (see PR #537, which replaced the order-dependent iterating index); it does not need to be random.

Nondeterministic values churn every dialect header in checked-in generated libraries (e.g. mavlink/c_library_v2) and defeat ccache's direct mode for any source file that includes a generated header: an ArduPilot SITL clean rebuild goes from 42% to 100% direct hits with this change, dropping warm-cache rebuild time from 19s to 5.6s.

Replace it with the first 8 bytes of sha256 of the dialect name, which is stable across runs, Python versions and platforms.

### DIFF
--- a/generator/mavgen_c.py
+++ b/generator/mavgen_c.py
@@ -6,6 +6,7 @@ Copyright Andrew Tridgell 2011
 Released under GNU GPL version 3 or later
 '''
 
+import hashlib
 import os
 from . import mavparse, mavtemplate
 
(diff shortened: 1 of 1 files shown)


peterbarker:
Some numbers:
```
489 distinct generated headers (every MAVLink dialect header) changed content on every generator run via the salted hash(); 817/1412 TUs (58%) include them, so every rebuild forever
  paid a full preprocessor pass per affected file.
```

peterbarker:
`hash` was deterministic when we started to use in (in Python2!).  That changed past Python 3.3

tridge (approved):

PR #1249, opened 2026-08-06
mavutil.py: parameterise the time we spend delaying between reconnect…

peterbarker:
… attempts

sometimes the user knows that it's going to be a very short time between connection going away and coming back again, so allow them to shorten the retry time


khancyr (commented):
LGTM

PR #1250, opened 2026-08-06
mavutil: remove executable behavior

khancyr:
Drops the #!/usr/bin/env python3 shebang and the if __name__ == '__main__' block. mavutil.py is a library module imported as `from pymavlink import mavutil`, not a script meant to be run directly; the __main__ block : a serial port auto-detect dump isn't related to mavutil usage


peterbarker (approved):
LGTM

PR #1251, opened 2026-08-06
mavutil: remove useless class inheritance of Object

khancyr:
leftover from old python2 codebase
https://docs.astral.sh/ruff/rules/useless-object-inheritance/

### DIFF
--- a/mavutil.py
+++ b/mavutil.py
@@ -74,7 +74,7 @@ def evaluate_condition(condition: str | None, vars: dict) -> Any:
 def u_ord(c: Any) -> Any:
     return c
 
-class location(object):
+class location:
     '''represent a GPS coordinate'''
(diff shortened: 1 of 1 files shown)


mavutil.py
```diff
@@ -334,7 +334,7 @@ def recv(self, n=None):
         '''default recv method'''
         raise RuntimeError('no recv() method supplied')
 
-    def close(self):
+    def close(self, n=None):
```
peterbarker: This appears to be mixed in with other changes

peterbarker (changes requested):

mavutil.py
khancyr: .... oh dmaned ... I hate when vscode do this ...

khancyr:
@peterbarker corrected

peterbarker (approved):

PR #1252, opened 2026-08-06
mavutil: tidy exceptions

khancyr:
tidy broad exceptions
remove useless pass
select.error is alias OSError : use OSError to make analyzer happy

### DIFF
--- a/mavutil.py
+++ b/mavutil.py
@@ -295,7 +295,7 @@ class mavfile(object):
             return
         try:
             magic = ord(buf[0])
-        except:
+        except TypeError:
             magic = buf[0]
         if not magic in [ 85, 254, 253 ]:
             return
@@ -356,7 +356,7 @@ class mavfile(object):
             return True
         try:
             (rin, win, xin) = select.select([self.fd], [], [], timeout)
-        except select.error:
+        except OSError:
             return False
         return len(rin) == 1
 
@@ -1000,7 +1000,7 @@ class mavserial(mavfile):
         '''enable/disable RTS/CTS if applicable'''
         try:
             self.port.setRtsCts(enable)
-        except Exception:
+        except AttributeError:
             self.port.rtscts = enable
         self.rtscts = enable
 
@@ -1008,7 +1008,7 @@ class mavserial(mavfile):
         '''set baudrate'''
         try:
             self.port.setBaudrate(baudrate)
-        except Exception:
+        except AttributeError:
             # for pySerial 3.0, which doesn't have setBaudrate()
             self.port.baudrate = baudrate
     
@@ -1098,7 +1098,7 @@ class mavudp(mavfile):
     def recv(self,n=None):
         try:
             data, new_addr = self.port.recvfrom(UDP_MAX_PACKET_LEN)
-        except socket.error as e:
+        except OSError as e:
             if e.errno in [ errno.EAGAIN, errno.EWOULDBLOCK, errno.ECONNREFUSED ]:
                 return ""
             raise
@@ -1133,7 +1133,7 @@ class mavudp(mavfile):
                     self.resolved_destination_addr = self.destination_addr[0]
                     self.destination_addr = (socket.gethostbyname(self.destination_addr[0]), self.destination_addr[1])
                 self.port.sendto(buf, self.destination_addr)
-        except socket.error:
+        except OSError:
             pass
 
     def recv_msg(self):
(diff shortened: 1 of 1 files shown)


mavutil.py
```diff
@@ -1943,15 +1942,15 @@ def connect(self):
...
-        except socket.error as e:
-            if e.errno in [errno.ECONNREFUSED, errno.EHOSTUNREACH]:
-                self.close()
-                return
-            raise
         except ssl.SSLError as e:
             print(f"SSL Error: {e}")
             self.close()
             raise
+        except OSError as e:
```
khancyr, line 1948: just a reorder as oserror cover sslerror , so we need to handle them first

peterbarker:
I do like the change, but there are several regressions.

I suggest maybe subset the easier ones to get them in first....


```
  ⎿  mavutil.py
       ●  994 [correctness] mavserial.__init__ narrows the fileno() guard to (AttributeError, serial.SerialException), but pySerial's Windows backend inherits io.RawIOBase.fileno(),
                            which raises io.UnsupportedOperation (an OSError/ValueError subclass, not SerialException) — so the exception now escapes. Codex independently flags the same
                            probe (and mavutil.py:1055) for ValueError from a closed file-like object. Same root cause at mavutil.py:1058.
       ● 1033 [correctness] mavserial.write narrows to (serial.SerialException, OSError), which no longer catches the bare `Exception` that FakeSerial.write deliberately raises,
                            breaking the force_connected / dead-device path.
       ●  132 [correctness] set_dialect only auto-regenerates a dialect module on ImportError, so a corrupt or partially written generated dialect no longer self-heals.
       ●  588 [correctness] param_fetch_one moves param_request_read_send out of the guarded block into an `else:` branch, so a numeric name that overflows the int16 param_index now
                            raises struct.error instead of falling back to a by-name request. Independently confirmed by Codex as the one structural (non-tuple) regression in the PR.
       ●  955 [correctness] set_close_on_exec, documented as "Ignore exceptions", narrowed from `except Exception` to `(ImportError, OSError)`, but fcntl.fcntl raises
                            ValueError/TypeError for bad fds. Codex flags the same gap independently.
       ●   29 [correctness] Top-level optional `from pymavlink.dialects.v10 import all` guard narrowed from `except Exception` to `except ImportError`, so any non-ImportError failure in
                            the generated module now aborts importing mavutil. Codex reaches the same conclusion and notes it contradicts the comment's stated best-effort intent.
       ● 2221 [cleanup]     auto_detect_serial_win32 narrowed from a bare `except:` to (ImportError, OSError), so non-ImportError/non-OSError failures in pySerial's win32 enumeration
                            now escape instead of degrading to an empty list.
       ● 1064 [cleanup]     mavserial.reset narrowed to (serial.SerialException, OSError), so non-OSError failures now escape reset() and re-escape write()'s dead-port handler,
                            violating its boolean failure contract. Codex flags the same, adding ValueError from `serial.Serial(...)` on an invalid baud as a second trigger.
       ● 1030 [cleanup]     A redundant `import serial` was added to the per-packet write() hot path, and serial.SerialException is already an OSError subclass so the tuple collapses to
                            `except OSError`.
       ● 2486 [cleanup]     `except OSError: raise` is a no-op handler left in place after the narrowing.
```

I'm particularly concerned about the win32 implications.

khancyr:
@peterbarker I tried to keep the one's that shouldn't be blocking. Other may need so other change so part of another PR

mavutil.py
```diff
@@ -2493,14 +2491,14 @@ def mode_string_v09(msg):
...
-    except Exception:
+    except (ValueError, AttributeError, TypeError):
         # inform the user of invalid custom mode map
         print("Error: invalid pymavlink custom mode map dict in " + _custom_mode_map_path)
         raise
 
     AP_MAV_TYPE_MODE_MAP = AP_MAV_TYPE_MODE_MAP_DEFAULT.copy()
     AP_MAV_TYPE_MODE_MAP.update(_custom_mode_map)
-except Exception:
+except (OSError, ValueError, AttributeError, TypeError, json.decoder.JSONDecodeError):
```
peterbarker, line 2501: Please revert this one.  You're missing at least `RecursionError`

I'd actually like to  consider removing the entire try/except block here on the basis that the mode mapping file is relatively static and the user probably *does* want an early-exit with mavutil.py is *imported*!

mavutil.py
```diff
@@ -1000,15 +1000,15 @@ def set_rtscts(self, enable):
         '''enable/disable RTS/CTS if applicable'''
         try:
             self.port.setRtsCts(enable)
-        except Exception:
+        except AttributeError:
```
peterbarker, line 1003:
```suggestion
        except (AttributeError, NotImplementedError):
```

peterbarker (changes requested):
I don't think I will be able to approve/merge these, I'm afraid, I think tridge will need to do that.

We would probably need to do an audit of MAVProxy to make sure that its resilience is not impacted by this.  Changing the API from "catch exception, return value" to what it should have been (if you can't handle an exception then don't look for the problem) is *not* a small change to be making!

PR #1256, opened 2026-08-10
Fixing message reading in Ada

ReladormAndry:
Prevent exceptions when data is broken by checking the CRC before conversion.

### DIFF
--- a/generator/Ada/mavlink-v1.adb
+++ b/generator/Ada/mavlink-v1.adb
@@ -230,8 +230,8 @@ package body MAVLink.V1 is
         Address => Incoming.In_Buf'Address;
       Last_Data : constant Positive := Incoming.In_Buf'First +
         Packet_Payload_First +
-          Natural (Header.Len - 1);
-      Last : constant Natural := Buffer'First + Natural (Header.Len - 1);
+          Natural (Header.Len) - 1;
+      Last : constant Natural := Buffer'First + Natural (Header.Len) - 1;
    begin
       Buffer (Buffer'First .. Last) := Incoming.In_Buf
         (Incoming.In_Buf'First + Packet_Payload_First .. Last_Data);
--- a/generator/Ada/v2/mavlink-v2.adb
+++ b/generator/Ada/v2/mavlink-v2.adb
@@ -123,8 +123,8 @@ package body MAVLink.V2 is
       if Incoming.Last = 0 then
          Incoming.Last := Incoming.Income_Buffer'First +
            Packet_Payload_First + --  header
-             Natural (Header.Len - 1) + --  data len
-           2; --  x25crc checksum
+             Natural (Header.Len) + --  data len
+           1; --  x25crc checksum
 
          if (Header.Inc_Flags and 1) > 0 then
             Incoming.Last := Incoming.Last + 13; --  SHA256 signature
@@ -411,7 +411,7 @@ package body MAVLink.V2 is
         Address => Incoming.Income_Buffer
           (Incoming.Income_Buffer'First +
              Packet_Payload_First +
-               Natural (Header.Len - 1) + 3)'Address;
+               Natural (Header.Len) + 2)'Address;
    begin
       if (Header.Inc_Flags and 1) > 0 then
          return Sig.Link_Id;
@@ -458,7 +458,7 @@ package body MAVLink.V2 is
          declare
             Last_Data   : constant Positive := Incoming.Income_Buffer'First +
               Packet_Payload_First +
-                Natural (Header.Len - 1);
+                Natural (Header.Len) - 1;
             Sig         : constant MAVLink.V2.MAV_Signature with Import,
               Address => Incoming.Income_Buffer (Last_Data + 3)'Address;
             Message_SHA : SHA_Digest;
@@ -649,16 +649,18 @@ package body MAVLink.V2 is
       Buffer   : out Data_Buffer;
       Last     : out Natural)
    is
-      Header    : constant V2_Header with Import,
+      Header     : constant V2_Header with Import,
         Address => Incoming.Income_Buffer'Address;
-      Last_Data : constant Positive := Incoming.Income_Buffer'First +
-        Packet_Payload_First +
-          Natural (Header.Len) - 1;
+      First_Data : constant Natural :=
+        Incoming.Income_Buffer'First +
+          Packet_Payload_First;
+      Len        : Natural;
    begin
-      Last := Buffer'First + Natural (Header.Len - 1);
+      Len  := Natural'Min (Natural (Header.Len), Buffer'Length);
+      Last := Buffer'First + Len - 1;
+
       Buffer (Buffer'First .. Last) := Incoming.Income_Buffer
-        (Incoming.Income_Buffer'First + Packet_Payload_First ..
-           Last_Data);
+        (First_Data .. First_Data + Len - 1);
    end Get_Message_Data;
 
    ----------------------
@@ -684,6 +686,36 @@ package body MAVLink.V2 is
    begin
       Get_Message_Data (Self.Incoming, Buffer, Last);
    end Get_Message_Data;
+
+   --------------------
+   -- Get_Msg_Length --
+   --------------------
+
+   function Get_Msg_Length (Incoming : Incoming_Data) return Interfaces.Unsigned_8
+   is
+      Header : constant V2_Header with Import,
+        Address => Incoming.Income_Buffer'Address;
+   begin
+      return Header.Len;
+   end Get_Msg_Length;
+
+   --------------------
+   -- Get_Msg_Length --
+   --------------------
+
+   function Get_Msg_Length (Self : Connection) return Interfaces.Unsigned_8 is
+   begin
+      return Get_Msg_Length (Self.Incoming);
+   end Get_Msg_Length;
+
+   --------------------
+   -- Get_Msg_Length --
+   --------------------
+
+   function Get_Msg_Length (Self : In_Connection) return Interfaces.Unsigned_8 is
+   begin
+      return Get_Msg_Length (Self.Incoming);
+   end Get_Msg_Length;
 
    ------------
    -- Encode --
--- a/generator/Ada/v2/mavlink-v2.ads
+++ b/generator/Ada/v2/mavlink-v2.ads
@@ -94,6 +94,8 @@ package MAVLink.V2 is
       return Boolean with Inline;
    --  Add the Value to the buffer and return True if a message has been read
    --  (collected all message data)
+   --  All methods that work with incoming data (like Get_*, Check_*) should be
+   --  called only when this method returned True.
 
    --  Get the message's information that is in the connection's buffer --
    procedure Get_Message_Information
@@ -121,6 +123,10 @@ package MAVLink.V2 is
       Comp_Id : out Component_Id_Type;
       Id      : out Msg_Id) with Inline;
    --  Same as above but do not check V2 signature
+
+   function Get_Msg_Length
+     (Self : Connection) return Interfaces.Unsigned_8 with Inline;
+   --  Returns length field from the packet header from the incoming buffer
 
    function Get_Message_Id (Self : Connection) return Msg_Id with Inline;
    --  Returns message's ID
@@ -188,6 +194,10 @@ package MAVLink.V2 is
       Sys_Id  : out System_Id_Type;
       Comp_Id : out Component_Id_Type;
       Id      : out Msg_Id) with Inline;
+
+   function Get_Msg_Length
+     (Self : In_Connection) return Interfaces.Unsigned_8 with Inline;
+   --  Returns length field from the packet header from the incoming buffer
 
    function Get_Message_Id (Self : In_Connection) return Msg_Id with Inline;
 
@@ -323,6 +333,9 @@ private
      (Self   : Connection;
       Buffer : out Data_Buffer;
       Last   : out Natural);
+   --  Copy data from the internal incoming buffer to Buffer,
+   --  and set Last to the index of the last copied byte.
+   --  Truncate the data if Buffer is too small to hold it all.
 
    -- In_Connection --
 
@@ -339,6 +352,9 @@ private
      (Self   : In_Connection;
       Buffer : out Data_Buffer;
       Last   : out Natural);
+   --  Copy data from the internal incoming buffer to Buffer,
+   --  and set Last to the index of the last copied byte.
+   --  Truncate the data if Buffer is too small to hold it all.
 
    -- Out_Connection --
 
@@ -387,6 +403,8 @@ private
       Link_Id   : out Link_Id_Type;
       Timestamp : out Timestamp_Type;
       Signature : out Three_Boolean);
+
+   function Get_Msg_Length (Incoming : Incoming_Data) return Interfaces.Unsigned_8;
 
    function Get_Message_Id (Incoming : Incoming_Data) return Msg_Id;
 
--- a/generator/mavgen_ada.py
+++ b/generator/mavgen_ada.py
@@ -572,27 +572,37 @@ def generate_message_methods(name, rev, spec, is_v1):
      (Message   : out {0};
       Connect   : {2}{1}.Connection;
       CRC_Valid : out Boolean);
-   --  Get the message from the Connect and delete it
-   --  from the Connect's buffer. CRC_Valid is set to
-   --  True if x25crc is valid for the message.
+   --  Get the message from the Connect if x25crc is valid and
+   --  set CRC_Valid to True.
+   --  Won't read data from the Connect if x25crc is False.
+   --  For v1: Won't read data from the Connect if message length mismatch
+   --    and set CRC_Valid to False in this case.
+   --  For v2: Truncate the extension fields.
 
    procedure Decode
      (Message : out {0};
       Connect : {1}.Connection);
-   --  Same as Above but does not check CRC
+   --  Get the message from the Connect.
+   --  For v1: May raise exception when message length mismatch
+   --  For v2: Truncate the extension fields.
 
    procedure Decode
      (Message   : out {0};
       Connect   : {2}{1}.In_Connection;
       CRC_Valid : out Boolean);
-   --  Get the message from the Connect and delete it
-   --  from the Connect's buffer. CRC_Valid is set to
-   --  True if x25crc is valid for the message.
+   --  Get the message from the Connect if x25crc is valid and
+   --  set CRC_Valid to True.
+   --  Won't read data from the Connect if x25crc is False.
+   --  For v1: Won't read data from the Connect if message length mismatch
+   --    and set CRC_Valid to False in this case.
+   --  For v2: Truncate the extension fields.
 
    procedure Decode
      (Message : out {0};
       Connect : {1}.In_Connection);
-   --  Same as Above but does not check CRC
+   --  Get the message from the Connect.
+   --  For v1: May raise an exception when message length mismatch
+   --  For v2: Truncate the extension fields.
 
    function Check_CRC
      (Connect : {2}{1}.Connection)
@@ -698,16 +708,21 @@ def generate_message_body_decode(name, rev, extra, body, is_v1, connection):
 """ % (name, mode, rev, connection))
 
     if is_v1:
-        body.write("""      Buf : Data_Buffer
-        (1 .. %s'Size / 8) := [others => 0]
-        with Address => Message'Address,
-        Convention   => Ada;
+        body.write("""      Buf : Data_Buffer (1 .. %s'Size / 8)
+        with Import,
+        Address    => Message'Address,
+        Convention => Ada;
    begin
-      pragma Assert
-        (Get_Msg_Len (Connect) =
-             Unsigned_8 (Integer (%s'Value_Size) / 8));
-      Get_Message_Data (Connect, Buf);
-      CRC_Valid := Check_CRC (Connect);
+      if Get_Msg_Len (Connect) =
+        Unsigned_8 (Integer (%s'Value_Size) / 8)
+      then
+         CRC_Valid := Check_CRC (Connect);
+         if CRC_Valid then
+            Get_Message_Data (Connect, Buf);
+         end if;
+      else
+         CRC_Valid := False;
+      end if;
 """ % (name, name))
     else:
         body.write("""      Data : Data_Buffer (1 .. %s'Value_Size / 8);
@@ -717,9 +732,11 @@ def generate_message_body_decode(name, rev, extra, body, is_v1, connection):
         with Address => Message'Address,
         Convention   => Ada;
    begin
-      Get_Message_Data (Connect, Data, Last);
-      Buf (1 .. Last) := Data (1 .. Last);
       CRC_Valid := Check_CRC (Connect);
+      if CRC_Valid then
+         Get_Message_Data (Connect, Data, Last);
+         Buf (1 .. Last) := Data (1 .. Last);
+      end if;
 """ % (name, name))
     body.write("""   end Decode;
 
@@ -1146,6 +1163,7 @@ pragma Warnings (Off); --  prevent "not used"
 with MAVLink.Raw_Floats;      use MAVLink.Raw_Floats;
 with MAVLink.Raw_Long_Floats; use MAVLink.Raw_Long_Floats;
 with MAVLink.SHA_256;
+with Interfaces;
 pragma Warnings (On);
 
 use MAVLink.V2;
@@ -1230,6 +1248,8 @@ is""");
    Buffer      : Data_Buffer (1 .. MAVLink.V2.Maximum_Buffer_Len);
    Last        : Positive;
 
+   use type Interfaces.Unsigned_8;
+
 begin
    Initialize (Sig,  1, Pass_Data, 200);
    Set_System_Id (In_Connect, 1);
@@ -1251,6 +1271,8 @@ begin
       Res := Parse_Byte (In_Connect, Hygrometer_Sensors_Data (Index));
    end loop;
    pragma Assert (Res);
+
+   pragma Assert (Get_Msg_Length (In_Connect) = 5);
 
    Get_Message_Information
      (In_Connect, Sig, Seq, Sys_Id, Comp_Id, Id, Link_Id, Timestamp, Signature);


peterbarker:
Sorry, I don't know Ada.

But Claude and Codex do, and they seem to think there might be some problems.  I eyeballed these quickly and they do seem legit to this layman.

```
● Code review(high · 7 findings)
  ⎿  generator/mavgen_ada.py
       ● 708 [correctness] The v1 guarded-Decode template still validates length only via pragma Assert after Check_CRC; since a CRC is
                           computable for any Len, an oversized-Len v1 packet with a valid checksum raises Assertion_Error (or overflows
                           Buf with assertions disabled) instead of setting CRC_Valid to False.
       ● 728 [correctness] Treating Last = 0 from Get_Message_Data as CRC failure rejects valid v2 packets whose payload carries unknown
                           trailing extension bytes from a newer dialect; MAVLink requires receivers to accept these by ignoring the
                           extra bytes.
       ● 706 [correctness] Decode previously always copied the received payload into Message before computing CRC_Valid; now Message
                           stays as the [others => 0] initializer whenever CRC_Valid is False, so the payload of CRC-failed packets is no
                           longer observable.
       ● 712 [correctness] The re-emitted generated code carries trailing whitespace after 'end if;' and non-standard 8/12-space
                           indentation (vs GNAT 3-space style), so generated dialect sources no longer pass GNAT style checking.
     generator/Ada/v2/mavlink-v2.adb
       ● 666 [correctness] The removed implicit Constraint_Error on oversized payload is replaced by a silent Last sentinel, but the
                           generated no-CRC Decode overload (mavgen_ada.py ~762) was not updated to check it, so it silently returns an
                           all-zero message.
       ● 658 [correctness] Header.Len = 0 wraps modularly in Natural (Header.Len - 1) to 255, so the new Last <= Buffer'Last guard
                           mis-handles zero-length packets: for caller buffers of 256+ elements the guard passes and the slice assignment
                           raises Constraint_Error (length mismatch 256 vs 0).
       ● 666 [cleanup]     The failure sentinel Last := Buffer'First - 1 paired with the generated check Last > 0 only works for 1-based
                           buffers; non-1-based public-API callers misread failure as success.
```

ReladormAndry:
@peterbarker Thank you! That information was very helpful. I have fixed the issues mentioned in it.

peterbarker:
> @peterbarker Thank you! That information was very helpful. I have fixed the issues mentioned in it.


Well, I still don't speak Ada... so I'm just going to be a meatsack intermediary again:

```
  ⎿  generator/Ada/v2/mavlink-v2.adb
       ●  668 [correctness]    `Check` gates the payload copy itself, so the `Check => False` path never copies any data and always
                               returns the failure sentinel `Last := -1`.
       ●  675 [correctness]    A payload longer than the receiver's message definition is reported as a CRC failure, breaking MAVLink v2
                               forward compatibility with extension fields.
       ●  657 [correctness]    The new `Header.Len = 0` guard is unreachable in practice because Parse_Byte's `Natural (Header.Len - 1)`
                               still wraps modularly for a zero-length payload.
       ●  658 [correctness]    The zero-length branch returns a different failure sentinel (`Buffer'First - 1`, i.e. 0) from the `-1`
                               that the generated checked decoder tests for.
     generator/mavgen_ada.py
       ●  725 [correctness]    `Decode` no longer populates `Message` when the CRC check fails; `Get_Message_Data` is now gated behind
                               `if CRC_Valid`, so callers that treat the flag as advisory silently read zeros.
       ●  752 [correctness]    The v1 non-CRC Decode still uses `pragma Assert` for the length check, so the fix reaches only 3 of the 4
                               emitted Decode variants.
       ● 1375 [test-coverage]  The generated self-test never exercises the non-CRC Decode or a malformed-length packet, so this
                               regression ships green.
       ●  706 [simplification] Length validation is duplicated into every generated v1 message body instead of living once in the
                               runtime, unlike v2.
     generator/Ada/v2/mavlink-v2.ads
       ●  429 [documentation]  The new `Check` parameter and the negative failure sentinel are undocumented, and `Last` silently widens
                               from Natural to Integer.
```

ReladormAndry:
@peterbarker Thanks, I fixed the issues that the AI pointed out.

peterbarker:
So I had my robots do lunch and they came up with a few things.

*Please* feel free to push back if you think they're talking crap.  They always sound like they know what they're talking about.  A quick look at the "blockers" seems to say that's true in this case (to me, at least).

```
  Combined review — ArduPilot/pymavlink#1256, "Fixing message reading in Ada"

  Three passes: two independent reviews (one of which had a Codex second opinion), plus a targeted Codex adjudication of the two points they
  disagreed on. Line numbers are on PR head d9bdab67c (local branch pr1256). No Ada toolchain is installed, so nothing was compiled — all
  findings are from reading the generator and the Ada it emits (--lang=Ada, both wire protocols, against minimal.xml).

  Verdict

  The central fix is correct, but incomplete and asymmetric. The same class of bug the PR fixes in one place survives in three others,
  including in the same file.

  What's right: Get_Message_Data at generator/Ada/v2/mavlink-v2.adb:649-664. The old Last := Buffer'First + Natural (Header.Len - 1) did the -
  1 in modular Unsigned_8, so Len = 0 wrapped to 255 and produced a 256-byte slice assignment into a 9-byte buffer. The replacement —
  Natural'Min (Natural (Header.Len), Buffer'Length), then Buffer'First + Len - 1 — is correct at both ends, handles truncated MAVLink 2
  payloads and over-long payloads from newer peers, and the offsets match the old code exactly. Reordering Check_CRC before Get_Message_Data
  is also safe: Is_CRC_Valid is side-effect-free for V2, and for V1 its one side effect (Extras_Added) is reset per message in Parse_Byte.

  Findings

  1. generator/Ada/v2/mavlink-v2.adb:126 — high. The identical underflow survives in Parse_Byte, making the fix unreachable

  Both reviews, independently.

  Incoming.Last := Income_Buffer'First + Packet_Payload_First
                   + Natural (Header.Len - 1) + 2;

  At Len = 0 this is 1 + 10 + 255 + 2 = 268 instead of 12; with the incompat-flags signature bit set, 281 — past Income_Buffer'Last = 280.
  Concrete failure: feed 0xFD, 0x00, 0x01, …. Parse_Byte returns True at Position = 280 via the Position = Income_Buffer'Last fallback; on the
  next byte the guard Incoming.Position >= Incoming.Last reads 280 >= 281 = False, Clear is skipped, and Income_Buffer (281) := Value raises
  Constraint_Error — or corrupts memory with checks suppressed. Even unsigned, Len = 0 makes the parser swallow 268 bytes and desynchronise,
  discarding real messages inside.

  Because framing is lost here, the PR's new Len = 0 handling downstream is never reached. MAVLink.V1.Parse_Byte uses Natural (In_Buf 
  (Pos_Len)) with no - 1 and is correct — only V2 has this.

  Two more sites share the pattern at lower severity: mavlink-v2.adb:414 and :461, both in the signature path. At Len = 0 they address
  Income_Buffer (269) — in bounds, but the wrong offset, so they read garbage as a signature rather than crashing.

  2. generator/Ada/mavlink-v1.adb:233-234 — medium. V1 got no equivalent fix

  Both reviews.

  Last_Data := In_Buf'First + Packet_Payload_First + Natural (Header.Len - 1);
  Last      := Buffer'First + Natural (Header.Len - 1);

  Still wraps at Len = 0, and still never clamps to Buffer'Length. A V1 frame with Len = 0 (8 bytes — Parse_Byte accepts it and reports
  complete) yields Buffer (1 .. 256) := In_Buf (7 .. 262) into the 9-byte Buf overlaying the caller's record: Constraint_Error with checks on,
  silent stack corruption under -gnatp. Applying the same Natural'Min (Natural (Header.Len), Buffer'Length) treatment makes the fix
  symmetric.

  3. generator/mavgen_ada.py:752 — medium. The V1 no-CRC Decode is still guarded only by pragma Assert

  The PR upgraded the CRC-checking V1 overload to a real if (line 710) but left the no-CRC overload emitting pragma Assert (Get_Msg_Len 
  (Connect) = …) followed by an unconditional Get_Message_Data. The generated mavlink_v1.gpr passes -gnata, but library consumers building
  their own project without it get no check at all — and the call lands in the unfixed Get_Message_Data from finding 2. Fixing finding 2 also
  covers this; fixing this alone does not cover finding 2.

  4. generator/mavgen_ada.py:710 — medium. V1 reports a pure length mismatch as CRC_Valid := False

  Raised by one review; confirmed by Codex. Generated V1 checked Decode:

  if Get_Msg_Len (Connect) = Expected_Length then
     CRC_Valid := Check_CRC (Connect);
     ...
  else
     CRC_Valid := False;
  end if;

  A packet with a mathematically valid checksum but a payload length differing from the selected message definition returns CRC_Valid = False
  without its CRC ever being checked. V2 differs: it checks CRC directly (:729) and lets Get_Message_Data clamp, so a valid longer-or-shorter
  V2 payload keeps CRC_Valid = True.

  Codex's assessment, which I agree with: rejecting the packet is correct — V1 has fixed payload lengths and no extension fields, so a length
  mismatch means it can't be safely decoded as that type. The problem is diagnostic precision. Corrupted packet, wrong decoder selected,
  incompatible dialect, and malformed-but-correctly-checksummed all collapse into one CRC_Valid = False, in an out-parameter whose name and
  doc comment specifically promise "x25crc is valid". Cheapest honest fix is a doc-comment correction; the cleaner one is a separate
  length/validity output or renaming to Message_Valid.

  5. generator/mavgen_ada.py:713-716, 729-733 — low, but the two reviews disagreed here

  Dispute: on CRC failure the PR now leaves Message zero-filled rather than holding the received bytes. Reviewer A called it a positive (no
  stale data leaks); Reviewer B called it an unmentioned behaviour change that also discards the generator's per-field "invalid" sentinel
  defaults.

  Resolution — Reviewer B is half wrong, and I verified the half that matters. The := [others => 0] initializer on the address-overlaid Buf is
  not new: it is on master at generator/mavgen_ada.py:701 (V1) and :717 (V2), unchanged by this PR. Since that initializer writes zeros
  across the whole message representation at elaboration, the per-field invalid defaults (emitted at :829 V1, :1110 V2, via get_default() at
  :483) were already being clobbered before this PR. That is a real pre-existing bug — it also means successfully-decoded shortened MAVLink 2
  messages get zeros for omitted extension fields instead of their XML invalid sentinels — but it is not this PR's and shouldn't block it.

  What is new is only the CRC-failure path: Message used to contain untrusted received bytes, now contains zeros. For any caller honouring the
  contract (read Message only when CRC_Valid = True) this is an improvement, so Reviewer A's characterisation stands. The one regression is
  for a caller deliberately inspecting corrupt decoded fields for diagnostics — which should be using the raw frame buffer anyway.
  Recommendation: not a blocker; worth one line in the commit message.

  6. generator/mavgen_ada.py:578, 583, 599 — low. Comment regressions

  Two separate problems in the rewritten spec comments:

  - The no-CRC overloads lost --  Same as Above but does not check CRC. Both Decode overloads now read as though they do the same thing, with
  nothing in the generated .ads warning that one performs no CRC validation. That works directly against the PR's own thesis.
  - --  Truncate the extension fields is emitted into the MAVLink 1 spec too (generate_message_methods is called with is_v1=True at :832, and
  the comment block is shared). V1 has no extension fields and, per finding 2, no Buffer'Length clamp — it overruns rather than truncating.
  Confirmed in the generated mavlink-v1-minimal-heartbeats.ads. The V1 spec now documents behaviour the V1 runtime does not implement.
  - --  Truncate the extension fields is emitted into the MAVLink 1 spec too (generate_message_methods is called with is_v1=True at :832, and the comment block is shared). V1 has no
  extension fields and, per finding 2, no Buffer'Length clamp — it overruns rather than truncating. Confirmed in the generated mavlink-v1-minimal-heartbeats.ads. The V1 spec now documents
  behaviour the V1 runtime does not implement.

  7. generator/Ada/v2/mavlink-v2.ads:125, 195, 400 — low. Get_Data_Length is dead, undocumented, and unsafe before Parse_Byte

  Both reviews. Nothing in mavgen_ada.py, the generated dialects, the examples, or the generated tests calls it, so it ships untested. It reads Header.Len straight out of Income_Buffer,
  which has no default initialization — on a fresh Connection, or any time before Parse_Byte returns True, it returns an uninitialized or stale byte. Get_Message_Information carries "Should
  be called only after Parse_Byte returned True"; this needs the same caveat, since its obvious use (sizing a Data_Buffer ahead of Get_Message_Data) is exactly where a garbage value bites.
  It also lacks the with Inline every neighbouring accessor has, and has no MAVLink.V1 counterpart.

  Suggested disposition

  ┌─────┬───────────────────────┬──────────┬──────────────────────────────────────────┐
  │  #  │         Site          │ Severity │                 Blocker?                 │
  ├─────┼───────────────────────┼──────────┼──────────────────────────────────────────┤
  │ 1   │ mavlink-v2.adb:126    │ high     │ Yes — makes the PR's own fix unreachable │
  ├─────┼───────────────────────┼──────────┼──────────────────────────────────────────┤
  │ 2   │ mavlink-v1.adb:233    │ medium   │ Yes — same bug class, left in place      │
  ├─────┼───────────────────────┼──────────┼──────────────────────────────────────────┤
  │ 3   │ mavgen_ada.py:752     │ medium   │ Subsumed by 2                            │
  ├─────┼───────────────────────┼──────────┼──────────────────────────────────────────┤
  │ 4   │ mavgen_ada.py:710     │ medium   │ No — doc fix suffices                    │
  ├─────┼───────────────────────┼──────────┼──────────────────────────────────────────┤
  │ 5   │ mavgen_ada.py:713/729 │ low      │ No — note in commit message              │
  ├─────┼───────────────────────┼──────────┼──────────────────────────────────────────┤
  │ 6   │ mavgen_ada.py:578/583 │ low      │ No                                       │
  ├─────┼───────────────────────┼──────────┼──────────────────────────────────────────┤
  │ 7   │ mavlink-v2.ads:125    │ low      │ No                                       │
  └─────┴───────────────────────┴──────────┴──────────────────────────────────────────┘

  Findings 1 and 2 are the ones to push back on: the PR is titled "fixing message reading in Ada" and demonstrably fixes one of four instances of a single bug.
```

peterbarker:
(incidentally, it's pretty clear that we need some regression tests here....)

ReladormAndry:
@peterbarker Fixed the issues. Thanks for the help. Too bad I can't access the AI report myself; thank you for taking the extra time to post it.

peterbarker:
Another round!
```
● Review of PR #1256 — "Fixing message reading in Ada"

  I read the diff, traced the index arithmetic against the buffer bounds, generated the Ada output for both wire protocols to inspect what's
  actually emitted, and ran Codex over it independently. Codex and I agree on everything below; no contradictions.

  The core fixes are correct

  Header.Len is Interfaces.Unsigned_8 (modular), so Natural (Header.Len - 1) wrapped to 255 at Len = 0. Every site the PR touches is now
  right, and I verified the boundaries both ways:

  - v2 (Packet_Payload_First = 10, Income_Buffer (1 .. 280)): payload 11..10+Len, CRC at 11+Len/12+Len, signature at 13+Len. Incoming.Last = 
    Len + 12 (+13 signed). At Len=255 signed: 280 exactly. ✓
  - v1 (Packet_Payload_First = 6, In_Buf (1 .. 263)): last payload byte Len+6, CRC ends at Len+8 → 263 at Len=255. ✓
  - Get_Link_Id / Check_Signature offsets land on 13+Len in both old and new code for Len≥1, and are now correct for Len=0. ✓
  - The new truncating Get_Message_Data (mavlink-v2.adb:649) is fine for Len = 0 — both sides degenerate to null slices — and doesn't assume
    Buffer'First = 1.

  The non-obvious risk is clean too: moving Check_CRC before Get_Message_Data is only safe if the CRC check has no destructive side effect.
  v1's Is_CRC_Valid is in out — but it only folds Extras into the running checksum, guarded by Extras_Added, and never touches In_Buf
  (mavlink-v1.adb:183-195). v2's Is_CRC_Valid indexes at most 267 for any Header.Len, so it can't raise before the clamp can help. Reordering
  is sound.

  Findings

  1. (Moderate) The stated goal isn't met for v1 — the non-CRC Decode overloads still crash.
  generator/mavgen_ada.py:750-759 still guards length with pragma Assert, which is a no-op without -gnata. It then calls Get_Message_Data, and
  v1's version (generator/Ada/mavlink-v1.adb:225-238) got the arithmetic fix but not the Natural'Min clamp its v2 sibling gained. Decoding a
  header claiming Len = 255 into a 9-byte Heartbeat overlay raises Constraint_Error in a release build. v2's unchecked Decode is now safe;
  v1's is not.

  Cheapest fix: mirror the clamp into mavlink-v1.adb:225's Get_Message_Data. One line, no API change, covers all four v1 overloads.

  2. (Moderate) v1 checked Decode now leaves out Message undefined; v2 doesn't.
  Dropping := [others => 0] for with Import (mavgen_ada.py:708-712) means that on either failure path (length mismatch or bad CRC) the
  procedure returns with CRC_Valid = False and Message never written. The v2 checked overload still has its [others => 0] overlay
  (mavgen_ada.py:729), so it deterministically zeroes Message. Same documented API, two different contracts — and the new comment block
  doesn't mention it. Field-level invalid defaults emitted at mavgen_ada.py:833 aren't preserved either.

  3. (Low) Generated Ada is misindented. mavgen_ada.py:721-723 emits the else / CRC_Valid := False; / end if; at columns 8/11/8 instead of
  7/10/7. Confirmed in generated output — it won't fail the build (no -gnaty in the test .gpr) but it's visibly wrong.

  4. (Low) Get_Message_Length nits.
  - Name diverges from v1's existing Get_Msg_Len; v2 has no matching accessor name, so callers writing version-agnostic code get two
    spellings.
  - Missing with Inline, unlike every neighbouring accessor in that spec.
  - Typo "lenght" in both comments (mavlink-v2.ads:128 and :198).
  - Get_Message_Length(In_Connect) at mavgen_ada.py:1273 is missing the space before (.

  5. (Low) Truncation doc added to only one overload. mavlink-v2.ads documents the new truncating behaviour on the Connection Get_Message_Data
  but not the In_Connection one right below it.

  6. (Info) The new test assertion is conditional, and nothing compiles this anyway. The pragma Assert (Get_Message_Length ...) sits inside if
  Has_Common:, so it only appears for dialects that include common.xml — generating from minimal.xml emits with Interfaces; and use type 
  Interfaces.Unsigned_8; with no use of either. Harmless (the with is inside the pragma Warnings (Off) block), but worth knowing. More
  significantly: .github/workflows/ has no Ada job at all, so none of this generated Ada is ever compiled in CI — the fix can't be validated
  automatically, and neither can a regression.

  Suggested verdict

  Approve on the arithmetic and the v2 truncation — those are real bugs correctly fixed. Ask for #1 (v1 clamp) before merge, since without it
  the PR's own stated goal is only half-delivered; #2 deserves at least a documented decision. The rest are nits.

  Want me to post these as PR comments (gh pr review --repo ArduPilot/pymavlink), or draft the one-line v1 clamp as a suggested change?
```

.... let me know if you've had enough and I'll merge this :-)

peterbarker:
> @peterbarker Fixed the issues. Thanks for the help. Too bad I can't access the AI report myself; thank you for taking the extra time to post it.

No worries.  I'm just prompting claude each time.

ReladormAndry:
Did some minimal changes, added comments and renamed Get_Message_Length to Get_Msg_Length as AI suggested.

peterbarker:
@ReladormAndry fresh review below, and I've asked for a review from tridge's pair of bots.

However, if tridge's bots are silent or don't say much then I'm going to assume you're now happy with this PR and will merge it.  If not, yell :-)

```
● Codex and I agree, and both landed on the same two items. Here's where this
  round stands.

  Round status — what the author fixed vs. what's still open

  I regenerated Ada for both wire protocols (--lang=Ada, --wire-protocol=1.0 and
  2.0, against common.xml+minimal.xml) and read the emitted source. No Ada
  toolchain here, so again nothing was compiled.

  Fixed since last round — all confirmed in the generated output:
  - mavlink-v2.adb:126 Parse_Byte modular underflow (the high-severity blocker)
    — now Natural (Header.Len) + 1. Frame arithmetic verified at both ends:
    Maximum_Buffer_Len = 280 = 10+255+2+13 exactly; v1's 263 = 6+255+2 exactly.
  - Signature offsets at :414 and :461.
  - Get_Message_Length → Get_Msg_Length, with Inline added (mavlink-v2.ads:128,
    :199), "lenght" typo, missing space before (.
  - Truncation doc now on both Get_Message_Data overloads.
  - Generated indentation of the else / end if block.
  
  Also worth noting for the record: Is_CRC_Valid in v2 has used the correct
  Natural (Header.Len) - 1 form since the original v2 commit (a5812a895) — so
  this PR brings the other sites into line with code that was already right,
  rather than inventing new arithmetic. And reordering Check_CRC ahead of
  Get_Message_Data is safe: v1's Is_CRC_Valid is in out, but its only side
  effect is the one-shot Extras_Added latch folding CRC_EXTRA into the running
  checksum — it never touches In_Buf. The new gating is actually a small
  improvement there, since a length mismatch no longer latches the wrong
  message's CRC_EXTRA.

  Still open — both are last round's moderate findings, unchanged:

  1. v1 Get_Message_Data never got the clamp its v2 sibling gained.
  generator/Ada/mavlink-v1.adb:234:
  Last : constant Natural := Buffer'First + Natural (Header.Len) - 1;
  The modular wrap is fixed, but there's still no Natural'Min (Natural 
  (Header.Len), Buffer'Length). Compare generator/Ada/v2/mavlink-v2.adb:659,
  which now has exactly that.

  2. The v1 no-CRC Decode is still guarded only by pragma Assert.
  generator/mavgen_ada.py:758-761 emits the assert and then calls
  Get_Message_Data unconditionally. Without -gnata the assert vanishes, and the
  call lands in the unclamped routine from #1: a header claiming Len = 255
  decoded into a 9-byte Heartbeat overlay raises Constraint_Error, or corrupts
  the caller's stack under -gnatp. That is precisely the "exception on broken
  data" the PR title promises to fix, and it survives in 2 of the 4 emitted v1
  overloads. Fixing #1 subsumes this; fixing this alone does not.

  - Get_Message_Length → Get_Msg_Length, with Inline added (mavlink-v2.ads:128, :199), "lenght" typo, missing space before (.
  - Truncation doc now on both Get_Message_Data overloads.
  - Generated indentation of the else / end if block.

  Also worth noting for the record: Is_CRC_Valid in v2 has used the correct Natural (Header.Len) - 1 form since the original v2 commit (a5812a895) — so this PR brings the other sites into line
  with code that was already right, rather than inventing new arithmetic. And reordering Check_CRC ahead of Get_Message_Data is safe: v1's Is_CRC_Valid is in out, but its only side effect ithe
  one-shot Extras_Added latch folding CRC_EXTRA into the running checksum — it never touches In_Buf. The new gating is actually a small improvement there, since a length mismatch no longer
  latches the wrong message's CRC_EXTRA.

  Still open — both are last round's moderate findings, unchanged:

  1. v1 Get_Message_Data never got the clamp its v2 sibling gained. generator/Ada/mavlink-v1.adb:234:
  - Generated indentation of the else / end if block.

  Also worth noting for the record: Is_CRC_Valid in v2 has used the correct Natural (Header.Len) - 1 form since the original v2 commit (a5812a895) — so this PR brings the other sites into
  line with code that was already right, rather than inventing new arithmetic. And reordering Check_CRC ahead of Get_Message_Data is safe: v1's Is_CRC_Valid is in out, but its only side
  effect is the one-shot Extras_Added latch folding CRC_EXTRA into the running checksum — it never touches In_Buf. The new gating is actually a small improvement there, since a length
  mismatch no longer latches the wrong message's CRC_EXTRA.

  Still open — both are last round's moderate findings, unchanged:

  1. v1 Get_Message_Data never got the clamp its v2 sibling gained. generator/Ada/mavlink-v1.adb:234:
  Last : constant Natural := Buffer'First + Natural (Header.Len) - 1;
  The modular wrap is fixed, but there's still no Natural'Min (Natural (Header.Len), Buffer'Length). Compare generator/Ada/v2/mavlink-v2.adb:659, which now has exactly that.

  2. The v1 no-CRC Decode is still guarded only by pragma Assert. generator/mavgen_ada.py:758-761 emits the assert and then calls Get_Message_Data unconditionally. Without -gnata the assert
  vanishes, and the call lands in the unclamped routine from #1: a header claiming Len = 255 decoded into a 9-byte Heartbeat overlay raises Constraint_Error, or corrupts the caller's stack
  under -gnatp. That is precisely the "exception on broken data" the PR title promises to fix, and it survives in 2 of the 4 emitted v1 overloads. Fixing #1 subsumes this; fixing this alone
  does not.

  3. v1 checked Decode leaves out Message undefined on both failure paths. generator/mavgen_ada.py:711 switched from := [others => 0] to with Import, and Get_Message_Data is now skipped on
  length mismatch and on bad CRC — so nothing writes Message. The v2 template at :731 still zero-fills. Same documented API, two contracts, undocumented.

  One correction to how this was framed in earlier rounds and by Codex this round: I checked the generated types, and every message field is type X is new Interfaces.Unsigned_N
  (mavgen_ada.py:267) — derived modular types where every bit pattern is a valid representation. There are no Ada enumeration or range types in message records. So this is garbage values,
  not a bounded error or erroneous execution; no Constraint_Error risk. It's still an asymmetry worth resolving, but it's a low, not a moderate.

  Minor, carried over and still not addressed: the no-CRC Decode overloads' generated comments no longer say they skip CRC validation — both overloads now read as if they behave identically.
  And --  For v2: Truncate the extension fields. is still emitted into v1 specs (verified in mavlink-v1-minimal-heartbeats.ads), where per #1 v1 doesn't truncate, it overruns.
  generate_message_methods already receives is_v1, so emitting only the applicable line is a two-line change.

  Recommendation

  The v2 half of this PR is now correct and is a real fix. The v1 half still ships the same bug class it fixes in v2 — one Natural'Min line in mavlink-v1.adb:225's Get_Message_Data closes #1
  and #2 together, with no API change and no regeneration concerns. I'd ask for that one line, treat #3 as a documented decision, and merge. Not worth another full round beyond that.

  Standing caveat: .github/workflows/ still has no Ada job, so none of this is compiled anywhere in CI — your "we need some regression tests here" remains the real fix.

  Want me to draft that one-line clamp as a suggested change, or post this as a PR comment?
```

ReladormAndry:
Hello @peterbarker 
Thank you for the help. I believe the MR is ready to be merged. The concerns regarding v1 decoding are valid; however, those methods include comments stating that exceptions are expected on invalid data. Since alternative decoding variants also preserve exceptions, it remains the user's choice which one to use.

peterbarker:
Thanks @ReladormAndry - merged it!

ReladormAndry:
@peterbarker Thank you!

PR #1264, opened 2026-08-20
mavutil: add Unix domain socket transport

tridge:
this allows for sim_vehicle.py and autotest.py to operate without colliding with other instances


khancyr (approved):
LGTM

wanted for long , I have some patch around for SITL as this is quite conveinant for swarming and docker too !

PR #1266, opened 2026-08-25
mavftp: return from process_ftp_reply on read completion

peterbarker:
A read reports its completion positively - EOF seen with no gaps outstanding (__check_read_finished) - but the synchronous process_ftp_reply wrapper never consulted that: its only exit was idle detection, waiting out idle_detection_time (3.7s) of link silence after the transfer.  On a fast or simulated link the quiet-period tail dwarfs the transfer: fetching @PARAM/param.pck from SITL takes 3.72s of which the transfer itself is 27ms.

The nested process_ftp_reply("TerminateSession") inside __terminate_session was the actual site of the wait, and its special-case discarded replies unparsed, so the terminate ack could never be recognised.  Route those replies through the normal packet handler, track the in-flight terminate with a pending_terminate flag, and return as soon as the read has completed and its session-terminate handshake has been acknowledged.  Idle detection remains the fallback for a lost terminate reply on a lossy link, and the exit for operations with no positive completion signal.

The completion signal has existed since the module was derived from MAVProxy's event-driven mavproxy_ftp (which completes immediately on it, and has no idle detection at all); the synchronous wrapper introduced in 50a291ad0 left it unwired, and 47824938b later raised the idle time from 1.2s to 3.7s.

Measured against ArduPilot SITL, three consecutive gets of @PARAM/param.pck (15580 bytes, 1408 parameters, all decoded correctly): 3.72s each before, 0.027s each after.

### DIFF
--- a/mavftp.py
+++ b/mavftp.py
@@ -15,6 +15,7 @@ SPDX-License-Identifier: GPL-3.0-or-later
 
 import logging
 import os
+import tempfile
 import random
 import struct
 import sys
@@ -377,7 +378,14 @@ class MAVFTP:  # pylint: disable=too-many-instance-attributes
         self.list_temp_result: List[DirectoryEntry] = []
         self.requested_size: int = 0
         self.requested_offset: int = 0
-        self.temp_filename = "/tmp/temp_mavftp_file"
+        # set per-download by __handle_open_ro_reply: a securely
+        # created unique staging file, so concurrent MAVFTP clients on
+        # one host (e.g. parallel simulator test runners) cannot share
+        # a staging file and its name is not predictable
+        self.temp_filename = None
+        # only close file handles this instance opened itself; cmd_put
+        # stores a caller-owned handle in self.fh
+        self.fh_owned = False
 
         self.master = master
         self.target_system = target_system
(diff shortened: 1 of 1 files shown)


amilcarlucas (approved):
Good catch and nice speedup!!!

Wouldn't it be better/safer to use tempfile.mkstemp()/NamedTemporaryFile(delete=False) to create the staging file securely?

amilcarlucas:
And here the AI review:

Yes—one meaningful response-ordering race:

- **P1:** Reset/terminate completion is accepted without correlating the reply to the request sequence number. [`mavftp.py:1318`](/home/amilcar/ardupilot/pymavlink/mavftp.py:1318) clears `pending_reset` or `pending_terminate` for any matching `req_opcode` in the current session; sequence is only used for RTT measurement, not validation. [`mavftp.py:1514`](/home/amilcar/ardupilot/pymavlink/mavftp.py:1514) then exits immediately. A delayed or duplicate `ResetSessions` reply—especially likely because a new instance begins in session 0—can make construction proceed before its own reset has been acknowledged. Likewise, a stale terminate reply can make a completed read return before the current termination is processed. Previously, the idle-delay behavior inadvertently masked this by continuing to wait.

Store the sent reset/terminate sequence and require the response sequence to be `(sent_seq + 1) % 256` before clearing the corresponding pending flag. Also validate the expected remote source system/component if a connection may carry multiple vehicles.

The pre-existing class is not thread-safe, too: concurrent calls on the same `MAVFTP` object can race on `seq`, `last_op`, session, and the new boolean flags. The new changes don’t introduce locking, so they should not be treated as making one instance safe for parallel callers.

peterbarker:
@amilcarlucas fixes pushed based on your review - thanks!

amilcarlucas:
Merged, thanks for this speedup!

PR #1268, opened 2026-08-31
Remove support for the 0.9 variant of the protocol

peterbarker:
```
  MAVLink 1.0 landed in this repo on 2011-07-31, commit 19c0a4f30 by Lorenz Meier — "Moved from MAVLink 0.x to MAVLink v1.0 prototype - YEEHAA!". That's where MAVLINK_STX moved off 0x55; it
  went to a provisional 0xD5 first, then settled on 254/0xFE on 2011-08-24 (43ded8f9d, "Upgraded to python generator").
```

I've run ArduPilot's autotest suite against this pymavlink without problems.


tools/magfit_gps.py
```diff
@@ -128,10 +128,6 @@ def magfit(logfile):
         m = mlog.recv_match(condition=args.condition)
         if m is None:
             break
-        if m.get_type() == "GPS_RAW":
```
khancyr, line 131: unrelated, but OK

khancyr (approved):
LGTM.
That will simplify a lot pymavlink support.
Question : shoulnd't we notify MAVLink holder of this change ?

tools/magfit_gps.py
peterbarker: > unrelated, but OK

Only in the 0.9 message set - we're being pretty brutal here, even removing the 0.9 STX (which might be going a bit far)

peterbarker:
> LGTM. That will simplify a lot pymavlink support. Question : shoulnd't we notify MAVLink holder of this change ?

I poked Hamish.

Who else do you think is interested?

khancyr:
v0.9 pre-date when I start contributing to ArduPilot, so I don't think there is still a lot of system building it explicitly. 
Beside MAVLink organisation that use pymavlink, I don't think we have other declared groups. 

GCS should be v1.0/2.0 already, and other systems such as ROS, DDS, MAVSDK, use v1.0 as minimal, so no big deal. 

Quick look at MAVLink documentation (https://mavlink.io/en/guide/mavlink_version.html) they state v0.9 is unsupported since 2019. So we are fine I think

hamishwillee:
Please please do it!

tridge:

Full report: https://uav.tridgell.net/DevCallReviews/2026_09_02/devcall_pr_reviews.html#prpymavlink-1268

Reviewed at head `8e2ca58846`. **APPROVE.**

The removal is functionally complete. Only `PROTOCOL_1_0` and `PROTOCOL_2_0` remain, `MAVXML(…, "0.9")` now raises `MAVParseError`, and `MAVLINK09`, the v09 dialect selection, the 0x55 start-byte detection, `mode_string_v09` and every conditional caller of `mavlink10()` are gone — `mavlink10()` survives only as two deprecated shims returning True. No remaining file branches on the old wire protocol.

The risk in a change like this is a generator quietly picking up the wrong constant when a conditional is collapsed, so that was checked per backend rather than in aggregate:

- **C, C#, Java** — `MAVLINK_LITTLE_ENDIAN`, CRC-extra 1, aligned-fields 1 and the unconditional `crc_extra` emission all match the old v1.0/v2.0 branch.
- **JavaScript / JS-stable** — `fmtstr = '<'` is the little-endian branch; the removed dictionary entries were unused.
- **Python** — `'<'`, `little_endian/crc_extra/sort_fields = True`, unconditional CRC-extra and field reordering.
- **Spin2** — `LITTLE_ENDIAN = True`.
- **WLua** — had no such conditional; correctly loses only the 0x55 protocol-table entry.

No backend picked up a wrong constant.

Two non-blocking notes:

- **This is a breaking API change and deserves a release note.** `mavparse.PROTOCOL_0_9` now raises `AttributeError` and `from …mavparse import PROTOCOL_0_9` raises `ImportError`. Third-party generators and tooling can break even if they have not generated 0.9 in years — worth naming the removed constant, `MAVLINK09` and the v09 dialect path explicitly.
- **One residual 0.9-ism.** `generator/mavparse.py:66` still translates the legacy `array[N]` type into `int8_t[N]`. `mavschema.xsd` now rejects that spelling, but `--no-validate` still accepts it (reproduced). It cannot re-enable 0.9 wire generation, so this is cleanup: either drop those two lines or note that the compatibility is deliberate.

All 21 checks pass.

tridge (approved):

PR #1278, opened 2026-09-08
generator: parse isDestination and require hasLocation with it

peterbarker:
A destination is somewhere the vehicle goes to, so any enum entry marked isDestination="true" must also be hasLocation="true". mavlink/mavlink#2582 solidifies this in common.xml.

Parse isDestination alongside hasLocation, store it on MAVEnumEntry and raise MAVParseError when the invariant does not hold.  The snapshot test copy of common.xml had two entries which violated it; update them to match upstream, which now marks them hasAltitudeOnly="true".

### DIFF
--- a/generator/mavparse.py
+++ b/generator/mavparse.py
@@ -170,7 +170,7 @@ class MAVEnumParam(object):
             self.description = description
 
 class MAVEnumEntry(object):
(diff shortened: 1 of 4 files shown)


peterbarker:
We're lacking several commits in ArduPilot's repository which allow this to function.

amilcarlucas (approved):

PR #1281, opened 2026-09-14
mavwp: fix invalid-index handling in fence and rally loaders

ARYA-mgc:
## Summary
`MAVRallyLoader.remove()`, `MAVRallyLoader.move()`, `MAVFenceLoader.move()`, and `MAVFenceLoader.remove()` validate the index, print an error on out-of-bounds input, and then **fall through into the list access instead of returning**, raising an uncaught `IndexError`. This is a one-line fix in each function.

## Root Cause
`MAVRallyLoader.set_alt()` already handles this correctly:

```python
def set_alt(self, i, alt, break_alt=None, change_time=True):
    if i < 1 or i > self.rally_count():
        print("Invalid rally point number %u" % i)
        return                              # <- guards correctly
    self.rally_points[i-1].alt = int(alt)
```

The four functions below are missing that `return`:

```python
def remove(self, i):
    if i < 1 or i > self.rally_count():
        print("Invalid rally point number %u" % i)
    self.rally_points.pop(i-1)              # <- executes anyway, crashes
```

## Fix
Add `return` immediately after the error print in all four functions, matching
the existing `set_alt()` pattern. No behavioral change for valid input.

## Testing
Verified directly against `ArduPilot/pymavlink@master` and the published
PyPI package — not just read through:

- Reproduced the crash in all four functions:
```python
  >>> mavwp.MAVRallyLoader().remove(5)
  IndexError: pop from empty list
  >>> mavwp.MAVRallyLoader().move(5, 0, 0)
  IndexError: list index out of range
  >>> mavwp.MAVFenceLoader().move(5, 0, 0)
  IndexError: list index out of range
  >>> mavwp.MAVFenceLoader().remove(5)
  IndexError: pop from empty list
```
- Applied the fix and confirmed all four now fail gracefully (print + return,
  no exception).
- Ran valid add/move/remove sequences on both `MAVRallyLoader` and
  `MAVFenceLoader` afterward to confirm no regression in normal operation.

## Checklist
- [x] Root cause identified and isolated to 4 functions
- [x] Fix matches existing correct pattern (`set_alt`) — no new logic introduced
- [x] Crash reproduced before fix, resolved after fix
- [x] Valid-input regression check passed


ARYA-mgc:
Hii! @peterbarker could you review this PR when you get a chance? Sorry for pinging you directly just wanted to make sure it didn't slip through

peterbarker:
I'm not convinced throwing the exception isn't the correct response here.  Fundamentally this *is* exception territory, we've been asked to do something which is wrong, and IndexError is the most appropriate thing to throw.

Have you looked at the history behind this structure?  Did something get lost?  Was an explicit decision made?  Anything in commit comments?  PR comments?  What about the history of `set_alt`?

What happens when MAVProxy hits this exception in practise?  What dies?  Are there places where it would be convenient to know that the operation actually failed (i.e. catch the exception and handle it gracefully)

ARYA-mgc:
Thanks for the thoughtful feedback, @peterbarker! That's a very fair question. I dug through the git history and MAVProxy call sites to answer each point directly.

### 1. Git History

* **April 19, 2014 (Tridge, `5a91fcb7`):**
  *"pymavlink: added rally point move and remove support"* - added `MAVRallyLoader.remove()` and `move()`. Both have the `if i < 1 ...: print(...)` guard but no `return`. The pre-existing `append_rally_point()` in the same class already used `print(...) + return` for its bounds check, so the pattern was already present but wasn't followed here.

* **April 21–22, 2014 (Tridge, `762bea09`, `198849ff`):**
  *"enable moving of fence points"* and *"added fence point remove()"* - added `MAVFenceLoader.move()` and `remove()` with the same missing `return`.

* **July 15, 2014 (Michael Day, `782ac5a9`):**
  *"added set_alt method to MAVRallyLoader."* - added `set_alt()` with the `print(...) + return` guard.

All four commit messages are one-liners with no body, and I couldn't find any discussion documenting an intentional exception-based design. The history therefore doesn't give a clear indication that the `IndexError` was deliberate; it looks more like these checks were intended to reject invalid indices, but the `return` was missed in these four functions.

### 2. What happens in MAVProxy in practice?

Modern MAVProxy manages fences and rally points through `MissionItemProtocolModule` (the current `mavproxy_fence.py` and `mavproxy_rally.py` don't import `MAVFenceLoader` or `MAVRallyLoader`).

Where the legacy loaders are still used (for example, `mavflightview.py` and the map module's older protocol path), I couldn't find callers wrapping these methods in `try/except`. An out-of-bounds index therefore prints the existing error message and then raises an unhandled `IndexError` traceback. I also couldn't find an existing caller that catches this exception and uses it to handle the failure.

### 3. How should we proceed?

I see two reasonable directions depending on the intended API:

**A. Minimal fix (current PR):** Add `return` after the `print`, matching the `set_alt()` / `append_rally_point()` pattern already present in the same classes. This is the smallest change and prevents the invalid list access.

**B. Explicit exception design:** Remove the `print(...)` statements and explicitly raise `IndexError` or the existing `MAVRallyError` / `MAVFenceError` types. For consistency, that would also suggest revisiting the existing bounds handling in `set_alt()` and `append_rally_point()`.

I'm happy to follow whichever behavior is considered correct for these loaders. Which approach would you prefer?

ARYA-mgc:
I agree that an invalid index could reasonably be considered exception territory. What makes me hesitant to treat the current behavior as intentional is the existing validation pattern: these methods explicitly check the bounds and print an "Invalid ... point number" message. `set_alt()` and `append_rally_point()` use the same pattern and return immediately after the check. The four affected methods are the only cases where execution continues after that validation failure.

So my current interpretation is that the `return` was accidentally omitted. If the intended API is instead to raise for invalid indices, I think that behavior should be made explicit and consistent across the loader methods rather than keeping the current print-then-`IndexError` behavior.

peterbarker:
On Mon, 14 Sep 2026, ARYA MGC wrote:

> So my current interpretation is that the return was accidentally omitted. If the intended API is instead to raise for invalid indices, I
> think that behavior should be made explicit and consistent across the loader methods rather than keeping the current print-then-IndexError
> behavior.

Is this code explicitly causing problems somewhere?  Has an exception been 
raised when it shouldn't?

I'm very hesitant to change the behaviour of a module like this - 
essentially deprecated - unless a bug has been shown.

i.e. I don't think I need to make a decision on which way this module 
*should* be consistent - we just wait until there *is* a problem and work 
out how to fix it then.  Whatever choice is made we run the risk of 
breaking *existing* out-of-tree users (e.g. those explicitly catching 
IndexError would break if we used the custom exceptions)

PR #1282, opened 2026-09-16
examples: cope with MAV_TYPE enumeration renaming in status_msg

peterbarker:
https://github.com/ArduPilot/mavlink/pull/521 renames several MAV_TYPE entries:

 - MAV_TYPE_VTOL_DUOROTOR  -> MAV_TYPE_VTOL_TAILSITTER_DUOROTOR
 - MAV_TYPE_VTOL_QUADROTOR -> MAV_TYPE_VTOL_TAILSITTER_QUADROTOR
 - MAV_TYPE_VTOL_RESERVED2 -> MAV_TYPE_VTOL_FIXEDROTOR
 - MAV_TYPE_VTOL_RESERVED3 -> MAV_TYPE_VTOL_TAILSITTER
 - MAV_TYPE_VTOL_RESERVED4 -> MAV_TYPE_VTOL_TILTWING

The old names are gone, so referencing them unconditionally raises AttributeError against message definitions which contain the renames. Look the affected names up with getattr and skip the ones which are absent, so the example works with definitions from either side of the rename.  This mirrors the existing handling in mavutil.py and the equivalent MAVProxy fix in ArduPilot/MAVProxy#1755.


AP-Review:

Reviewed at head `e1664271a0`.

Full report, with everything that was checked: https://uav.tridgell.net/DevCallReviews/2026_09_16_AIReview/devcall_pr_reviews.html#prpymavlink-1282

**Verdict: APPROVE — no blockers.**

No blocking defects. The rename-coping mechanism is correct and was proven by building pymavlink against *both* sides of the rename and running the example end to end, including a reverse mutation that confirms the fix is load-bearing rather than cosmetic.

### Notes

- `examples/status_msg.py:228` — Cross-PR inconsistency with the sibling ArduPilot/MAVProxy#1755: that PR's plane-type list includes the five new names but only two of the old ones, omitting `VTOL_RESERVED2/3/4`. So against pre-rename definitions MAVProxy will not classify MAV_TYPE 22/23/24 as a plane while this example will. Harmless in practice (ArduPilot only emits 19/20/21 from `get_mav_type()`), and **the more complete list is the one in this PR** — if anything changes it should be #1755, not this. ardupilot#34415 uses the same five new names and values, so all three are consistent on the names themselves.

- `mavutil.py:2402` — The PR body says this "mirrors the existing handling in `mavutil.py`". It does, but the example is strictly *more* complete: `AP_MAV_TYPE_MODE_MAP_DEFAULT`'s rename loop covers only the DUOROTOR/QUADROTOR pairs. Values 22/23/24 were never in that map under either name, so this is a pre-existing absence rather than a rename regression — just noting the library is not the better model here.

- `examples/status_msg.py:187` — The dict plus the getattr loop are rebuilt on every call to `__classify_vehicle_type`. Pre-existing (the dict was already function-local) and it is called once per run; not worth changing.

**As documentation** — which matters for an example — it models the right pattern: a name-string list, `getattr(..., None)`, an explicit `is None` skip, and an inline `# renamed to X` comment on each legacy name so a reader understands why the pairs are adjacent. A user copying it gets code that works on both sides of the rename.

---
Independent cold pass: no findings, APPROVE — agreeing with the primary read.

Reviewed by: Claude (full read of the diff, thread and surrounding source) + an independent Codex cold pass that was not shown these findings. Findings marked _both reviewers_ were reproduced independently twice.

PR #1283, opened 2026-09-16
.github: check branch conventions with a script

peterbarker:
Replaces the inline shell in test_branch_conventions.yml with a script adapted from ArduPilot's Tools/scripts/check_branch_conventions.py, so the checks can also be run locally:

  .github/scripts/check_branch_conventions.py --base-branch origin/master

Adds a check for the commit subjects GitHub's web interface writes for you -- "Apply batched suggestions from code review", "Update <file>" and friends.  Those reach master when review suggestions are committed from the browser and the branch is merged without being tidied up; the change belongs squashed into the commit it fixes.  28 such commits are in this repository's history.

Also carries across the fixup!, subject-length, placeholder-author-email and trailing-newline checks, and the rejection of commits flagged as unfinished.  That last one is reworked to key off a "WIP:", "[WIP]" or "(wip)" marker rather than off a subsystem prefix: ArduPilot requires every subject to carry one, but a third of the last 1000 commits here do not, so that requirement is deliberately not carried across.


AP-Review:
> **Deprecated — see below for the updated review.**

<details><summary>Previous review (2026-09-16)</summary>

Reviewed at head `7f6e2325db`.

Full report, with everything that was checked: https://uav.tridgell.net/DevCallReviews/2026_09_16_AIReview/devcall_pr_reviews.html#prpymavlink-1283

**Verdict: COMMENT** — 2 bugs, 1 issue below.

**The gate is real** — every violation class exits 1, the workflow has no `continue-on-error` or `|| true`, and it fires on `pull_request`. It is a faithful trim of ArduPilot's already-merged `check_branch_conventions.py`. Two latent filtering bugs let a real violation through silently, which is the one failure mode a gate must not have.

### Bugs

- `.github/scripts/check_branch_conventions.py:265` _(found independently by both reviewers)_ — **The trailing-newline check silently skips any path git quotes.** `git diff --name-status` emits `"h\303\251llo.py"` — literal quotes, octal escapes — for non-ASCII names, and likewise for a tab, backslash or quote in a name. `pathlib.Path('"h\303\251llo.py"').suffix` is `.py"`, which is not in `SOURCE_EXTENSIONS`, so the file is dropped before it is ever checked. Reproduced: adding `héllo.py` with no trailing newline alongside `with space.py` with identical content reports *only* `with space.py` — the violation passes invisibly. (Even if the suffix matched, `git cat-file blob HEAD:"h\303\251…"` would fail and the helper returns `None`, also skipped.) Fix: `git diff -z --name-status`, or `git -c core.quotePath=false`. No non-ASCII paths exist in the tree today, so this is latent. **Independently reproduced by both reviewers**, the cold pass with `café.py`.

- `.github/scripts/check_branch_conventions.py:246` — Symlinks with a source extension are false positives. `git cat-file blob HEAD:link.py` on a symlink returns the *target path* as blob content, which never ends in a newline, so `ln -s base.txt link.py` + commit produces `✗ link.py is added without a trailing newline.` and exit 1. The repo does track symlinks (`generator/javascript/node_modules/jspack`, `long`) — none currently with a source extension, so this is latent too. Fix: filter on file mode (`git diff --raw` / `git ls-tree`), not just extension.

### Issues

- `.github/scripts/check_branch_conventions.py:300` — `--base-branch ""` makes the whole gate a silent no-op: `git log ..HEAD` degenerates to `HEAD..HEAD`, so `commits` is empty and it prints "Branch adds no commits; nothing to check" and **exits 0** without running a single check. The `is None` test at :300 means an empty string does not trigger the merge-base fallback either. Not reachable from the workflow (`base.ref` is always populated on `pull_request`), but the script is advertised for local use in its own docstring. `if not self.base_branch:` fixes it.

_6 further minor notes (style, coverage, pre-existing items) are in the report rather than here, so they don't bury the above._

---
Independent cold pass: COMMENT, and it independently reproduced the git-quoted-filename BUG with the same mechanism and the same suggested `-z` fix.

Reviewed by: Claude (full read of the diff, thread and surrounding source) + an independent Codex cold pass that was not shown these findings. Findings marked _both reviewers_ were reproduced independently twice.


</details>

AP-Review:

Re-reviewed at head `349977799b` (previously `7f6e2325db`); my earlier comment above is superseded.

Full report: https://uav.tridgell.net/DevCallReviews/followups/2026_09_17_0037/devcall_pr_reviews.html#prpymavlink-1283

---

**Verdict: APPROVE** — all three previous findings are fixed, and I re-ran the original reproducers rather than reading the diff. CI is 21/21 green.

### Previous round — all three resolved

| Finding | Status |
|---|---|
| Trailing-newline check silently skipped any path git quotes | **RESOLVED** — `git diff -z --raw --no-abbrev -M` at `:267-274` with NUL-field parsing at `:278-283`. Re-ran the exact reproducer: `héllo.py`, `with space.py`, `tab\there.py`, `quote".py` and `back\slash.py` are now **all five** flagged, exit 1 (previously only `with space.py`). Rename-to-unicode also caught |
| Symlinks with a source extension were false positives | **RESOLVED** — `REGULAR_FILE_MODES` at `:92`, applied to `new_mode` at `:287` and `old_mode` at `:298`. `ln -s base.txt link.py` plus a dangling `dangling.py` → exit 0; genuine cases still fire |
| `--base-branch ""` made the gate a silent no-op | **RESOLVED** — `if not self.base_branch:` at `:315`; it now falls back to the merge base and runs every check |

I ran the script against ~45 distinct inputs rather than reading it: the quoted-path cases, symlinks, modifications, deletions, pure renames, rename+strip, uppercase extensions, empty files, merge commits, `fixup!`/`squash!`/`amend!`, eight WIP/DRAFT/DEBUG/TMP/FIXME spellings against three near-misses that are correctly allowed (`WIP fix a thing`, `TEMP sensor: add driver`, `Debug the foo handler`), nine GitHub-web-interface subjects against five legitimate `Update…` subjects, the 160/161-char boundary, `example.com` authors, detached HEAD, HEAD==base, empty commits, unborn HEAD, not-a-git-repo, and running from a subdirectory. Everything behaved as documented except the one case below.

The convention also matches project practice: of 3394 commits the GitHub-UI rule matches exactly **28**, which reproduces your PR body's "28 such commits" claim precisely, with zero false positives on hand-written subjects. And the workflow hardening is right — `pull_request` (not `pull_request_target`), no `continue-on-error`, no `|| true`, `permissions: contents: read`, `fetch-depth: 0`, and PR-controlled values reaching `run:` via `env:` rather than `${{ }}` interpolation (`test_branch_conventions.yml:19-23`), which closes a real shell-injection vector the old inline version had.

### Notes — none blocking

- `:284` — type changes (`T`) are skipped, so a symlink converted to a regular source file escapes the newline check. `rm tc.py && printf 'evil = 1' > tc.py` where `tc.py` was a symlink gives `:120000 100644 … T`, and the script prints the pass message and exits 0. `if status[0] not in "AMRT"` fixes it — the existing `new_mode`/`old_mode` guards already do the right thing for both branches. This is the only case I found where the gate passes a real violation.
- `:210` — "subject too long" misdiagnoses a missing blank line. `git log --pretty=%s` folds the whole first *paragraph* into one subject, so a commit whose real first line is 22 chars but which lacks the blank line before its body is reported as a 237-char subject. Not hypothetical: **all 22 commits in pymavlink's history that exceed 160 chars are folded paragraphs of exactly this shape** (most recent 2024-02-01); I inspected the first four and every one has a short real subject. Comparing `%s` against the first line of `%B` and saying "no blank line between the subject and the body" would be a much better message.
- `:110` — git failures surface as an uncaught traceback rather than a message. Fail-safe (every path I exercised exits 1, never 0), but user-hostile for the advertised local use: `--base-branch nosuchbranch` prints `RuntimeError: git log nosuchbranch..HEAD … failed: fatal: ambiguous argument`. Catching `RuntimeError`/`CalledProcessError` in `__main__` and printing `f"{FAIL} {e}"` before `sys.exit(1)` would cover it.

peterbarker:
Tested here: https://github.com/peterbarker/pymavlink/pull/1

PR #1285, opened 2026-09-17
Default unused MAV_CMD params to NaN instead of 0

hamishwillee:
## Summary
- MAV_CMD entries with fewer than 7 declared `<param>` elements are auto-filled by `mavparse.py` for docs generation. The synthesized default was hardcoded to `0`; this changes it to `NaN`, matching the convention used for "unused"/ignored float command params elsewhere in MAV_CMD docs.
- `param5`/`param6` also double as `int32` lat/lon fields when the command is sent via `COMMAND_INT` or `MISSION_ITEM_INT`, so their auto-generated description now explains the `INT32_MAX` default used in that case.
- This supports a companion mavlink repo PR that removes redundant `Empty`/`Reserved` `<param>` declarations from `common.xml`, relying on this auto-fill for the generated docs.

## Test plan
- [x] `MAVXML(...)` still parses all mavlink message_definitions files without error
- [x] Verified generated docs (C/C++/C#/Java comments, Python `enums["MAV_CMD"][N].param[idx]`) show `Reserved (default:NaN)` / the `INT32_MAX` note for param5/6


peterbarker:
I really think we need the flightstacks and GCSs to catch up before we can seriously progress this.  i.e. the stacks have to be happy accepting either of 0 or NaN as potentially-filled-in values first.

If users start to send in NaNs as a matter of course, ArduPilot will be absolutely inundated with "fix FPE" PRs spot-fixing commands rather than a correct structural fix.

https://github.com/ArduPilot/ardupilot/pull/32291 also highlights a conceptual issue.  I believe that was written as a talking point - that we must understand what all non-NaN ( or 0) values do or we fail the command.  This was because a vehicle honouring new enumeration values in a packet would fly significantly different to a vehicle which didn't understand (and would thus ignore) those values.  That's  only *related to* which of NaN and 0 we use.

https://github.com/ArduPilot/ardupilot/pull/28866/changes is my attempt at converting from command-long-flag-values to command-int-flag-values, which would be required by this.

PR #1286, opened 2026-09-18
mavgen_python: emit aliases for renamed enumeration entries

peterbarker:
https://github.com/ArduPilot/mavlink/pull/521 renames several MAV_TYPE entries, removing the old names.  Code referencing the old names then raises AttributeError, and code using the new names fails against definitions which predate the rename.

The XML carries no machine-readable record of a rename, so keep a table of (old name, new name) pairs in the generator.  Whichever name of a pair is missing from the definitions is emitted as a module-level alias of the other, so code written against either side of the rename keeps working.  The enums dict is unchanged and still reports the name used in the definitions.


peterbarker:
I've verified that an old MAVProxy runs with Python bindings generated from using the message definitions linked from https://github.com/ArduPilot/ardupilot/pull/34415, where that older MAVProxy definitely does not 

```
pbarker@threads:~/rc/ardupilot$ mavproxy.py 
  File "/home/pbarker/.local/bin/mavproxy.py", line 38, in <module>
    from MAVProxy.modules.mavproxy_link import preferred_ports
  File "/home/pbarker/.local/lib/python3.10/site-packages/MAVProxy/modules/mavproxy_link.py", line 77, in <module>
    class LinkModule(mp_module.MPModule):
  File "/home/pbarker/.local/lib/python3.10/site-packages/MAVProxy/modules/mavproxy_link.py", line 714, in LinkModule
    mavutil.mavlink.MAV_TYPE_VTOL_QUADROTOR,
AttributeError: module 'pymavlink.dialects.v10.all' has no attribute 'MAV_TYPE_VTOL_QUADROTOR'
pbarker@threads:~/rc/ardupilot$ cd ../pymavlink
```

I think this is really only a problem for the pymavlink generator as we ship the bindings in the same package as the generator.

AP-Review:

Full report: [https://uav.tridgell.net/DevCallReviews/2026_09_18/devcall_pr_reviews.html#prpymavlink-1286](https://uav.tridgell.net/DevCallReviews/2026_09_18/devcall_pr_reviews.html#prpymavlink-1286)

Reviewed at head `2e97a599ef` — **APPROVE, no blockers.** Noting this is still a draft, so treat it as a review of what is there rather than a merge sign-off.

Two independent passes ran on this: a full read by me, and a separate cold `codex exec` agent given only the PR number and no knowledge of my findings. **The cold agent returned "no findings, APPROVE".** Between us we checked:

- **Generated code runs, both directions.** Generated and imported bindings from the post-rename `minimal.xml` (emits `MAV_TYPE_VTOL_DUOROTOR = MAV_TYPE_VTOL_TAILSITTER_DUOROTOR` ×5) and from a pre-rename copy (the mirror set). `MAV_TYPE_VTOL_DUOROTOR == 19 == MAV_TYPE_VTOL_TAILSITTER_DUOROTOR`, `QUADROTOR == 20`, `RESERVED2 == 22`.
- **No reverse-lookup pollution** — the thing that would have made this unmergeable. After import, `enums['MAV_TYPE'][19].name` is still the canonical `MAV_TYPE_VTOL_TAILSITTER_DUOROTOR`, `[20]` and `[22]` likewise, and `len(enums['MAV_TYPE']) == 51`, unchanged. Checked the MAVProxy and ArduPilot consumers; MissionPlanner's C# path is unaffected.
- **No collision is possible** — the `old in values and new not in values` / `elif` guard means no alias is emitted at all when a dialect carries both names, so an alias can never shadow a real entry.
- **Additive only** — base-vs-PR generator output is byte-identical for a dialect with no renamed entries, and for `minimal`/`common`/`ardupilotmega`/`all` the sole difference is the six-line alias block. Same under `--wire-protocol=1.0`.
- **Lint and types on the generated file** under the repo's own configs: `flake8` clean, `mypy` (with `allow_redefinition = False`) *Success: no issues found*.
- **The table is accurate and the hardcoding is justified** — all five pairs match upstream `minimal.xml`; `_TILTROTOR` and `_RESERVED5` were correctly not renamed and are correctly absent. `mavparse.py` does parse `<deprecated replaced_by=>`, but the renamed entries carry no such tag, so there is no machine-readable rename record to derive the table from.

### One thing worth adding before it leaves draft

**ISSUE — no test exercises the new code, and CI cannot reach the half that matters.** Nothing under `tests/` touches `generate_enum_entry_aliases`, and CI builds against `ArduPilot/mavlink` master, which still carries the **old** names — I fetched `message_definitions/v1.0/minimal.xml` from that branch (46788 bytes) and it returns `MAV_TYPE_VTOL_DUOROTOR`, `_QUADROTOR`, `_RESERVED2/3/4`, and ArduPilot/mavlink#521 is still open. So the 21 green checks exercise only old→new aliasing; the **new→old direction, which is the one in the traceback that motivated the PR, is untested everywhere**.

`tests/snapshottests/resources/minimal.xml` already carries the post-rename names, so a small pytest running the generator over it and asserting `mod.MAV_TYPE_VTOL_DUOROTOR == mod.MAV_TYPE_VTOL_TAILSITTER_DUOROTOR` would cover exactly the gap CI cannot reach today.

### Minor notes

- **`ENUM_ENTRY_RENAMES` placement.** The table is generator-local to Python, so the C/C#/Java/JS emitters still break across the rename — at compile time rather than at runtime. Your answer in the thread (bindings ship in the same package as the generator) is fair and I am not re-raising it as an objection; the cheap improvement is to put the table in `generator/mavparse.py` so a future `mavgen_c.py` `#define` emitter reuses it rather than copying it.
- **No guard against a malformed table.** Two rows sharing an alias target, or a name appearing as `old` in one row and `new` in another, would emit two module-level assignments with last-one-wins and no error. The current five pairs are disjoint so nothing is wrong today; making the table a dict would make a future bad edit loud.
- **Feature-detection.** After this, `hasattr(mavlink, 'MAV_TYPE_VTOL_TAILSITTER_DUOROTOR')` is always true, so `hasattr`-based detection of which side of the rename the definitions are on stops working. pymavlink's own two instances (`mavutil.py:2402-2412` and the rewritten `examples/status_msg.py`) are benign because the alias is the same integer and both names collapse to one dict key — but worth a line in the commit message so third parties know their workarounds become no-ops.
- `values` is built as a dict but only membership is tested; a `set` would read more clearly.

peterbarker:
... I should also say I started to investigate throwing DeprecatedExceptions.  It's non-trivial but probably doable.

